PR #26, opened 2026-06-28
fix: ⌘W dismisses the focus-stealing cover instead of closing the session under it

umputun:
when a full-pane cover is shown (the quick terminal, the scratch terminal, or a session overlay), ⌘W closed the session behind the cover instead of the cover the user is looking at. It's a data-loss bug: ⌘W terminates a shell the user never targeted.

reported for the quick terminal; the same leak hit the scratch terminal (⌘J) and a session overlay. Root cause: the "Close Session" command (the ⌘W menu item + the "Close Session" palette item, both via `AppActions.closeActiveSession()`) closed `selectedSessionID` unconditionally, with no check for a cover. The covers are in-app overlays over the same window, and ⌘W is a menu key-equivalent that fires regardless of which surface holds first responder, so it leaked straight to the session-close path.

**fix.** `closeActiveSession()` now dismisses the topmost visible cover and only closes the selected session when none is up, in z-order precedence (the quick terminal is window-topmost, an overlay sits above the scratch):

- quick terminal -> `hide()` (shell survives, same as the margin-tap / toggle)
- overlay -> `closeOverlay()` (its only dismiss; fires for full and floating overlays, both hold focus)
- scratch -> `toggleScratch()` (hides keep-alive, same as ⌘J)

both ⌘W and the palette item go through this one seam, so both are fixed. No modality needed, this mirrors the existing cover-aware guards already used for ⌘F (`coverHidesActiveSession`).

**keep-in-sync:** neutral, not a new action. Control already covers each dismiss (`quick hide`, `session overlay close`, `session.close`).

**tests:** new ⌘W cases for the quick terminal (`QuickTerminalUITests`) and the scratch + overlay (`ControlAPIUITests`). All pass; `swift test` is 768/768 green.


umputun:
related to #27

PR #32, opened 2026-06-29
feat: per-session background watermark (session.background)

fkirill:
### Motivation

Make it obvious which environment a session targets — PROD, STAGE, DEV. A large environment name rendered behind the terminal grid keeps that context in view at a glance, so a PROD session is hard to mistake for a scratch one. The goal is fewer accidental destructive commands in important environments: the watermark is passive, always-on context rather than a prompt you can click through. It's driven over the control channel, so a wrapper or login script can stamp the environment automatically when it opens the session.

### What it adds

A new control command and CLI surface to set a per-session background image or rasterized text behind the terminal:

```
agtermctl session background image <path> [--opacity] [--fit] [--position] [--repeat]
agtermctl session background text  <text> [--color #rrggbb] [--opacity] [--fit] [--position]
agtermctl session background clear
```

libghostty auto-fits the image to the surface and re-fits on resize, so there's no app-side resize code. A `.text` watermark rasterizes to a PNG under `<stateDir>/watermarks/` via an app-side `WatermarkRenderer` (AppKit), tinted by the terminal foreground by default.

### Design

- `BackgroundWatermark` spec is host-free + `Codable`, persisted on `SessionSnapshot` — survives restart, re-rendering the `.text` PNG on restore.
- Applied as a per-surface ghostty config overlay (`WatermarkConfig.overlayText`): the `background-image*` lines plus `background-opacity = 1` (so the image shows under window translucency — image opacity is relative to it, `0 × anything = 0`) and a `font-size` line (so a per-surface `update_config` doesn't reset cmd-+/- zoom).
- Each per-surface config is retained and freed only on surface teardown. A global config reload re-asserts each surface's watermark after the shared broadcast wipes it.

### Validation & safety

Inputs validated at the control boundary by a shared host-free `WatermarkConfig`, with the same checks backing CLI `validate()` so a bad value fails before any round-trip:

- `fit` / `position` enums, `opacity` finite `0...1`, `color` `#rrggbb`.
- `text` capped at 256 chars so the rasterized bitmap stays bounded; `renderText` also guards a 64Mpx ceiling.
- Config emission uses fixed-point (no scientific notation — ghostty's parser rejects `1e-05`) and is total against non-finite / out-of-range values.

Rendered `.text` PNGs are swept on `clear` and on permanent session removal (`closeSession` / `removeWorkspace`). Window-close keeps the session, so it isn't swept.

### Scope of the change

- `agtermCore`: `BackgroundWatermark`, `WatermarkConfig`, `WatermarkStorage`, `Command.sessionBackground` + `ControlArgs` fields, `SessionSnapshot` field.
- App target: `WatermarkRenderer`, the `setBackground` control dispatch, per-surface config overlay handling.
- CLI: `session background image|text|clear` in `agtermctl`.
- Bundled agent skill docs updated for the new command.

### Tests

- `swift test` (host-free): protocol round-trip, CLI parse/validate, `WatermarkConfig`, `WatermarkStorage`.
- e2e coverage in the control-API UI tests.
- Manually verified: `image`, `text`, `clear`; restart restore; resize re-fit; global config reload re-asserts.

Happy to adjust naming, defaults, or scope to fit the project's conventions.


umputun (changes requested):
happy to take this, both image and text. Keep-in-sync is all there (four surfaces + the 46→47 bump) and `swift test` + the Debug build are green locally. Few things to fix before merge:

**1. image path can inject ghostty config lines.** `WatermarkConfig.overlayText` writes `background-image = <path>` raw and unquoted, and the `.image` boundary only checks the extension + `fileExists`. So a path with a newline injects arbitrary keys into the per-surface overlay: a file literally named `x.png\nclipboard-read = allow\ny.png` passes `isSupportedImage` + `fileExists`, and the overlay then carries a standalone `clipboard-read = allow` line that wins on that surface. Owner-only socket + the `fileExists` gate keep it well short of RCE, but it's persisted and re-applied on restore, and this surface is meant to be driven by agents from semi-trusted input. `imagePath` is the only free-text field reaching a config line, so one guard at the boundary closes the whole thing:

```swift
guard !path.unicodeScalars.contains(where: { $0.value < 0x20 }) else {
    return ControlResponse(ok: false, error: "image path must not contain control characters")
}
```

mirror the same check in the CLI `Image.validate()` so it fails before the round-trip.

**2. config.reload desyncs font zoom on a watermarked session.** in `SettingsModel.reloadGhosttyConfig`, `reloadConfig(...)` runs first and `reapplyWatermarkIfNeeded` re-emits `font-size = <old zoom>` from the still-old `session.fontSize`, then `resetSessionFontSizesAllWindows()` nils it. Net: the watermarked pane keeps its zoom on screen but the snapshot persists `fontSize == nil`, so model and surface disagree until the next font change. Swapping the two lines (reset before reload) makes `reapplyWatermarkIfNeeded` read the cleared value and keeps it consistent with the documented "reload clears per-session zoom" contract.

**3. window.delete leaks `.text` PNGs.** the PNG sweep is wired into `closeSession`/`removeWorkspace`/clear but not `WindowLibrary.removeWindow`, so deleting a window permanently drops its sessions yet leaves their `<stateDir>/watermarks/<id>.png` behind. Collect the window's session ids before removal and `removeRenderedText` each.

**4. minor: `ownedConfigs` grows per apply, not per change.** `ControlServer.setBackground` calls `applyWatermark` unconditionally and every `config.reload` re-applies, so a scripted set-loop or repeated reloads keep appending retained C configs freed only at teardown. Teardown-bounded and consistent with the existing never-free decision on the shared config, so low priority, but a change-gate in `setBackgroundWatermark` (skip re-apply when the spec is unchanged) would cap it.

the `background-opacity = 1` override making a watermarked pane opaque under translucency is fine as intended, just noting it's there and documented.

unrelated aside: root `CLAUDE.md` still says "44-command catalog", it was already 2 behind before this PR, worth a one-line fixup whenever.

umputun (commented):
round 2 - #1 (injection) and #4 are sorted. #2 and #3 each still have a gap, and the branch needs a rebase before merge.

still open:

1. **#2 fix only covers one of the two reload callers.** `reloadGhosttyConfig` is reset-before-reload now, but `apply()` (`SettingsModel.swift:484→488`) still reloads then resets - and that's the path Settings font/theme changes and the theme-preview palette take. Since `reloadConfig` now re-asserts the watermark, a watermarked + ⌘±-zoomed session keeps its zoom on screen after a theme change while the snapshot persists `fontSize == nil` - same desync as round 1, just the more common caller. Reorder `apply()` to reset-before-reload (or hoist the reset into `reloadConfig`), plus a test running a watermarked session through a theme/settings apply.

2. **#3 still orphans `.text` PNGs when deleting a CLOSED window.** `removeWindow` sweeps via `stores[id]?.workspaces` (`WindowLibrary.swift:341`), which is empty when the window isn't open - so `window.delete` on a closed window drops the snapshot + index entry but leaves `<stateDir>/watermarks/<id>.png` behind for good (the session ids go with it). When `stores[id] == nil`, load the per-window snapshot first and sweep each session's PNG. A closed-window-delete test with a preexisting PNG would lock it in.

minor: switching a session from text→image leaves the old `<id>.png` until the next clear or teardown (clear reclaims it since the path is id-keyed). A one-liner dropping the rendered PNG when the previous watermark was `.text` and the new one isn't would close it.

rebase: branch is behind master and conflicts. `git merge-tree` shows just one trivial conflict in `AppStoreTests.swift` - you and #35 both append a `@Test` after the same anchor, keep both blocks; everything else auto-merges. The 46→47 count is correct post-merge.

#4 note: the set-loop path is gated and tested. The reload re-assert still appends one owned config per watermarked surface per reload, but that's the teardown-bounded class I flagged as low last round, so fine as-is - just the commit msg's `including on config.reload re-assert` overstates it (the reload path isn't gated, only bounded).

fkirill:
Round 2 addressed and rebased on master.

**#2 — `apply()` caller.** Both reload callers now funnel through one private `reloadConfigClearingSessionZoom()` helper in `SettingsModel` (reset-before-reload). `apply()` — the path Settings font/theme changes and the theme-preview palette take — and `reloadGhosttyConfig()` can't drift apart now. App-target, no unit host, and the desync isn't observable over the control API, so it's guarded by the single funnel rather than a test. (`86ad309`)

**#3 — closed-window `.text` PNGs.** `removeWindow` reads the session ids from the persisted snapshot when there's no live store (`stores[id] == nil`) and sweeps each. It passes the library's own `directory` as `stateDir` — the same root `WatermarkStorage` resolves against, so it's production-identical and lets a test sweep into a temp dir without touching `AGTERM_STATE_DIR`. Added closed- and open-window delete tests, each with a preexisting PNG. (`86ad309`)

**minor (text→image).** `setBackgroundWatermark` drops the rendered `<id>.png` when the previous watermark was `.text` and the new one isn't. (`86ad309`)

**#4 / commit msg.** The round-1 fix commit was reworded in the rebase — its #4 bullet scopes to the set-loop only, no "including on config.reload re-assert" claim. The reload re-assert stays teardown-bounded as you noted.

**rebase.** On current master. The `AppStoreTests.swift` conflict with #35 was already resolved — this branch carries #35, so both `@Test` blocks are present. 46→47 count holds post-merge.

Two extra fixes from a self-review pass while in there (`bfcfc13`):
- `reloadConfig` now resolves theme colors before re-asserting watermarks. A default-tinted `.text` watermark re-renders reading `terminalForegroundColor`, so without this its color lagged one reload behind a theme change.
- `WatermarkRenderer.renderText` logs a warning on each failure path (bitmap alloc, PNG encode, write) instead of returning `false` silently — a `.text` PNG that fails to render no longer reports success with no diagnostic.

`swift test` green (824), Debug build green, UI tests green (`ControlAPIUITests` + `MultiWindowUITests`, 94 tests).

umputun (commented):
round 3 - all three still-open items from round 2 check out:

1. **#2 font-zoom** - both reload callers funnel through `reloadConfigClearingSessionZoom()` now, reset before reload (`SettingsModel.swift:319-324`), and `apply()` (`:496`) routes through it too. The common caller is covered.
2. **#3 closed-window PNG** - `removeWindow` loads the persisted snapshot when `stores[id] == nil` and sweeps each session's PNG (`WindowLibrary.swift:346-350`), with tests for both the open and closed paths.
3. **text→image** - `setBackgroundWatermark` drops the stale `<id>.png` on the transition (`AppStore.swift:623-625`).

824 swift tests green here.

still needs a rebase - branch is 1 behind master (#38 merged after your last rebase). Two textual conflicts only, `control-api.md` and `ControlProtocol.swift`, both just #38 and this PR adding near the same anchors, no logic to reconcile. One correction to my round-2 note: #38 added a `--sound` option to `session.status`, not a new command, so master's catalog stays at 46 and 47 here is correct, nothing to reconcile on the count.

three minor things, none blocking:

- restore doesn't re-validate the persisted watermark. `AppStore.restore` (`:726`) assigns it raw and `overlayText` emits `fit`/`position`/`imagePath` without re-running `isValidImagePath`. Only reachable by hand-editing `workspaces.json`, which is the same trust as editing the ghostty conf directly, so it's defense-in-depth, not the agent-injection vector closed in round 1. If you want belt-and-suspenders, re-check control chars in `materialize`/`overlayText`.
- `fit`/`position` are raw `String` checked against the `validFits`/`validPositions` arrays, while `Kind` in the same struct (and `AgentStatus`/`PaneRole` elsewhere) are `CaseIterable` enums. A `Fit`/`Position` enum would serialize identically, drop the two arrays, and give compile-time safety. Consistency nit.
- no read-back: you can set/clear a watermark but `tree` doesn't report it. Fine as a follow-up for set/clear/query symmetry, not needed here.

rebase is the only blocker from my side, the three above are all optional.

fkirill:
Rebased onto current master and addressed the three round-3 optionals.

**Rebase.** Master moved past your round-3 reference — `session.resize` merged after #38, so master's catalog is 47 and this PR is now **48**, not 47. Updated every count reference (`control-api.md`, `SKILL.md`, `CLAUDE.md`) to 48. Conflicts were all additive (`ControlProtocol.swift`, `Commands.swift`, the three docs) — no logic to reconcile.

**Round-3 minors** (`0a486bb`):
1. **restore re-validation** — `overlayText` (and `WatermarkRenderer.materialize`) re-check `imagePath` for control chars on emit, so a hand-edited `workspaces.json` can't inject a config line on restore. `fit`/`position` are typed enums now (below), so `imagePath` is the only remaining free-text field.
2. **fit/position enums** — `BackgroundWatermark.Fit`/`Position` are `CaseIterable` enums like `Kind`. Raw values match ghostty's keys so the wire + persisted JSON stay byte-identical; dropped `validFits`/`validPositions` (derived from `.allCases`), and the boundary converts the validated `String` args to the enums. Also makes them injection-proof.
3. **tree read-back** — `tree` now carries each session's watermark on `ControlSessionNode.background` (omitted when nil), for set/clear/query symmetry.

Tests: node round-trip (+omit-nil), fit/position serialize as raw strings, overlay drops a control-char path, e2e tree read-back. Skill + control-api docs updated.

**One unrelated fix** (`0a486bb`, `878570d`): the five-tab Settings refactor moved the notification-badge and attention-indicator toggles to the new Notifications tab but left `testNotificationBadgeToggleHidesAndShowsBadge` and `testAttentionButtonTogglesLiveFromSettings` pointing at General — repointed them so the full suite is green. Happy to split these into a separate PR if you'd rather keep this one watermark-only.

`swift test` (860), the Debug build, and the full XCUITest suite are all green.

umputun (commented):
round 4. The round-3 items all check out: typed `fit`/`position` enums, restore-path `imagePath` re-validation, `tree` read-back, both reload callers going through `reloadConfigClearingSessionZoom`, the closed-window PNG sweep, the text→image PNG drop. Injection stays closed on both the set and restore paths.

blockers before merge:

**1. a bad persisted watermark wipes the whole saved tree on restore.** `SessionSnapshot` is built to never fail the load on an unknown field, every optional carries the "decodes as nil instead of failing the load and wiping the saved tree" note. But `Optional` only tolerates a *missing* key, not a present-but-invalid one, and `BackgroundWatermark.kind` is non-optional (`fit`/`position` are typed enums). So a persisted watermark with an unknown `kind`/`fit`/`position` throws `DataCorrupted`, fails the entire `SessionSnapshot`, and `PersistenceStore.load` starts fresh, so every workspace and session is gone on next launch. Realistic trigger is a downgrade after a later release adds a `fit`/`position`/`kind` value the older build can't decode, or a hand-edit typo. Fix: decode the watermark lossily (`try?` so a bad spec drops to nil while the rest of the session survives), plus a regression test with an invalid `fit`/`kind` that still restores the session.

**2. swiftlint `--strict` is red, one of them is yours:** `function_parameter_count` on the `ControlServer.swift` `session.background` dispatch, 10 params. Option struct.

the two `type_body_length` hits (`AppStoreTests.swift`, `ControlAPIUITests.swift`) are `.swiftlint.yml` ceiling bumps, leave those, I'll do them after merge.

**3. reload path still leaks a config per reload.** I called this teardown-bounded and low in round 2, but the round-4 gate only covers the set path, so `reapplyWatermarkIfNeeded` still appends a fresh retained `ghostty_config_t` on every `config.reload` per watermarked surface. Since `config.reload` is scriptable, cap it now: free or replace the prior owned config on re-apply instead of appending.

fkirill:
Round-4 blockers fixed, and rebased onto latest master (v0.4.2) as a linear history (the branch had a `Merge branch 'master'` on it — replayed onto master instead so it stays rebase-based; all the merged content — #33 `window.zoom`, #61 command-restore, #63 — is preserved since it's on master).

**Round-4 blockers**

1. **Tree-wipe on a bad persisted watermark.** `SessionSnapshot` now decodes `backgroundWatermark` lossily via a custom `init(from:)` — a present-but-invalid `kind`/`fit`/`position` (a downgrade past a newer enum value, or a hand-edit typo) drops to nil instead of throwing `DataCorrupted`, so the session and the rest of the tree survive. Every other field keeps `decodeIfPresent` for the missing-key forward-compat. Regression test covers an invalid `fit`/`position`/`kind` still restoring the session.
2. **`function_parameter_count`.** `setBackground` takes `(target, ControlArgs?)` and unpacks the watermark fields inside — down from 10 positional params.
3. **reload-path config leak.** `ownedConfigs` is capped at one per surface: each re-apply frees the prior config and keeps only the current (safe once `update_config` has switched the surface off the old one), so a scriptable `config.reload` loop can't grow it. Still freed on teardown.

**Rebase integration.** #61 added `SessionSnapshot.initialCommand`; since the lossy decode gave `SessionSnapshot` a custom `init(from:)`, I made sure `initialCommand` is in the `CodingKeys` + decoded (else it'd be silently dropped — command-restore would break). Command catalog is now **49** (master's 48 incl `window.zoom` + `session.background`); every count reference updated and both commands are in the catalogs.

`swift test` (868), `swiftlint --strict` (0 violations), and the Debug build are green; the `session.background` e2e passes. The swiftlint red you'd have seen was on the older merge base — latest master already carries the ceiling bump.

umputun (approved):
round 5 - all three round-4 blockers are sorted:

1. lossy watermark decode keeps the saved tree on a bad spec - a present-but-invalid `kind`/`fit`/`position` drops to nil instead of throwing `DataCorrupted`, and `initialCommand` is in the custom decoder's keys so command-restore survives the rebase. Regression test covers all three bad-value cases.
2. `setBackground(target, args)` unpacks internally, swiftlint's green.
3. the `ownedConfigs` cap sits in `applyWatermarkFromSession` now, so the reload path is bounded too, not just the set path.

`swift test` (868) + `swiftlint --strict` green here, CI green too - merging, thx.

PR #33, opened 2026-06-29
feat: double-click window header to zoom

fkirill:
## What

Double-clicking the window header now zooms the window — the standard macOS maximize-to-screen toggle (fills the screen's visible frame, not native fullscreen; a second double-click restores the prior frame).

## Why

The custom titlebar is a SwiftUI view, not AppKit's native title bar, so the OS double-click-to-zoom never reached it. As a side effect the whole header was not draggable either (only the native ~28px band was).

## How

`WindowControlArea` — a transparent `NSViewRepresentable` placed behind `customTitlebar`'s decorative regions — handles `mouseDown`: `clickCount == 2` → `NSWindow.zoom(nil)`, otherwise `performDrag` (so the full header drags, not just the native band). `mouseDownCanMoveWindow = false` so our handler sees the double-click. The header's decorative parts (traffic-light spacer, divider gap, title text) opt out of hit-testing via `.allowsHitTesting(false)` so their region falls through to the layer; the buttons render in front and keep their clicks. It lives on `customTitlebar` (a top-level ZStack child), not in the `sessionDetail`/`NSSplitView` subtree, so it doesn't perturb the split.

Per the control-API keep-in-sync convention, the action is also drivable headless via a new **`window.zoom`** command:
- `windowZoom` case in `ControlProtocol`
- `.windowZoom` dispatch arm in `ControlServer` → `WindowRegistry.zoom`
- `agtermctl window zoom <id>` subcommand
- agent-skill + rule docs updated (command count 46 → 47)

## Test plan

- `cd agtermCore && swift test` — 770/770 pass (incl. the `window.zoom` protocol round-trip).
- `make build` — succeeds.
- New e2e `ControlAPIUITests.testWindowZoom` — first zoom enlarges the window, a second restores it; passes.
- Full XCUITest suite — green except two pre-existing failures unrelated to this change (`SettingsUITests.testCompactToolbarTogglePersists`, reproduced on master with this change stashed; `FocusWorkspaceUITests…` flaked once, passed on re-run).
- Manual: double-click the header (empty area / title text / traffic-light strip) zooms and restores; single-click-drag across the full header moves the window; header buttons still click.

### DIFF
--- a/.claude/rules/control-api.md
+++ b/.claude/rules/control-api.md
@@ -86,7 +86,7 @@ paths:
   The skill is a REFERENCE/knowledge skill (both user-invocable via `/agterm` and model-triggered,
   `allowed-tools: Bash(agtermctl *)`; the agent-neutral `description` carries the trigger nouns since
   Codex may ignore the extra `when_to_use` field — unknown frontmatter is harmless),
-  authored at `agterm/Resources/agent-skill/` (`SKILL.md` overview + model + addressing + 47-command
+  authored at `agterm/Resources/agent-skill/` (`SKILL.md` overview + model + addressing + 48-command
   summary + the image-display helper + a troubleshooting/reporting pointer;
   `reference.md` full per-command detail + keymap format; `examples.md` agtermctl recipes;
   `troubleshooting.md` diagnosing the common problems (keymap editor, custom actions,
@@ -156,7 +156,7 @@ paths:
   exact `uuidString` (case-insensitive), or a git-style unique prefix.
   Zero prefix hits → `notFound` error, ≥2 → `ambiguous` error listing the candidates.
   `--target` defaults to `active`, so scripts rarely type an id and never for "the current one".
-- **Command catalog (47 commands):**
+- **Command catalog (48 commands):**
   - `tree`
   - `workspace.new`/`workspace.rename`/`workspace.delete`/`workspace.select`/`workspace.move`/`workspace.focus`
   - `session.new`/`session.close`/`session.select`/`session.rename`/`session.move`/`session.type`/`session.split`/`session.scratch`/`session.focus`/`session.resize`/`session.go`/`session.copy`/`session.search`/`session.status`/`session.flag`/`session.overlay.open`/`session.overlay.close`/`session.overlay.result`
@@ -164,7 +164,7 @@ paths:
   - `sidebar`/`sidebar.mode`/`sidebar.expand`/`sidebar.collapse`
   - `notify`
   - `font.inc`/`font.dec`/`font.reset`
-  - `window.new`/`window.list`/`window.select`/`window.close`/`window.rename`/`window.delete`/`window.resize`/`window.move` (see the Windows section)
+  - `window.new`/`window.list`/`window.select`/`window.close`/`window.rename`/`window.delete`/`window.resize`/`window.move`/`window.zoom` (see the Windows section)
   - `keymap.reload` (see the Keymap section)
   - `config.reload` (see the Settings section)
   - `theme.set`/`theme.list` (see the Theme picker section)
--- a/.claude/rules/windows.md
+++ b/.claude/rules/windows.md
@@ -146,7 +146,32 @@ never two bundles in one window.
   `AGTERM_SOCKET` is the path `ControlServer` *actually bound* (`ControlServer.boundSocketPath`,
   nil before bind → the var is omitted), so a test-overridden `AGTERM_CONTROL_SOCKET` and the injected
   env agree.
-- **`window.*` control additions (eight commands).**
+- **`window.zoom` (maximize-to-screen toggle, control + double-click-header GUI).** `WindowRegistry.zoom(_:)`
+  drives the standard `NSWindow.zoom(nil)` — toggles between the normal frame and the screen's visible frame
+  (NOT native fullscreen); a second call restores.
+  Unlike `resize`/`move` it has a GUI surface: a custom-titlebar SwiftUI view can't receive the OS
+  double-click handling, so `WindowControlArea` (an `NSViewRepresentable` behind `customTitlebar`'s decorative
+  regions in `ContentView.swift`) handles `mouseDown` — `clickCount == 2` runs the user's configured title-bar
+  action, else `performDrag` (also making the FULL header draggable, not just the native top band);
+  `mouseDownCanMoveWindow = false` so our handler sees the double-click.
+  The double-click honors the macOS **Desktop & Dock ▸ "Double-click a window's title bar to"** setting
+  (`AppleActionOnDoubleClick` in `NSGlobalDomain`, read LIVE per click): Zoom/Fill → `window.zoom(nil)`,
+  Minimize → `performMiniaturize`, "Do Nothing" → no-op; the key is absent until the user changes it from the
+  macOS default (Zoom), so an untouched system still zooms (the prior behavior).
+  So the GUI double-click is NOT always-zoom — only the `window.zoom` control command unconditionally zooms.
+  A UITest env override (`AGTERM_UITEST_DOUBLECLICK_ACTION`, read ahead of the system default) pins the action
+  so the gesture tests are hermetic regardless of the host setting; it rides the environment, not launch
+  arguments (FB11763863 — see `ui-tests.md`).
+  The header's decorative parts (the traffic-light spacer, the divider gap, the title text) opt out via
+  `.allowsHitTesting(false)` so their region falls through to the layer; the buttons stay in front.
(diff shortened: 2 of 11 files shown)


umputun (commented):
can you rebase? It conflicts with master after #35 merged into the same control-api files (`ControlProtocol.swift`, `ControlServer.swift`, `ControlProtocolTests.swift`) plus `ContentView.swift` and the skill docs. Textual overlaps, no real semantic clash, but it won't merge until rebased.

one design thing: the double-click is hardcoded to zoom (`window?.zoom(nil)`, `ContentView.swift:1187`), so it ignores the macOS "double-click a window's title bar to" setting (Desktop & Dock: fill / zoom / minimize / do nothing). A user who set minimize or do-nothing still gets a zoom, even though the docstring says it behaves "like a real title bar". Worth checking if `mouseDownCanMoveWindow = true` on the decorative regions gets you AppKit's drag plus the configured double-click action for free, instead of the custom `mouseDown` + `performDrag`. I'm not sure the double-click reaches a background view that way, so confirm before going down that route.

can you also add more test coverage? The e2e only checks zoom/restore through the control command. Nothing exercises the actual gestures: double-click on the empty header / title / traffic-light strip, single-click-drag moving the window, buttons still taking their own clicks. Those are the parts most likely to regress.

fkirill:
Thanks for the review — addressed all three.

**Rebased** onto current master (post-#35). Overlaps were textual; the only real merge calls were keeping #35's attention button alongside the title's `.allowsHitTesting(false)` fall-through in `ContentView`, and bumping the skill command count to 47.

**Double-click now honors the system setting.** You're right that "behaves like a real title bar" didn't hold while it hardcoded zoom. It now reads `AppleActionOnDoubleClick` (NSGlobalDomain) live per double-click: Zoom/Fill → `zoom(nil)`, Minimize → `performMiniaturize`, "Do Nothing" → no-op. The key is absent until you change it from the macOS default (Zoom), so an untouched system still zooms — no change for the common case. ("Fill" maps to zoom — closest standard `NSWindow` action; true Fill needs the newer tiling APIs, out of scope.)

On A vs B: I went with **A** (read the pref in our own `mouseDown`) over **B** (`mouseDownCanMoveWindow = true`, let AppKit run the configured action). Not pretending to be a macOS/Swift expert — just reasoning from the options. Two things pushed me to A:

1. **Coverage.** Our custom header is 48px (30px compact), but AppKit's native titlebar band is only the standard ~28px — that's the same reason the full header wasn't draggable before this PR. So even if B's events reached our layer, AppKit would only run the configured action within its own ~28px band; the bottom ~20px of the visible header would do nothing. A handles the full header uniformly.
2. **Reliability.** Our layer is a transparent `.background` view behind decorative regions that already opt out of hit-testing, so I wasn't confident B's double-click would reach it at all (your doubt too). A keeps every event in the handler we own — nothing lost — and leaves the proven drag/hit-test layering untouched.

The one thing A doesn't get free is any future action AppKit adds to that setting, but the four current values map explicitly in a few lines. Happy to switch to B if you'd rather — A just felt clearly lower-risk here.

**Gesture coverage added** (`ControlAPIUITests`):
- `testDoubleClickHeaderZoomsAndRestores` — real mouse double-click zooms; a second restores.
- `testHeaderButtonsStillReceiveClicksOverControlArea` — a header double-click doesn't trip a button, and `quick-terminal-toggle` still opens despite the layer behind it.
- `testDragHeaderMovesWindow` — single-click-drag on the full header moves the window.

The double-click/button tests are solid; the drag one is the borderline case you flagged, so it asserts a tolerant origin delta, not an exact position.

**Tests:** `swift test` green (785). Full `agtermUITests` suite: the 3 new gesture tests pass; everything green except `SettingsUITests.testCompactToolbarTogglePersists` and `testRestoreRunningCommandTogglePersists`, both of which fail identically on master with this change stashed (pre-existing, unrelated). Two others (`FlaggedView…`, `FocusWorkspace…`) flaked once and passed on re-run.

umputun (commented):
approach A is fine, thx for digging into B. Your read matches mine: the ~28px native band wouldn't cover the full header, and I doubted the double-click would reach the background view. Keeping A.

one thing the round-2 change introduced: the gesture tests aren't hermetic. The double-click tests assert the window grows, but `performTitlebarDoubleClickAction` now honors `AppleActionOnDoubleClick`, so on a machine set to Minimize (or None) the double-click minimizes (or no-ops) and the grow assert fails. It passes for you because your machine is on Zoom/unset; it fails on mine, which is set to Minimize. The button-coexistence test also double-clicks the header first, so it minimizes the window before clicking `quick-terminal-toggle`.

fix: pin the setting for these tests, e.g. force `-AppleActionOnDoubleClick Maximize` into the argument domain at launch. But watch the FB11763863 launch-argument caveat in `ui-tests.md` (args trip the window-no-present variant, which is why the sidebar sentinel goes through the env), so an app-honored env override may be safer. An explicit Minimize/None case would also lock in the new behavior.

nit: `.claude/rules/control-api.md:28` still says "46-command summary"; the catalog count four lines down is already 47.

rest of the keep-in-sync checks out (protocol/server/CLI/round-trip+e2e plus skill docs and count), and WindowControlArea sits on customTitlebar as a background so it doesn't perturb the split or grab first responder; `swift test` is green (785).

fkirill:
Yeah, that makes sense... Will fix it tomorrow.

Just a small comment about the command count in the .md files (46 vs. 47). You see, it might have worked for you while you were developing the prototype on your own. Maintaining this counter to be accurate in a multi-branch + rebase environment might cause constant pain/inaccuracy. Might be worth considering auto-populating it during build or dropping completely.
Honestly, I would probably look in the direction of autogenerating a portion of the LLM materials from the source code automatically (just place the annotations/fields that we need to create the content from around the commands) and then produce the templated .md as necessary. Kind of spec-first style.

umputun:
nah, don't worry about it, just a nit. I'll deal with this

umputun:
rebased onto master to clear the conflict that #38 introduced in the control-API files (it just merged). Your two commits are intact and still authored to you; I only resolved the catalog overlap in `control-api.md`: bumped the command count to 47 and kept both `window.zoom` and the new `session.status --sound` entries. Local `swift test` is green and the app builds, CI will re-run on the push. It was a force-push, so re-sync your local branch if you have unpushed work.

fkirill:
**Round 2 addressed — tests are hermetic now, commit `fca0974`.**

Synced onto your rebase (`2d343a0`); the count nit's already gone (47 in `control-api.md:28` from your resolve, nothing to do there).

**Non-hermetic gesture tests — fixed.** You were right: the double-click tests assert the window grows, but `performTitlebarDoubleClickAction` honors `AppleActionOnDoubleClick`, so on your Minimize machine they minimized instead and failed (they only passed here because I'm on unset/Zoom). The button-coexistence test had the same problem — it double-clicked the header first, so it minimized the window before clicking `quick-terminal-toggle`.

Fix: a new `AGTERM_UITEST_DOUBLECLICK_ACTION` env override that `performTitlebarDoubleClickAction` reads ahead of the system default. It rides the environment, not launch arguments — launch args trip FB11763863 (the `ui-tests.md` caveat you flagged), env sidesteps it. `setUp` pins `Maximize` for the gesture tests so they zoom deterministically regardless of the host setting.

**Added the Minimize/None lock-in.** `testDoubleClickHeaderHonorsNoneSetting` opts into `None` and asserts the double-click is a no-op — proves the gesture honors the setting instead of hardcoding zoom.

**Verified against your failing case.** Ran the four gesture tests with the host set to `Minimize` (`defaults write -g AppleActionOnDoubleClick Minimize`) and with it unset — green both ways. `swift test` green (794).

fkirill:
Rebased onto current master (`69878ca`) and force-pushed — the three commits are intact (`5b6bf2f` → `08015d7` → `662f9cf`), now linear on top.

Conflicts were doc-only, from the semantic-line-break reflow (#55) landing on `control-api.md` / `windows.md` while this branch still edited the old format; the code auto-merged clean. Resolved by taking master's reflowed docs and re-grafting this branch's additions:

- `window.zoom` added to the catalog, and the count bumped **47 → 48** across all three surfaces (catalog, the skill-install bullet, `SKILL.md`'s summary). Master had reached 47 *without* `window.zoom` (session.resize etc.), so it's +1, not the earlier 46→47.
- Re-authored the `window.zoom` bullet in `windows.md` in the honor-setting form, with the env-override note and all four gesture tests (incl. `testDoubleClickHeaderHonorsNoneSetting`) in its keep-in-sync audit.

Green on the rebased tree: `swift test` 817, app builds, and the four gesture tests + `testWindowZoom` pass.

PR #46, opened 2026-06-30
feat: read a session's buffer over the control API (session.text)

### LINKED ISSUE #34: feat: read a session's buffer (screen + scrollback) over the control API
`session copy` only returns the current selection. There's no way to grab the whole buffer.

I'd like something like `agtermctl session text --ansi` that dumps a session's screen + scrollback. Then I can pipe it into vim, less, or my own scripts in an overlay to search, copy, or do other custom things.

A few examples,:
- open overlay URLs from the visible text, list them in fzf, open the chosen one in the browser
- open vim in overlay with ANSI colors, so it's easy to select, copy, and search
- open overlay same as `cmd+e` but for file paths, so I can pick one and insert it

Plain text, optional ANSI colors. By default return visible screen, optionally full history. or N last lines

Happy to open a PR if you're up for it.

paul-nameless:
Closes #34.

`agtermctl session text` prints a session's buffer as plain text in `result.text`. `session.copy` only returns the current selection, so until now there was no way to grab the whole screen over the socket.

- default: the visible screen
- `--all`: screen plus scrollback
- `--lines N`: the last N lines
- `--pane left|right`: which pane (defaults to the focused one)

Plain text only. No `--ansi` support for now.

### DIFF
--- a/.claude/rules/control-api.md
+++ b/.claude/rules/control-api.md
@@ -86,7 +86,7 @@ paths:
   The skill is a REFERENCE/knowledge skill (both user-invocable via `/agterm` and model-triggered,
   `allowed-tools: Bash(agtermctl *)`; the agent-neutral `description` carries the trigger nouns since
   Codex may ignore the extra `when_to_use` field — unknown frontmatter is harmless),
-  authored at `agterm/Resources/agent-skill/` (`SKILL.md` overview + model + addressing + 49-command
+  authored at `agterm/Resources/agent-skill/` (`SKILL.md` overview + model + addressing + 50-command
   summary + the image-display helper + a troubleshooting/reporting pointer;
   `reference.md` full per-command detail + keymap format; `examples.md` agtermctl recipes;
   `troubleshooting.md` diagnosing the common problems (keymap editor, custom actions,
@@ -156,10 +156,10 @@ paths:
   exact `uuidString` (case-insensitive), or a git-style unique prefix.
   Zero prefix hits → `notFound` error, ≥2 → `ambiguous` error listing the candidates.
   `--target` defaults to `active`, so scripts rarely type an id and never for "the current one".
-- **Command catalog (49 commands):**
+- **Command catalog (50 commands):**
   - `tree`
   - `workspace.new`/`workspace.rename`/`workspace.delete`/`workspace.select`/`workspace.move`/`workspace.focus`
-  - `session.new`/`session.close`/`session.select`/`session.rename`/`session.move`/`session.type`/`session.split`/`session.scratch`/`session.focus`/`session.resize`/`session.go`/`session.copy`/`session.search`/`session.status`/`session.flag`/`session.background`/`session.overlay.open`/`session.overlay.close`/`session.overlay.result`
+  - `session.new`/`session.close`/`session.select`/`session.rename`/`session.move`/`session.type`/`session.split`/`session.scratch`/`session.focus`/`session.resize`/`session.go`/`session.copy`/`session.text`/`session.search`/`session.status`/`session.flag`/`session.background`/`session.overlay.open`/`session.overlay.close`/`session.overlay.result`
   - `quick`
   - `sidebar`/`sidebar.mode`/`sidebar.expand`/`sidebar.collapse`
   - `notify`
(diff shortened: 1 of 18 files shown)


umputun (commented):
needs a rebase before merge, it's conflicting on master (8 commits behind, and `control-api.md` / `ControlServer.swift` / `GhosttySurfaceView.swift` / `Commands.swift` all moved on both sides). While rebasing, note the `Session.subcommands` line in `Commands.swift` with `Text.self` added is 222 chars, over the swiftlint `line_length` 200 limit, and master's CI runs `swiftlint --strict` (added in #57, after your branch point), so it fails the lint job until you reflow that array.

a few things worth a look:

1. server-side arg validation. The CLI `validate()` rejects `--all`+`--lines` and `--lines <= 0`, but `readText` passes `args.all`/`args.lines` through unchecked. A raw socket client sending `lines: 0` or `-1` gets the full scrollback instead of an error (non-nil `lines` forces SCREEN mode, then `n <= 0` skips the trim). Worth re-checking server-side, like the "invalid mode" arms do.

2. omitted `--pane` reads `session.activeSurface`, but when a scratch terminal or overlay covers the session the visible surface is `topmostSurface`. The search arm and `AppActions.focusedSurface` both resolve `coverIsScratch ? topmostSurface : activeSurface`. So `session text` with no `--pane` on a session showing its scratch returns the hidden pane underneath, not what's on screen.

3. `readScreenText` returns nil both for a failed/torn-down read and for a genuinely blank screen, and `readText` maps every nil to `ok` + empty string. So a caller can't tell a read failure from an empty terminal. The empty-is-ok case is documented as deliberate, but is swallowing an actual read failure intended too, or should that be an error?

minor:
- `writeResponse` is a blocking `write()` with no `SO_SNDTIMEO` (only the read side sets `SO_RCVTIMEO`). Responses were small before, `--all` can be multi-MB now, so a client that stops reading could park the accept loop. Only `agtermctl` talks to it and it always reads, so low risk, but a send timeout would close it off.
- `--lines N` only strips one trailing newline, so on a non-scrolled screen (prompt at row 1, blank grid below) it can return blank padding rather than the last N content lines.

paul-nameless:
Addressed all comments and rebased

umputun (commented):
re-checked after the rebase. All 5 round-1 points are fixed, each with its own test - server-side arg validation, the shared `onScreenSurface` pane resolution, the blank-vs-failed-read split, and the `--lines` trailing-blank trim. `swift test` green, swiftlint clean against the base.

still needs a rebase (conflicting, 5 behind).

one more on the write timeout, my round-1 minor #4. `SO_SNDTIMEO` closes the case where a client stops reading outright, but there's no cumulative write deadline like `readDeadlineSeconds` has on the read side. A slow-drip reader draining a multi-MB `--all` a few bytes per interval keeps each `write()` making progress, so the 5s per-write timeout never trips and the accept loop stays parked. Can you add an overall write deadline in `writeResponse` mirroring `readLine`'s deadline? Small, and symmetric with what's already there.

not blocking, optional: `Session.onScreenSurface` has no direct unit test (covered only indirectly), and the blank-screen `ok+""` path is untested.

paul-serenity:
addressed suggestions and rebased

umputun (approved):
lgtm, all the round-2 items check out. Write deadline now bounds total write time (mirrors `readDeadlineSeconds`), count reconciled to 50, and the swiftlint growth is handled by splitting the e2e into `SessionTextUITests` rather than bumping limits. Also ran it on a dev build: default/`--lines`/`--all` reads, full scrollback (1600+ lines, nothing dropped), and reading an unfocused session all work.

two non-blocking follow-ups, both from Copilot, worth a look but not merge-blockers:

1. the `--pane` help and the `ControlArgs.pane` doc say "defaults to the focused pane", but the default is actually the on-screen surface (scratch when it covers, else the focused pane). The behavior is right, just the wording is stale. Worth rewording to match.

2. client-side `SocketClient.readLine` doesn't retry `EINTR` like the server side does, and its doc says nil-on-EOF while it actually returns the partial buffer. Harmless for a short-lived CLI (fails loudly, no corruption), but the `EINTR` retry would make it symmetric with the server.

PR #48, opened 2026-06-30
feat: show unseen notification count as a Dock icon badge

### LINKED ISSUE #47: feat: show unseen notification count as a Dock icon badge (macOS 26 adaptive-icon workaround)
## Summary

Surface the app-wide unseen-notification total on the **Dock icon** as a count badge — the same `Session.unseenCount` the sidebar's red pills already track, summed across every open window. The sidebar shows it per-session (and as a per-workspace roll-up), but there's no glanceable indicator when the window is hidden/minimized or you're in another app; the Dock badge is the conventional macOS surface for that.

Behavior: capped at `99+`, cleared at zero, gated by the existing **"Show notification badges"** Settings toggle (one switch governs both the sidebar pill and the Dock badge).

## The macOS 26 gotcha (why this is non-trivial)

The obvious implementation — `NSApp.dockTile.badgeLabel = "3"` — **silently does nothing** for agterm. The badge is set (the value round-trips) but the system never paints it.

Isolated it to the **adaptive Icon Composer icon**:

- A throwaway minimal app with a *generic* icon and `dockTile.badgeLabel` set → badge paints fine (so `badgeLabel` works on macOS 26.5.1 in general).
- agterm, whose Dock icon is the live LaunchServices-rendered adaptive `AppIcon.icon` → label never paints, **even after forcing a static `NSApp.applicationIconImage`**.

Conclusion: on macOS 26, `dockTile.badgeLabel` is not composited over a live adaptive `.icon`.

## Chosen approach

Draw the count pill **into the icon ourselves** and set that as `applicationIconImage` while a badge is present; restore the live adaptive icon (`nil`) when the unseen total returns to zero.

- **Trade-off:** while a badge is up the Dock shows the *static* (non-adaptive) icon — it won't track light/dark/tint during that window. The adaptive rendering returns the moment the count hits zero (the common state), so the static icon only appears when there's actually something to show.
- This is a deliberate, badge-scoped exception to the project's "never set `applicationIconImage`" rule (which exists to avoid permanently freezing the adaptive icon).

## Implementation notes

- `WindowLibrary.totalUnseenCount` (host-free, unit-tested) — sum of `unseenCount` across every open window's sessions.
- `DockBadgeController` (`@MainActor` singleton, next to `NotificationManager`) — Observation re-registration (`withObservationTracking`) drives `apply()` on every notification bump / focus-clear / session add-remove, with explicit `refresh()` pokes on the two unobservable store-drops (window `willClose`, `window.close`) and an `.agtermAppearanceChanged` observer for the (non-`@Observable`) toggle flip.
- Pure derived chrome — no new control command (the underlying `unseenCount` is already driven by `notify` / `session.select`).

PR incoming.

vnazarenko:
Closes #47.

## What

Show the app-wide unseen-notification total on the **Dock icon** as a count badge — the same `Session.unseenCount` the sidebar pills track, summed across every open window. Capped at 99, cleared at zero, gated by the existing **"Show notification badges"** Settings toggle (one switch governs both the sidebar pill and the Dock badge).

## How

Uses the modern **`UNUserNotificationCenter.setBadgeCount`**, which renders the pill over the live adaptive Icon Composer icon with full light/dark/tint/clear adaptivity — no `applicationIconImage`, no icon override. `NotificationManager` now requests `[.alert, .badge]` (the legacy `dockTile.badgeLabel` is silently suppressed without the badge grant).

## Changes

- `agtermCore`: `WindowLibrary.totalUnseenCount` (host-free, unit-tested).
- `DockBadgeController` (`@MainActor` singleton): Observation re-registration drives `apply()` on every notification bump / focus-clear / session add-remove; explicit `refresh()` pokes on the two unobservable store mutations — window **close** (`willClose` teardown in `ContentView`, `window.close` in `ControlServer`) and window **reopen** (`ContentView.resolveStore` after `loadStore`); `.agtermAppearanceChanged` observer for the (non-`@Observable`) toggle flip.
- `clear()` on `applicationWillTerminate` zeroes the OS-level badge on quit (it outlives the process, and `unseenCount` is ephemeral, so otherwise a quit with unseen > 0 leaves a stale count pinned on the Dock icon).
- `NotificationManager` requests `.badge`; controller wired in the scene `.task`.
- `.claude/rules/notifications.md` documents the subsystem.

Keep-in-sync **exempt**: pure derived chrome — `unseenCount` is already driven by `notify` / `session.select`, so there's nothing new to expose over the control socket.

## Note on existing installs

`setBadgeCount` needs the per-app **Badges** setting enabled. Fresh authorizations get it on by default; a bundle previously granted `[.alert]`-only keeps `badgeSetting == .disabled` until Badges is toggled on once in System Settings (standard UN behavior — no API to force it).

## Testing

- `swift test` (agtermCore): **825 tests pass**, incl. `totalUnseenCount`.
- App builds (Debug); swiftlint clean.
- Verified end-to-end on macOS 26.5.1: the pill renders over the live adaptive icon (icon stays adaptive, no static swap) with Badges enabled. Dock-tile rendering isn't XCUITest-coverable, so it's manually verified like the other AppKit chrome.

### DIFF
--- a/.claude/rules/notifications.md
+++ b/.claude/rules/notifications.md
@@ -59,6 +59,30 @@ paths:
   reloads the rows).
   This gates ONLY the red count pill — the agent-status glyph (drawn just left of it by `StatusIconView`)
   is always on, never gated by this.
+- **Dock badge (`DockBadgeController`, app target).**
+  A `@MainActor` singleton (next to `NotificationManager`, wired in the scene `.task`) that shows the
+  app-wide unseen total on the Dock icon.
+  The total is the host-free `WindowLibrary.totalUnseenCount` (sum of every open window's sessions'
+  `unseenCount`, unit-tested); it's gated by the SAME `notificationBadgeEnabled` toggle as the sidebar
+  pill (badge count is 0 when off).
+  **Uses `UNUserNotificationCenter.setBadgeCount(_:)`, NOT `NSApp.dockTile.badgeLabel`.** For agterm the
+  legacy dock-tile label is silently suppressed — the value sets and persists on the tile but the Dock
+  never draws the pill — because it needs the `.badge` authorization option; `NotificationManager.start`
+  now requests `[.alert, .badge]`.
+  The modern UN badge renders correctly over the LIVE adaptive Icon Composer icon with no loss of
+  light/dark/tint/clear adaptivity, so there is NO `applicationIconImage` override.
+  Cleared to 0 on quit (`DockBadgeController.clear()` from `applicationWillTerminate`) — the OS badge
+  outlives the process and `unseenCount` is ephemeral, so without it a quit with unseen > 0 would leave a
+  stale count pinned on the Dock icon (the `willClose` poke can't do it: `isTerminating` makes `closeWindow`
+  no-op, so the total recomputes unchanged).
+  Reactivity is the Observation re-registration pattern (`apply()` reads `totalUnseenCount` inside
+  `withObservationTracking` and re-arms on change), with explicit `refresh()` pokes on the two
+  unobservable store mutations — a window CLOSE (`willClose` teardown in `ContentView`, `window.close` in
+  `ControlServer`) AND a window REOPEN (`ContentView.resolveStore` after `loadStore`, so a reopened
+  window's bumps aren't stale) — and an `.agtermAppearanceChanged` observer for the (non-`@Observable`)
+  toggle flip.
+  Keep-in-sync EXEMPT — pure derived chrome (`unseenCount` is already driven by `notify` / `session.select`),
+  nothing new to drive over the socket.
 - **Agent-status glyph.**
   Mirrors the `notify-badge` cell pattern (see the Control API `session.status`).
   `StatusIconView` (an `NSImageView` sibling of `BadgeView` in `WorkspaceSidebar`) draws the row's tinted
--- a/agterm/ContentView.swift
+++ b/agterm/ContentView.swift
@@ -1099,7 +1099,10 @@ private struct WindowAccessor: NSViewRepresentable {
                     library.closeWindow(windowID)
                     // closing a window drops its (unobserved) store, so the Dock badge's observation
                     // tracking won't fire — refresh it explicitly so the unseen total drops this window's.
-                    DockBadgeController.shared.refresh()
+                    // guard on isTerminating: on quit the willClose fires after applicationWillTerminate's
+                    // clear(), and closeWindow no-ops (stores stay loaded), so an unguarded refresh would
+                    // recompute the still-positive total and re-pin the badge clear() just zeroed.
+                    if !library.isTerminating { DockBadgeController.shared.refresh() }
                 }
             }
             titlebarObservers.append(closeToken)
--- a/agterm/ContentView.swift
+++ b/agterm/ContentView.swift
@@ -91,6 +91,9 @@ struct ContentView: View {
         }
         store = resolved
         resolvedID = id
+        // reopening a window loads an @ObservationIgnored store, which the Dock badge's observation can't
+        // see (symmetric to the willClose close poke) — recompute so a reopened window's unseen total counts.
+        DockBadgeController.shared.refresh()
     }
 
     /// The window id this view adopts: normally the next id in the library's claim queue. If the
@@ -1094,6 +1097,9 @@ private struct WindowAccessor: NSViewRepresentable {
                         session.scratchSurface?.teardown()
                     }
                     library.closeWindow(windowID)
+                    // closing a window drops its (unobserved) store, so the Dock badge's observation
+                    // tracking won't fire — refresh it explicitly so the unseen total drops this window's.
+                    DockBadgeController.shared.refresh()
                 }
             }
             titlebarObservers.append(closeToken)
--- a/agterm/Control/ControlServer.swift
+++ b/agterm/Control/ControlServer.swift
@@ -1332,6 +1332,8 @@ final class ControlServer {
             if library.isOpen(id) {
                 library.closeWindow(id)
             }
+            // dropping a store is unobserved, so poke the Dock badge to drop this window's unseen total.
+            DockBadgeController.shared.refresh()
             return ControlResponse(ok: true, result: ControlResult(id: id.uuidString))
         }
     }
--- a/agterm/Notifications/DockBadgeController.swift
+++ b/agterm/Notifications/DockBadgeController.swift
@@ -1,0 +1,93 @@
+import agtermCore
+import Foundation
+import Observation
+import UserNotifications
+
+/// Drives the Dock icon's unseen-notification badge from the app-wide total of unseen terminal
+/// notifications — the same `Session.unseenCount` the sidebar's red pills show, summed across every open
+/// window by `WindowLibrary.totalUnseenCount`. Capped at 99 (mirrors the sidebar pill's `99+`), cleared at
+/// zero, gated by the SAME `GhosttyApp.notificationBadgeEnabled` ("Show notification badges") Settings
+/// toggle — one switch governs both surfaces.
+///
+/// **Uses the modern UserNotifications badge, NOT `NSApp.dockTile.badgeLabel`.** For agterm the legacy
+/// dock-tile label is silently suppressed — the value sets and persists on the tile, but the Dock never
+/// draws the pill — because it requires the `.badge` authorization option (which `NotificationManager` now
+/// requests alongside `.alert`). `UNUserNotificationCenter.setBadgeCount(_:)` renders the count correctly
+/// over the LIVE adaptive Icon Composer icon with no loss of light/dark/tint/clear adaptivity, so the badge
+/// is purely the number — no self-drawn icon, no `applicationIconImage` override.
+///
+/// `@MainActor` singleton like `NotificationManager`. Reactivity uses the Observation re-registration
+/// pattern: `apply()` reads the observable inputs inside `withObservationTracking` and re-arms itself on
+/// the next change, so a notification bump, a focus/select clear, and a session add/remove all refresh the
+/// badge. Changes that are NOT observable are poked explicitly: a window CLOSE drops, and a window REOPEN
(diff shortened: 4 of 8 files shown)


umputun (changes requested):
the macOS 26 premise this is built on doesn't hold. The native `dockTile.badgeLabel` does composite over an adaptive Icon Composer `.icon`, with no loss of adaptivity.

Mimestream uses the same `.icon` format and shows its unread count pill while its icon keeps adapting across light/dark. Toggle the system appearance with Mimestream unread and the pill stays while the icon re-renders. So the native badge works over an adaptive `.icon`, nothing here to work around.

that makes `applicationIconImage` the wrong route. It swaps the live adaptive icon for a static bitmap while a badge is up, dropping the light/dark/tint/clear adaptation exactly when the badge is on screen. Losing icon adaptivity is a hard no, and the native badge already does the job without it.

please switch to plain `NSApp.dockTile.badgeLabel` (set the count, `nil` at zero, then `dockTile.display()`). That removes `DockBadgeController`'s self-drawn `badgedIcon`, the `applicationIconImage` override, the `iconOverridden` latch, and the `AppDelegate` exception. The host-free `WindowLibrary.totalUnseenCount` and its test stay, and the observation wiring mostly carries over.

one note on verification so you don't chase a ghost: setting `badgeLabel` directly took (read it back, held `"3"`) but drew no pill in an isolated dev build (`.debug` bundle id, launched via `open -n` from DerivedData). That reads as a LaunchServices/dev-build artifact rather than the icon, and is likely what produced the "even after forcing a static `applicationIconImage`" result. Verify `badgeLabel` in a properly installed/registered build before concluding it doesn't render. Mimestream is proof it does.

separate and smaller: a window reopened via `WindowLibrary.loadStore` re-arms no observed state (it only writes the `@ObservationIgnored stores` dict) and gets no `refresh()` poke, so an unseen bump in a reopened window can leave the Dock total stale until some other observed change fires. The close path pokes `refresh()`, the reopen path doesn't. Worth a symmetric poke, or at least a recheck, in the rework.

umputun:
follow-up to the review above, and a correction to it. I chased this down empirically. The premise is wrong as I said, but my "looks like a dev-build artifact, verify in an installed build" caveat was also wrong, scrap that part.

the badge works over the live adaptive icon with no adaptivity loss. A minimal AppKit app and a minimal SwiftUI app both badge fine via the same dev launch, so it's not a dev-build / registration / install-location thing.

the real issue is the API. For agterm the legacy `NSApp.dockTile.badgeLabel` is silently suppressed: the value sets and persists on the dock tile, but the Dock never draws the pill. The modern UserNotifications badge renders fine. `NotificationManager` requests `requestAuthorization(options: [.alert])` (NotificationManager.swift:35), no `.badge`, so the badge permission is never granted.

fix, verified end to end (pill on the live adaptive icon, light/dark/tint intact, no `applicationIconImage`):

1. `NotificationManager` → `requestAuthorization(options: [.alert, .badge])`.
2. set the badge via `UNUserNotificationCenter.current().setBadgeCount(library.totalUnseenCount)`, drop `dockTile.badgeLabel` / `applicationIconImage` entirely. That removes `badgedIcon`, the `iconOverridden` latch, the `applicationIconImage` override, and the `AppDelegate` exception.
3. keep `WindowLibrary.totalUnseenCount` and the observation wiring. `setBadgeCount` takes an Int, so the `99+` cap becomes `min(count, 99)` or just the raw count.

net: a handful of lines on top of `totalUnseenCount`, full adaptivity, no convention exception.

vnazarenko:
Thanks for chasing this down — you're right, and I've reworked the PR to your prescription. Rebased onto latest `master` (v0.4.2) and force-pushed.

**Done:**

- ✅ Dropped the whole self-drawn-icon route — `badgedIcon`, the `iconOverridden` latch, the `applicationIconImage` override, and the `AppDelegate` exception note are all gone. No adaptivity loss.
- ✅ `NotificationManager.start` now requests `[.alert, .badge]`.
- ✅ Badge set via `UNUserNotificationCenter.current().setBadgeCount(totalUnseenCount)`; `dockTile.badgeLabel` removed entirely.
- ✅ `WindowLibrary.totalUnseenCount` + its unit test stay; observation wiring carried over. `setBadgeCount` takes an `Int`, so the `99+` cap is gone — it's the raw count now (happy to `min(_, 99)` if you'd prefer a ceiling).
- ✅ **Reopen poke:** added `DockBadgeController.shared.refresh()` in `ContentView.resolveStore()` right after `loadStore`, symmetric to the `willClose` / `window.close` close pokes — so an unseen bump in a reopened window isn't left stale.

`swift test` 825 pass, app builds.

**One nuance worth flagging** (verified via `getNotificationSettings`, not asking you to re-verify — just a heads-up): `setBadgeCount` needs `badgeSetting == .enabled`, and on a bundle that was **already** authorized `[.alert]`-only, adding `.badge` to `requestAuthorization` does **not** re-enable it — I measured `authorizationStatus = .authorized`, `badgeSetting = .disabled`, and it stayed disabled across a relaunch with the new options. Fresh authorizations get badges on by default; existing users (who granted alert-only on a prior build) would keep it disabled until they flip **Badges** on in System Settings. There's no API to force-enable it. Fine to ship as-is (standard UN behavior) — just noting it in case you want to handle the upgrade path. Your end-to-end verification was presumably a fresh grant, which is the path that "just works".

vnazarenko:
Circling back with an independent visual confirmation (macOS 26.5.1): with Badges enabled, `setBadgeCount` renders the count pill over the **live adaptive icon** — I watched the icon stay adaptive with the pill up, no static swap. 👍

Also added the `min(count, 99)` cap you mentioned, and rebased onto latest `master`. The only outstanding caveat is the upgrade-path one above (a bundle previously granted `[.alert]`-only keeps `badgeSetting == .disabled` until Badges is toggled on once — no API to force it). Happy to add a one-time Settings deep-link nudge if you want to smooth that for existing users, otherwise it's fine as standard UN behavior.

Ready for another look whenever — thanks for the precise diagnosis.

umputun (changes requested):
lgtm overall, ran it locally (swift test 825 pass including the new totalUnseenCount test, swiftlint clean, Debug build green) and CI is green now too. One thing worth fixing before merge:

**clear the Dock badge on quit.** `applicationWillTerminate` never zeroes it, and `closeWindow` is a no-op once `isTerminating` is set, so the willClose `refresh()` poke just recomputes the same positive total from the still-open stores. `setBadgeCount` writes an OS-level badge that outlives the process while `unseenCount` isn't restored, so quitting with unseen > 0 leaves a stale count on a pinned Dock icon until the next launch clears it. One line:

```swift
func applicationWillTerminate(_: Notification) {
    UNUserNotificationCenter.current().setBadgeCount(0)
    // ...existing teardown
}
```
(or a small `DockBadgeController.clear()` if you'd rather keep badge logic in one place)

a few notes:

- the PR description still describes the old `applicationIconImage` approach (Copilot flagged the same on notifications.md:74), worth updating so the squash commit doesn't inherit it.
- Copilot's `assumeIsolated` flags (DockBadgeController.swift:48/79) are fine to leave: the observer uses `queue: .main` so it always runs on the main thread, and it matches the existing NotificationCenter pattern in ContentView/ControlServer. The onChange double-defer is a harmless extra hop, not a trap.
- upgrade path (existing alert-only grants keep badges disabled): fine as-is, standard UN behavior, no need for the settings deep-link nudge.

vnazarenko:
Good catch on the stale-pinned-badge — fixed. Pushed:

- **Clear on quit:** added `DockBadgeController.clear()` (single `setBadgeCount(0)`) called from `applicationWillTerminate`, before `isTerminating` is set. Kept it in the controller so all badge writes live in one place, as you suggested.
- **PR description** rewritten to the `setBadgeCount` approach (no more `applicationIconImage`), and dropped the "PR #48" reference in `notifications.md:74` so the rule stays evergreen (also folded in the clear-on-quit note).
- Left the `assumeIsolated` / double-defer as-is per your call, and skipped the deep-link nudge.

`swift test` 825 pass, Debug build green. Ready for another look.

umputun (commented):
round-2 items are all in - clear on quit, description reworked to `setBadgeCount`, `#48` dropped from notifications.md, `min(99)` cap. One problem with the clear-on-quit itself though.

`clear()` in `applicationWillTerminate` (agtermApp.swift:874) runs before `isTerminating = true` (agtermApp.swift:877), and the per-window `willClose` fires after that during quit and calls `DockBadgeController.shared.refresh()` unconditionally (ContentView.swift:1102). `closeWindow` no-ops under `isTerminating`, so the stores stay loaded, so `refresh()` recomputes the still-positive `totalUnseenCount` and re-pins the badge `clear()` just zeroed. So ⌘Q with unseen > 0 still leaves the stale count until the next launch, which is the case round 2 was meant to fix. (`clear()` is fire-and-forget too, so there's no guarantee `setBadgeCount(0)` even flushes before exit.)

guard the poke the same way `closeWindow` guards itself:

```swift
library.closeWindow(windowID)
if !library.isTerminating { DockBadgeController.shared.refresh() }
```

worth a manual ⌘Q-with-unseen check either way, it's not test-coverable (no accessibility hook on the dock badge).

minor, not blocking: the reopen `refresh()` in `resolveStore` (ContentView.swift:96) is effectively a no-op. `unseenCount` is ephemeral, so a reopened window's sessions start at 0 and only move on a new notification, which is observable and self-fires anyway. Harmless, the comment just overstates what it does.

umputun:
about to cut a new release and wanted this in, so I pushed the one-liner myself (a3f9763). Guards the `willClose` `refresh()` on `isTerminating` so quit's `clear()` isn't undone. Rest looks good, thx.

vnazarenko:
Thanks for catching the ordering bug and landing the guard — you're right, I zeroed the badge in `applicationWillTerminate` but the per-window `willClose` `refresh()` fired afterward and recomputed the still-positive total (stores stay loaded under `isTerminating`), re-pinning exactly what `clear()` cleared. Gating the poke on `isTerminating` is the correct fix. Appreciate you pushing it in for v0.5.0. 🙏

PR #56, opened 2026-06-30
Add fish shell integration for agent-status hooks

korjavin:
...


umputun (commented):
nice port. Tested it here (fish 4.8.0) - `fish -n` clean, and driving the preexec/prompt handlers with a stub flips active/idle right (no-op on a non-agent command), though I didn't do a full live round trip inside agterm.

one thing to fix before merge: `appendShellRC` gained a `scriptName` param but nothing tests it. Add a case in `AgentHooksInstallTests` passing `scriptName: "shell/integration.fish"` and asserting the emitted `source` line.

a question: the installer now creates `~/.config/fish/config.fish` (and the `fish/` dir) for everyone who runs Install Agent Status Hooks, including non-fish users. It's consistent with how we already create `.zshrc`/`.bashrc`, and it no-ops outside agterm, so it's harmless. Still, materializing a config for a shell the user may not have is a bit much. Should we gate it on `~/.config/fish/` already existing? Your call.

couple of nits, not blockers:
- `integration.fish` leaves a literal `/..` in `AGTERM_AGENT_BIN` (`(dirname (status filename))/..`); the bash version normalizes with `cd .. && pwd`. Resolves fine at exec, just cosmetic.
- the `"shell/integration.fish"` literal is hardcoded in `AgentHooksInstaller`, could sit next to `integrationRelativePath` in `AgentHooksInstall` for consistency.

umputun (approved):
all addressed - test added (passes, full `AgentHooksInstall` suite green), fish wiring now gated on `~/.config/fish/` existing, and the path cleanup is in.

lgtm, thx

PR #62, opened 2026-07-01
feat: show the configured shortcut in toolbar and sidebar tooltips

taras-mrtn:
## What

Toolbar and sidebar-bar button tooltips now include the button's current keyboard shortcut in parentheses — hovering **Toggle Sidebar** shows `Toggle Sidebar (⌃⌘S)`, **Split** shows `Split right (⌘D)`, etc. A button whose action has no configured shortcut (e.g. the flagged-view toggle) shows just its label, unchanged.

## Why

agterm is keyboard-first — every action has a shortcut — but the toolbar/sidebar buttons were the one surface that never revealed theirs. To learn or confirm a button's key you had to open the action palette or hunt through the menu bar. Both of *those* surfaces already show the shortcut as a macOS glyph; the tooltips were the gap. Surfacing it on hover teaches the shortcut in place, and because the tooltip reads the live keymap, a user who rebinds a built-in in `keymap.conf` sees the new chord rather than a stale default — reinforcing the customization story.

## How

The action→glyph resolution the action palette already used (`AppActions.paletteHint`) is lifted into a single host-free resolver in `agtermCore`, so the palette and the tooltips share one source of truth and can't drift:

- `Keymap.glyphHint(for:) -> String?` — the effective chord (`equivalent(for:)` = user override else shipped `defaultChord`) rendered via `Chord.glyphString`, falling back to `BuiltinAction.arrowGlyphFallback` for the six arrow-bound actions (whose shortcuts can't round-trip through `Chord`). `nil` = no shortcut, so the caller shows nothing.
- `BuiltinAction.arrowGlyphFallback: String?` — the hardcoded arrow glyphs (`⌥⌘↑` …) lifted out of the app target (they were duplicated inside `paletteHint`); `nil` for every other action. The display counterpart of the menu's hardcoded arrow `.keyboardShortcut`.
- `AppActions.paletteHint` → `AppActions.shortcutGlyph` — the shared entry point now, delegating to `glyphHint`; the palette call sites are renamed accordingly (mechanical).
- `WindowContentView.helpHint(_:_:)` (`ContentView`) composes `"<label> (<glyph>)"` when a glyph exists, else the bare label — wired into the 7 toolbar/sidebar buttons that map to a `BuiltinAction`. Buttons with no keymap action ("Clear focus", the search-bar controls) are left untouched, so they correctly show no shortcut.

### Keep-in-sync (control API): exempt

I checked whether this needs a control command, and it doesn't — it's the "pure visual chrome with nothing to drive" exemption (like `reveal`). It adds no user action to `AppActions`/`AppStore` and no new state: `shortcutGlyph`/`helpHint` are read-only formatters over the existing keymap, and a tooltip is hover-only OS chrome with nothing to invoke headless. The control command count stays 48; the keymap format, the window/workspace/session model, and the agent-skill docs are unchanged.

## Test plan

- `cd agtermCore && swift test` — **830/830 pass**, incl. 7 new cases: `Keymap.glyphHint` (default→glyph, override wins, arrow fallback, override beats arrow fallback, `nil` when unconfigured) and `BuiltinAction.arrowGlyphFallback` (covers the six arrows, `nil` otherwise).
- `swiftlint --config .swiftlint.yml` on the changed files — **0 violations**.
- `make build` — **succeeds**.
- Rebased onto current `master` (post-v0.4.1). The only merge-relevant overlap was keeping #35's "Show Attention" palette item on the renamed `shortcutGlyph`.
- Manual: hover each toolbar/sidebar button — bound ones show the glyph and track a `keymap.conf` rebind after Reload Keymap; the flagged-view toggle (no default shortcut) shows only its label.

Note: `helpHint` is a trivial formatter over the unit-tested `glyphHint`, and the SwiftUI `.help()` tooltip itself isn't practically XCUITest-observable, so it's covered by the resolver unit tests plus manual hover. Happy to extract the formatter into `agtermCore` for an explicit host-free test if you'd prefer belt-and-suspenders coverage.

### DIFF
--- a/.claude/rules/menu-actions.md
+++ b/.claude/rules/menu-actions.md
@@ -211,6 +211,15 @@ paths:
   The visible list is a `@State` array recomputed on query/mode change — NOT a computed property — so
   the rendered rows and the Enter target can't drift out of sync; results sort by score then title. ⌃P
   opens the session switcher, ⌃⇧P the action palette (the session/action shortcut split is deliberate).
+- **Built-in shortcut hints are one resolver, shared by the palette AND the toolbar/sidebar tooltips.**
+  `AppActions.shortcutGlyph(for:)` (formerly `paletteHint`) → the host-free `Keymap.glyphHint(for:)`
+  (`equivalent(for:)?.glyphString ?? BuiltinAction.arrowGlyphFallback`, nil = no shortcut) renders an
+  action's CURRENT chord as macOS glyphs (`⌃⌘S`), tracking a `keymap.conf` rebind live.
+  `paletteActions()` reads it for the right-aligned palette hint; `WindowContentView.helpHint(_:_:)`
+  appends `" (<glyph>)"` to the `.help(…)` tooltip of the 7 `BuiltinAction`-backed toolbar/sidebar buttons
+  (a button with no configured shortcut shows just its label). One resolver so the two surfaces can't
+  drift — the display analogue of the `defaultChord`-single-source-of-truth rule. Tooltip text is pure
+  visual chrome, so it is control-API keep-in-sync EXEMPT (no command, nothing to drive headless).
 - **Attention list (the `.attention` palette mode).**
   A fourth `PaletteMode` (`.attention`, placeholder "Go to a session that needs attention…") lists the
   window's NON-IDLE sessions — broader than the ⌃⌥↑/↓ attention-NAV, which steps only `needsAttention`
--- a/agterm/AppActions.swift
+++ b/agterm/AppActions.swift
@@ -415,74 +415,63 @@ final class AppActions {
 
     // MARK: - Command palettes
 
-    /// The palette shortcut hint for a rebindable built-in: its currently-bound chord rendered as macOS
-    /// menu glyphs (so it tracks rebinds and reads like the menu equivalent), or `nil` when the action
-    /// has no chord. The arrow-bound actions fall back to their hardcoded arrow glyph when no override is
-    /// set — `defaultChord` is nil for them (arrows can't round-trip through `parseKeybind`), mirroring
-    /// `agtermApp.arrowShortcut(for:)`.
-    private func paletteHint(for action: BuiltinAction) -> String? {
-        if let chord = settingsModel?.keymap.equivalent(for: action) {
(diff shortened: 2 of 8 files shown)


umputun (commented):
lgtm overall. Verified locally: swift test 830/830, swiftlint --strict clean. Couple of asks:

1. the attention bell (`show_attention`, ⌃⇧I) is the one shortcut-bearing toolbar button still on a static `.help(...)`. Mind routing it through `helpHint(..., .showAttention)` too? Same two-state string as split/scratch, so it's the odd one out now.

2. minor doc nit while you're in there: the `PaletteItem.shortcut` comment (`Palette.swift:12-13`) says "four arrow-bound actions" (there are six) and "both render in kitty syntax", but built-ins render as glyphs via `glyphString` and only custom commands use kitty. The rename on the line above made it contradict itself.

3. not a blocker, just flagging: `helpHint` reads `actions.settingsModel` at render, and since `AppActions` isn't `@Observable` and settingsModel is wired in the scene `.task`, the non-arrow tooltips can paint as bare labels on the first render until the next store change recomputes them. You'd basically never see it in practice (hover delay + constant relayout at launch), so your call whether to harden it (wire settingsModel before first render, or read via an observable).

taras-mrtn:
Addressed 1 and 2 in the follow-up commit:

**1.** Attention bell now routes through `helpHint(…, .showAttention)` — same two-state string as split/scratch, so it picks up ⌃⇧I like the rest.

**2.** Fixed the `PaletteItem.shortcut` comment: built-ins render as macOS glyphs (only custom commands use kitty syntax), and it's six arrow-bound actions, not four.

**3.** I'd leave this as-is. `helpHint` reads `settingsModel` the same way `paletteActions()` already does for the palette hints, so it's consistent with the existing pattern rather than a new hazard. The stale window only exists for the milliseconds between first render and the keymap load, and the toolbar re-renders on the first store mutation (session select / cwd report at launch) — both well before the hover + tooltip delay, so a bare label can't actually surface. The only real harden is forcing `settingsModel` non-nil at first render, i.e. wiring it ahead of `actions` in the scene, which the init order deliberately avoids (the `AppActions.settingsModel` doc notes it's wired in the `.task` to dodge the construction-order break). Happy to harden it if you'd rather, but it trades a never-visible edge for that init-order risk.

`swift test` 830/830, `swiftlint --strict` clean, `make build` succeeds.

PR #74, opened 2026-07-02
feat: follow macOS light/dark appearance via ghostty's dual theme value

### LINKED ISSUE #36: Support system light/dark themes
I would like have ability to change theme automatically based on system appearance.

Would you like me to open PR?

paul-nameless:
Follow the macOS light/dark appearance automatically, fixes #36.

Set a dark theme alongside the light one and agterm tracks the system appearance live, switching on each flip (via an app-level KVO observer, so it survives sleep/wake); leave it unset and the single theme always applies. Under the hood it emits ghostty's `light:NAME,dark:NAME` conditional and lets ghostty resolve the active side.

In Settings, the Theme picker sets the theme for the current appearance. A "Follow system appearance" toggle (off by default) reveals a second picker for the other appearance; turning the toggle off collapses back to the single theme.

The control API mirrors the pickers:

- `theme set NAME` sets the light/single slot
- `theme set --dark NAME` sets the dark slot and starts syncing
- `theme set --dark none` clears it, back to a single theme
- `theme set` resets to ghostty's built-in

`theme.list` and every `theme.set` response report the full state (light, dark, sync).


umputun (changes requested):
the design here is heavier than the feature needs, and it comes down to one wrong assumption.

it flattens `light:X,dark:Y` to a single `theme` line in `writeGhosttyConfig` and rewrites + reloads the config on every appearance flip, on the premise that the pinned libghostty can't switch a conditional theme at runtime. It can. I checked the pinned tree (`4dcb09ada`): `theme = light:X,dark:Y` is a first-class runtime conditional, and ghostty's own "Light to Dark" test in `src/config/Config.zig` asserts a color-scheme flip re-resolves to the other side. The reason it looked like it couldn't switch is the flattening itself, `writeGhosttyConfig` hands ghostty a single already-resolved theme so there's never a conditional left to switch. The app is defeating the native path it's working around.

so the simplest version is a lot smaller:

- keep the existing theme picker as-is, no relabel. It stays the theme for the current appearance.
- add a "follow system appearance" toggle, off by default. When it's on, reveal one "alternative" theme picker (the theme for the other appearance). Most users don't want this, so it stays behind the toggle instead of sitting there as a second dropdown.
- map to ghostty's `light:X,dark:Y` under the hood: the existing theme goes to the slot matching the appearance at the time it's set, not always the light one. A user whose default is a dark theme seeds the dark slot and picks a light alternative.
- `ghosttyConfigLines`: emit that dual value raw when an alternative is set, else the single theme as today. One `if`, no parsing.
- on flip: you already call `ghostty_surface_set_color_scheme`, so just trigger the existing reload and let ghostty re-resolve the active side. No active-side math, no rewrite.

that drops `ThemeResolution` and the whole compose/collapse logic in `setLightTheme`/`setDarkTheme`/`commitThemePreview`. That intricacy only exists because two themes are packed into one string and have to be parsed back out to fill two pickers, two plain fields never parse. The chrome refresh rides the reload too, so the new `COLOR_CHANGE` arm isn't needed for this feature (its only separate value is OSC 10/11 from a program, which as written also retints every window's chrome, not just the source surface).

one thing to confirm on your build before dropping the rewrite: that agterm's reload re-feeds the dual `theme` line so ghostty actually swaps. Load `theme = light:Nord,dark:Nord Light`, flip the system appearance, check the terminal + sidebar follow. Source says ghostty does the switch, the only unproven bit is agterm's own reload holding the dual value through `update_config`.

paul-nameless:
Loaded the raw `theme = light:Nord Light,dark:Nord` on this build and flipped the appearance back and forth: ghostty re-resolves both ways (checked with an OSC 11 query from inside the session). My earlier "no re-resolve" test ran against the flattened config, so there was never a conditional to switch.

Reworked as you laid out: two plain fields plus the follow toggle, raw dual emission, existing reload on flip. `ThemeResolution` and the `COLOR_CHANGE` arm are gone.

One gap: the sidebar/titlebar tint reads colors back out of the config, and `ghostty_config_get` on a host-loaded config always answers with the light side. Ghostty.app takes its chrome colors from the config the core returns in the app-target `CONFIG_CHANGE` action, so agterm now does the same: clone it in the callback, read it on reload, free it. About 40 lines, the only callback arm left.

umputun (changes requested):
tests/lint/build all green, and the `CONFIG_CHANGE` clone/take/free lifecycle checks out (single-free, C-callback-contract-clean), so keeping that as-is: the accurate dark-side chrome it buys is worth the ~40 lines.

one thing to fix before merge:

**font zoom is wiped on automatic appearance flips.** `appearanceChanged` routes through `reloadConfigClearingSessionZoom` → `resetSessionFontSizesAllWindows`, so every per-session ⌘+/⌘− zoom is cleared on each OS light/dark switch (and a dark-mode launch with follow-on takes the first-attach reload, so restored zoom is wiped too). Clearing zoom on a manual File ▸ Reload is the documented contract, but an automatic/scheduled flip doing it silently is a surprise. Give the appearance flip its own reload path that re-applies each session's `fontSize` after the theme re-feed (the way `reapplyWatermarkIfNeeded` re-asserts watermarks), instead of routing through `reloadConfigClearingSessionZoom`. A regression test would help: set a session `fontSize`, simulate a follow-on flip, assert it survives.

two smaller ones:

1. drop `migratingLegacyDualTheme`: the `theme = "light:X,dark:Y"` single-string format never shipped, so it's dead code for any real settings.json (only intermediate dev builds of this PR ever wrote it).
2. `nonEmpty` (`ControlServer.swift:712`) triplicates `ThemeCatalog.resolvedName` / `trimmed`; reuse `ThemeCatalog.resolvedName` (already unit-tested) and delete it.

doc gap: README's Appearance paragraph (~line 150) still says just "font and theme", so add the Follow-system toggle + alternate picker + `theme set --dark/--light`. The agent-skill + `.claude/rules` surfaces are already updated.

paul-nameless:
I noticed an issue with my implementation: if appearance changed while mac is asleep, agterm does not detect it and stuck with strange bugs. Reworked with different approach - similar to what ghostty does. Still testing...

umputun:
a couple of changes merged while this was open that overlap it, noting them for the rebase.

#146 merged as a narrow fix for the same `light:X,dark:Y` form: the selected sidebar row rendered white-on-white (unreadable) on a light theme, so it added a host-free `ThemeName.resolved(from:isDark:)` in agtermCore and used it in `GhosttyApp.resolveSelectionColors`. #74 reworks that same theme-resolution area, so you'll want to reconcile on rebase; the dual-theme handling here likely subsumes it, or can reuse `ThemeName.resolved`.

#150 also merged (themed the chrome hairlines in `WindowContentView.swift`), which #74 doesn't touch, so no conflict there; just flagging it's in the tree.

and #149 (a live-flip follow-up filed against #146) is now closed as superseded by this PR, since the `ghostty_app_set_color_scheme` + reload-on-flip wiring here is exactly its fix.

paul-nameless:
merged master, reconciled the #146 overlap: dropped my own `dualThemeSides` and reuse your `ThemeName.resolved`. #150/#149 nothing to do.

heads-up though: since you last looked, i reworked the appearance-following into an app-level KVO observer (`SystemAppearanceObserver`) - the old per-view hook kept wedging after sleep/wake. This implementation works reliably now.

umputun (commented):
all three items from my last pass are resolved: zoom survives an automatic flip now (`reloadConfigPreservingSessionZoom` + `reapplySessionConfigIfNeeded`, and sampling `fontSize` continuously through the flip is the right oracle), `migratingLegacyDualTheme` is gone, `nonEmpty` folded into `ThemeCatalog.resolvedName`, and the README paragraph is updated. Single-theme emission is byte-identical, keep-in-sync is complete.

the KVO rework reads clean - app-lifetime observation auto-invalidates, `[weak self]`, `isDark` resolved in the `@Sendable` closure before the main hop. No lifecycle or thread issue.

few things from this round, all minor:

1. the `debug.appearance` seam (`ControlServer.swift:394`) re-reads `GhosttyApp.currentIsDark()` right after `NSApp.appearance = …` for both the posted `isDark` and the response text. `effectiveAppearance` may not have settled yet, so it can report the wrong side and make the seam flaky. Use the already-validated `side` directly (`side == "dark"`) - the same "don't re-read, take the delivered value" rule the production observer follows. Copilot flagged this one too.

2. the PR body still says "two pickers (Light, Dark with a none row)", but the code is the current-appearance + alternate model and clearing the dark slot is the follow toggle, not a none row. Worth reconciling the body (the `.claude/rules` already match the code). And if you ever add a none row to the alternate picker, `setAlternateTheme(nil)` would leave `followSystemAppearance == true` with `darkTheme == nil` - route it through `setDarkTheme(nil)` instead.

3. nit: `--dark none` is case-sensitive, so `--dark None` falls through to "unknown theme: None". A lowercased compare would be friendlier.

copilot also called the KVO hop at `SystemAppearanceObserver.swift:37` a possible compile error - that's a false positive, it's the established `DispatchQueue.main.async` -> @MainActor convention (GhosttyCallbacks does it throughout) and the build is green.

paul-nameless:
fixed the round:

- `debug.appearance` now takes the validated side instead of re-reading `currentIsDark()` - deterministic, no settle race.
- `--dark none` is case-insensitive now.
- `setAlternateTheme` routes a nil dark slot through `setDarkTheme(nil)`, so following can't be left on with no dark theme (even with no none row in the picker today).

also updated the PR body to the current picker model. and yeah, the `SystemAppearanceObserver:37` copilot flag is a false positive like you said - build's green.

umputun (commented):
last round's items are all done: `debug.appearance` takes the validated `side` (`ControlServer.swift:393`), `setAlternateTheme(nil)` routes through `setDarkTheme(nil)`, `--dark none` is case-insensitive, and the PR body matches the current picker model.

one to fix before merge:

**theme preview can corrupt the light/dark pair if the OS flips mid-preview.** `beginThemePreview` captures only the value (`themePreviewOriginal = effectiveTheme`, `AppActions+Palette.swift:245`), not which slot, and the revert `previewThemeImmediate` writes whichever slot `rendersDarkSlot` names at call time (`SettingsModel.swift:322`; `rendersDarkSlot` reads live `currentIsDark()` at :136). So: open the picker in light mode while following, preview a theme (writes the light slot), let macOS flip to dark, press Esc - `cancelThemePreview` writes the captured light value into the DARK slot and leaves the light slot on the un-reverted preview. Both slots end up wrong, and the next settings save persists it. Snapshot the target slot (or the whole tuple) at `beginThemePreview` and revert against that, rather than re-resolving the slot at Esc-time.

two minor ones:

1. scratch terminal loses its font zoom on an automatic flip (fine as a follow-up). `liveSurfaces()` includes `scratchSurface` (`SettingsModel.swift:702`), so the preserving-zoom reload pushes the default-font config to it, but `reapplySessionConfigIfNeeded` guards on `session?.fontSize != nil` (`GhosttySurfaceView.swift:551`) and the scratch is created with `view.session` unset (`agtermApp.swift:385`) - so a zoomed session's open scratch drops to default font on a flip while the primary keeps zoom. Same class as the primary-pane zoom fix, just doesn't reach the cover surface.
2. `control-api.md:201` says "the public command count stays 50" but the catalog header (:175) and the skill-mirror line (:852) say 52. And the comment at `AppearanceFlipUITests.swift:50` still says the flip fires `viewDidChangeEffectiveAppearance`, which this PR removed (KVO observer + the debug seam now).

paul-nameless:
fixed the preview bug: `beginThemePreview` snapshots the whole `(theme, darkTheme)` pair and Esc reverts both slots, so a mid-preview flip can't strand a value in the wrong slot. two doc nits cleared, merged master (count is 53 now). scratch-zoom left as a follow-up like you flagged.

umputun (commented):
prior round's items check out - the Esc-path pair-snapshot revert, the KVO observer, and zoom-across-flip all verified (the zoom test samples `fontSize` continuously through the flip, so it's a real oracle, not a settled-read false pass).

one to fix before merge - the commit-side twin of the Esc-path bug from last round:

**commit persists an unconfirmed pre-flip preview into the other slot.** `beginThemePreview` snapshots the pair and Esc reverts both, but commit has no symmetric guard. Sequence while following: open the picker in light, preview theme X (writes `settings.theme`, uncommitted) → OS flips to dark (the reload leaves the slots alone) → preview theme Y (writes `settings.darkTheme`) → Enter. `commitTheme()` saves the whole `settings` (`SettingsModel.swift:338`), so it persists `(X, Y)` - X leaks into the light slot though it was only ever a browsed preview, and it resurfaces on the next light flip. Same class as the Esc revert you already fixed: restore the non-active slot from `themePreviewOriginal` before saving, so only the slot active at Enter-time commits. Worth a regression test (preview → flip → preview → commit, assert the un-confirmed slot is unchanged), same shape as the flip test already there.

two minor:

1. `SettingsModel.swift:42` - the `appearanceDebouncer` comment still describes the old per-view mechanism ("posted once per deck surface, plus at first window attach"); it's the app-level KVO observer now (`appearanceChanged`'s own doc at :113 is already correct).
2. `theme set --dark none --light X` fires two reloads (`setDarkTheme(nil)` then `setLightTheme`) - correct, just noting it.

paul-nameless:
fixed the commit-side leak: `commitTheme` takes the captured pair and restores the off-screen slot on Enter, so only the slot active at commit-time persists. added the regression test you suggested (preview -> flip -> preview -> commit), confirmed it fails without the fix. debouncer comment updated, merged master.

umputun (approved):
verified the commit-side fix: `commitTheme` restores the off-screen slot to the open-time snapshot (symmetric with the preview write), and the preview -> flip -> preview -> commit test fails without it. All prior-round items closed.

lgtm, thx

PR #78, opened 2026-07-02
refactor: Move keymap store into core

melonamin:
## Summary
- move keymap loading and starter keymap generation into `agtermCore`
- keep startup behavior the same: missing files are clean, unreadable/invalid UTF-8 files report one diagnostic
- preserve starter keymap default-chord docs, including `(not expressible)` and `(no default)` entries

## Validation
- `make lint`
- `xcodebuild -project agterm.xcodeproj -scheme agterm -configuration Debug -derivedDataPath build/DerivedData build CODE_SIGNING_ALLOWED=NO`
- `xcodebuild -project agterm.xcodeproj -scheme agterm -configuration Debug -destination 'platform=macOS' -derivedDataPath build/DerivedData -only-testing:agtermUITests/KeymapUITests test`\n- manual app launch with isolated keymap configs: starter/valid reload `0`, malformed reload `1`, invalid UTF-8 reload `1`, clean reload `0`

### DIFF
--- a/agterm/SettingsModel.swift
+++ b/agterm/SettingsModel.swift
@@ -341,21 +341,9 @@ final class SettingsModel {
     /// file that EXISTS but can't be read (permissions, invalid UTF-8) is surfaced as a single line-0
     /// diagnostic so the warning banner fires, rather than being silently treated as missing.
     private func loadKeymap() {
-        let url = keymapURL()
-        do {
-            let text = try String(contentsOf: url, encoding: .utf8)
-            let parsed = parseKeymap(text)
-            keymap = parsed.keymap
-            keymapDiagnostics = parsed.diagnostics
-        } catch {
-            keymap = Keymap(builtinOverrides: [:], commands: [])
-            guard FileManager.default.fileExists(atPath: url.path) else {
-                // truly missing — not an error.
-                keymapDiagnostics = []
-                return
-            }
-            keymapDiagnostics = [KeymapDiagnostic(line: 0, message: "could not read keymap.conf: \(error.localizedDescription)")]
-        }
+        let loaded = KeymapStore(configDirectory: configDirectoryURL()).load()
+        keymap = loaded.keymap
+        keymapDiagnostics = loaded.diagnostics
     }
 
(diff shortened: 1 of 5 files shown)


umputun:
heads up - I have been working on a massive mechanical refactoring for the last two days, reducing the size of the biggest file and enforcing file-size linter rules. This one will likely affect (conflict with) all your recent PRs.

melonamin:
Ok, I'll put this on pause until you merge your changes then.

umputun (changes requested):
one thing before this goes in: the starter `keymap.conf` got rewritten/condensed here rather than moved as-is. The hoist itself is a clean extraction, but `starterKeymapConf()` drops the concrete examples (Open in Zed / Lazygit / Deploy), the leader-sequence explanation, and shrinks the `{AGT_*}` safety note to a single line.

can you move the starter text verbatim and keep this PR a mechanical hoist? If you want to trim or rewrite the starter docs, let's do it as its own PR so the content change is reviewable on its own. The `{AGT_*}` note especially is load-bearing - it explains *why* raw tokens are unsafe (a remote host can set the session title via OSC and the pwd via OSC 7), and I don't want that rationale lost in what's meant to be a move.

everything else checks out: the macOS repoint resolves the same path old `keymapURL()` used, the `KeymapStore` tests cover missing/malformed/unreadable, and I ran a release build to confirm no whole-module-optimizer regression.

melonamin:
> the starter keymap.conf got rewritten/condensed here rather than moved as-is

My bad, I messed up when were extracting it from a bigger PR, restored.

umputun (approved):
verified the restore. Diffed the two function bodies and only two lines differ (the rename to `public static func` and `chordSyntax` becoming `starterChordSyntax`), the whole comment block is byte-identical. The Zed/Lazygit/Deploy examples and the full `{AGT_*}` note are back, and the new `ConfigPaths` test asserts them so it can't silently regress again.

one tiny thing, outside this diff and not a blocker: `BuiltinActionTests.swift:102` still says "the same check chordSyntax runs app-side", after this move it's `starterChordSyntax` and runs in core. Worth a one-liner whenever.

lgtm

PR #79, opened 2026-07-02
refactor: Move overlay capture contract into core

melonamin:
## Summary

Hoists the overlay exit-status capture contract into `agtermCore` and rewires the macOS overlay setup to use the shared constants/parser.

## Validation

- focused overlay XCUITests
- live `agtermctl session overlay open "sh -c 'exit 7'" --block` smoke


umputun (approved):
lgtm. Verified the reconstructed `sh -c '( eval "$AGTERM_OVL_CMD" ); echo $? > "$AGTERM_OVL_CODE"'` is byte-identical to the old inline literal, so no behavior change.

one optional nit: the old comment in `agtermApp.swift` noted that the `( eval ... )` subshell is what makes an inline `exit N` propagate to `$?`. That rationale didn't carry into `OverlayCapture.swift`'s doc. A one-liner on `shellLine` would keep it, so nobody later flattens `( eval "$AGTERM_OVL_CMD" )` to a bare `eval` and breaks inline-exit capture. Non-blocking.

PR #81, opened 2026-07-02
refactor: Move custom command matcher into core

melonamin:
## Summary

Hoists the custom-command keybind matcher into `agtermCore` and rewires the macOS runner to use the shared engine while keeping AppKit event handling, leader timers, context resolution, and process spawning app-side.

## Validation

- `make lint`
- Debug `xcodebuild`
- focused `KeymapUITests`
- `codex review --base upstream/master`

### DIFF
--- a/agterm/Commands/CustomCommandRunner.swift
+++ b/agterm/Commands/CustomCommandRunner.swift
@@ -5,7 +5,7 @@ import os
 private let logger = Logger(subsystem: "com.umputun.agterm", category: "CustomCommandRunner")
 
 /// Drives user-defined custom commands: an app-wide `NSEvent` local key monitor turns key presses
-/// into chords, a `KeybindMatcher` resolves them to a command (firing simple chords and leader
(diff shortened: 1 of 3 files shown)


umputun (approved):
lgtm. Traced the two edges that could bite in a matcher hoist and both are clean: the `.fired(id)` -> `commandsByID[id].map(...) ?? .unmatched` fallback can't actually fire (every bind id comes from the same `commands` array as the dict), and the switch from `Dictionary(uniqueKeysWithValues:)` to `uniquingKeysWith: { first, _ in first }` only makes dup-id handling safer.

one nit, non-blocking: `rebuild()` now parses each shortcut with `parseKeybind` just to log the invalid ones, then `CustomCommandEngine.init` parses them all again. Cold path (keymap reload), so the double-parse is negligible - only worth changing if you'd rather have the engine surface the invalid shortcuts and log from that. Fine as-is.

PR #83, opened 2026-07-02
refactor: hoist static command catalog

melonamin:
## Summary
- add a host-free `PaletteCommand` catalog and `PaletteContext` visibility predicates in `agtermCore`
- rewire macOS `paletteActions()` to build static rows from the core catalog with an exhaustive command switch
- keep window, move-session, and custom-command rows in macOS code
- add focused core and palette UI coverage for titles/order and visibility predicates

## Validation
- `swift test --filter PaletteCatalogTests`
- `make lint`
- `./scripts/setup.sh`
- `xcodegen generate`
- `xcodebuild -project agterm.xcodeproj -scheme agterm -configuration Debug -derivedDataPath build/DerivedData build`
- `xcodebuild -project agterm.xcodeproj -scheme agterm -configuration Debug -derivedDataPath build/DerivedData test -only-testing:agtermUITests/PaletteUITests/testActionPaletteClearFlaggedVisibilityTracksFlagState -only-testing:agtermUITests/PaletteUITests/testActionPaletteHidesExpandCollapseInFlaggedView`

### DIFF
--- a/agterm/AppActions+Palette.swift
+++ b/agterm/AppActions+Palette.swift
@@ -19,71 +19,77 @@ extension AppActions {
         return keymap.glyphHint(for: action)
     }
 
+    private var paletteContext: PaletteContext {
+        let activeStore = store
+        return PaletteContext(
+            canRemoveWorkspace: activeStore?.canRemoveWorkspace == true,
+            hasFlaggedSessions: activeStore?.flaggedSessions.isEmpty == false,
+            sidebarShowsWorkspaceTree: activeStore?.sidebarMode == .tree,
+            sidebarShowsFlaggedOnly: activeStore?.sidebarMode == .flagged,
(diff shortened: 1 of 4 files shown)


umputun (approved):
lgtm. Checked the hoist is behavior-preserving: titles, order, and visibility predicates all match the old inline code command-for-command, and PaletteCatalog stays Foundation-only. `swift test` green.

two tiny optional things, take or leave:

1. the no-arg `title` var (PaletteCatalog.swift:62) is only used by the test. Could inline `title(in: PaletteContext())` in the test and drop it.
2. `isVisible` is the one switch with a `default: return true` while the other three are exhaustive. Fine as-is, just means a future conditional command that forgets its predicate shows unconditionally instead of failing to compile.

PR #84, opened 2026-07-02
refactor: hoist resolution error messages

melonamin:
## Summary
- add pure control resolution error-message helpers in agtermCore
- route the macOS ControlServer resolution errors through the core helper
- pin exact not-found and ambiguous wire strings in core tests and focused UI coverage

## Validation
- `cd agtermCore && swift test --filter ControlResolveTests`
- `make lint`
- `make build`
- `xcodebuild test -project agterm.xcodeproj -scheme agterm -destination 'platform=macOS' -only-testing:agtermUITests/ControlAPIUITests/testSessionSelectAmbiguousPrefixErrors -only-testing:agtermUITests/ControlAPIUITests/testCrossWindowUnknownIDErrors`


umputun (commented):
lgtm. One small visibility note: `notFoundMessage` and `ambiguousMessage` come out `public`, but nothing outside agtermCore calls them, only `errorMessage` does (from the app resolver). The core tests use `@testable import`, so they'd test fine as `internal`. I'd drop those two to internal unless you mean them as first-class reusable wire-string helpers. Not a blocker.

PR #85, opened 2026-07-02
feat: optional wrap-around for session next/prev navigation (#80)

### LINKED ISSUE #80: feat: session next/prev navigation should wrap around at the ends (cycle), like attention-nav does
## Summary

Session next/previous navigation stops dead at the ends of the list. Stepping *next* while the last (bottom) session is selected does nothing, and *previous* on the first (top) session does nothing. I'd expect it to **wrap around** — next past the last selects the first, previous past the first selects the last — so navigation cycles like tab/session switching does in most terminals and editors.

## Current behavior

`AppStore.navigateSession(.next/.previous)` clamps at the ends (documented as intentional: *"stop at the ends, no wrap"*). Affects every caller of that seam:

- the ⌥⌘↑ / ⌥⌘↓ menu + palette navigation
- `agtermctl session go next` / `prev` (and the `session.go` control command)

## Why wrap is the better default

- **Consistency inside agterm:** attention-navigation (`next-attention` / `prev-attention`) *already* wraps around the filtered set. Plain next/prev being the only non-wrapping stepper is a surprise.
- **Consistency with the ecosystem:** tab cycling in terminals, browsers, and editors is overwhelmingly cyclic — reaching the end and stopping reads as "navigation broke."
- **No lost capability:** `first` / `last` still exist for people who want to jump to an explicit end, so wrapping doesn't remove the ability to land deterministically at the top/bottom.

## Scope of the change

Small and contained: the `.next`/`.previous` arm of `AppStore.navigateSession` switches from a bounds-guard to modular indexing (the same wrap arithmetic `attentionTarget` already uses). Single-session and empty-tree cases stay no-ops. `first`/`last` are untouched.

Happy to send a PR (I have it ready). If you'd rather this be opt-in behind a setting rather than a behavior change to the default, say so and I'll gate it — but given attention-nav already wraps, matching it as the default seems most consistent.

vnazarenko:
Closes #80.

## What

Adds a **General → Sessions** setting, **"Wrap session navigation"** (default **off**), that makes session next/previous **cycle at the ends** instead of stopping. With it on, stepping *next* while the last (bottom) session is selected selects the first, and *previous* on the first selects the last.

Left off by default, behavior is exactly as today — the clamp-at-the-ends was deliberate, so this doesn't change anyone's existing experience; it's opt-in.

## Why a setting (not a default change)

The stop-at-the-ends behavior is documented as intentional in `navigateSession`, so rather than flip it for everyone this exposes it as a preference. It also mirrors what the codebase already does for **attention-navigation** (`next-attention` / `prev-attention`), which cycles — so users who want that consistency can now get it for plain next/prev too.

## How it works

- `AppStore.navigateSession(_:wrap:)` gains a `wrap: Bool = false` parameter. When `true`, `.next`/`.previous` use modular indexing (the same wrap arithmetic `attentionTarget` already uses); the default `false` keeps the bounds-guard clamp. Wrapping stays **within the visible/filtered set** (flagged mode, focused workspace) and never leaks across the filter. `first`/`last` and attention-nav are unaffected.
- `AppSettings.wrapSessionNavigation: Bool?` (nil = off) — an app-level flag, **not** a ghostty key.
- The two callers read it on demand and pass it, so the surfaces can't drift:
  - **GUI / palette:** `AppActions.selectNextSession` / `selectPreviousSession`
  - **Control API:** `ControlServer`'s `session.go next` / `prev`
- `SettingsModel.setWrapSessionNavigation` is save-only (read on demand at nav time), like the new-session-directory setters — a toggle takes effect immediately, no surface reload.

## Keep-in-sync / Control API

This is a GUI settings toggle in the same family as `compactToolbar` / `rightClickPaste` / `attentionButtonEnabled` — **keep-in-sync exempt** per the settings convention (only `theme.set` / `config.reload` touch settings over the socket). The navigation *action* itself already has full control coverage via `session.go next` / `prev`, and that command **honors this setting at dispatch**, so the CLI and menu behave identically. No new `Command` case is warranted; adding a one-off settings-mutation command would break the established pattern (there is no `settings.*` family). Happy to add socket-level toggling if you'd prefer it, but I matched the existing GUI-settings exemption.

## Tests

`agtermCore` — 903 pass (`swift test`):
- `AppStore`: `navigateNextAtLastStaysPut` / `navigatePreviousAtFirstStaysPut` (default clamp) kept; new `navigateNextAtLastWithWrapCyclesToFirst`, `navigatePreviousAtFirstWithWrapCyclesToLast`, `navigateWithWrapCyclesFullyInBothDirections`, and `navigateWithWrapStaysWithinFocusedWorkspaceSet` (wrap respects the filter).
- `AppSettings`: `wrapSessionNavigationRoundTripsAndIsNotAConfigLine` (Codable round-trip, legacy-absent → nil, not emitted as a ghostty key).

App target builds (Debug). Added `SettingsUITests.testWrapSessionNavigationTogglePersists` mirroring the restore-running-command toggle test (XCUITest — not run in CI, per the repo's CI setup).

Docs updated (keep-in-sync): `settings.md`, `control-api.md`, `menu-actions.md`, `sidebar.md`, and the bundled agent-skill `reference.md`.

### DIFF
--- a/.claude/rules/control-api.md
+++ b/.claude/rules/control-api.md
@@ -247,7 +247,7 @@ paths:
   `session.go` navigates BETWEEN sessions — `args.to` is `next`|`prev`|`first`|`last`|`next-attention`|`prev-attention`
   and acts on the target store's CURRENT selection (it is RELATIVE, so it resolves the placement store
   via `resolvePlacementStore` rather than a session target — there is NO `--target`),
-  stops at the ends on next/prev (no wrap), jumps to the ends for first/last,
+  WRAPS around on next/prev (an end lands on the opposite end, within the filtered set), jumps to the ends for first/last,
   and for `next-attention`/`prev-attention` steps through ONLY the sessions needing attention (`AgentStatus.needsAttention`
   = `blocked`/`completed`) WRAPPING around (skipping idle/active), drives `AppStore.navigateSession`,
   and returns the newly-selected id in `result.id`.
--- a/.claude/rules/menu-actions.md
+++ b/.claude/rules/menu-actions.md
@@ -182,7 +182,7 @@ paths:
   They are real menu items (the Navigate menu — see the menu-split note below),
   so AppKit menu dispatch swallows the shortcut before libghostty and nothing leaks to the shell.
   The pure logic is `AppStore.navigateSession(_:)` (host-free, unit-tested):
-  it flattens the tree (`workspaces.flatMap(\.sessions)`), stops at the ends on next/prev (no wrap),
+  it flattens the tree (`workspaces.flatMap(\.sessions)`), WRAPS around on next/prev (an end lands on the opposite end),
   jumps to the ends for first/last, falls to first on no/invalid selection,
   no-ops on an empty tree, and routes through `selectSession` (recency + badge + persist + workspace-derivation).
   It is shared by the menu, the action palette, and the control channel (`session.go`) so the three can't
--- a/.claude/rules/sidebar.md
+++ b/.claude/rules/sidebar.md
@@ -124,8 +124,8 @@ paths:
   including the stale-focus-id fallback), so clearing the flag/focus naturally restores the full set.
   `navigateSession(_:)` flattens `navigableSessions` for EVERY direction — next/prev/first/last AND attention-nav
   (next-attention/prev-attention scope to the filtered set too) — keeping the same "no/invalid selection
-  → first of the filtered list", "stop at ends, no wrap (attention wraps)" semantics over the filtered
-  list.
+  → first of the filtered list", "next/prev WRAP within the filtered set (like attention-nav)" semantics
+  over the filtered list.
   This is shared by `session.go` (control, no ControlServer change — it already routes through `navigateSession`),
   the ⌥⌘↑/↓ + ⌃⌥↑/↓ menu/palette nav, the Ctrl-Tab MRU switcher (`SessionSwitcher.begin()` scopes its
   candidate set to `store.navigableSessions.map(\.id)`; the MRU ORDER still comes from `sessionRecency`),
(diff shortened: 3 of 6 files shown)


umputun (commented):
fine to change this, but I wouldn't gate it behind a setting - just make wrap the only behavior. Since attention-nav already wraps, plain next/prev wrapping too is the consistency I'd want, not a toggle.

that collapses most of the PR. What's left is the clamp -> modular index in `navigateSession` (drop the `wrap:` param), the doc updates (stops -> wraps), and converting the two clamp tests to wrap tests. The rest can go: `AppSettings.wrapSessionNavigation`, `setWrapSessionNavigation`, the `SettingsView` toggle, `AppActions.wrapSessionNavigation`, the `ControlServer` read, the `settings.md` block, and the two settings tests.

core change is just:

```swift
let step = direction == .next ? 1 : -1
target = ids[((i + step) % ids.count + ids.count) % ids.count] // cycle within the filtered set
```

two housekeeping things: CI is red - swiftlint type_body_length on `AppStoreTests.swift` (1855 > 1850), the 4 added tests tipped it over. The slimmer version adds fewer tests so it'll probably clear on its own; if not, trim rather than raising the limit.

and the branch conflicts with master right now, so it needs a rebase - rebasing on current master to redo this the slimmer way sorts out the conflict at the same time.

vnazarenko:
Done — dropped the setting, wrap is now the only behavior. Agreed on the consistency argument (attention-nav already wraps).

The PR is now just the slim core change you outlined:
- `navigateSession`: clamp → modular index (no `wrap:` param)
- docs: stops → wraps
- the two clamp tests → wrap tests (plus the three scoped/flagged/focus tests that asserted clamp-at-in-set-end now assert wrap-within-the-filtered-set)

Everything else is gone (`AppSettings.wrapSessionNavigation`, the setter, the SettingsView toggle, the `AppActions`/`ControlServer` reads, the settings.md block, and the two settings tests).

On the housekeeping notes: both were from before a rebase. The branch is now rebased onto current master (`a34aec2`), so the conflict is resolved, and the type_body_length red is gone too — the nav tests live in `AppStoreNavigationTests.swift` since the file-size refactor landed, and the slim version adds no net lines. Locally: `swift test` green, swiftlint `--strict` clean, Debug build OK.

umputun (approved):
slimmed exactly as discussed: clamp → `((i+step)%n+n)%n` over `navigableSessions`, setting machinery gone, no `wrap:` param. Tests converted to assert wrap, including the within-filter no-leak cases (focused-workspace + flagged-set wrap to their own ends, never cross). Rebased, CI green. Merging.

PR #90, opened 2026-07-02
feat: pane-addressable session.type and the AGT_PANE keymap token

fkirill:
## What

Two small, paired additions that make split panes reachable from automation:

1. **`session.type` gains `--pane left|right`** (protocol: `args.pane`, reused field — no new `ControlArgs` member). Omitted/`left` is the main pane; `right` injects into the split surface. Mirrors `session.text`'s pane semantics and error strings.
2. **Custom keymap commands get `{AGT_PANE}` / `$AGT_PANE`** (`left`|`right`, always `left` without a split): the pane the shortcut fired from. Keybind path derives it from the focused surface's identity; palette path from `splitFocused`.

Together they close a loop that was previously impossible: a hotkey script can now type back into the exact pane it was invoked from — `agtermctl session type --pane "$AGT_PANE"`.

## Why

`session.type` always injected into `session.surface` (the main pane), so any control-channel automation invoked from a split pane landed its keystrokes in the wrong shell. Concrete case: a keymap-bound popup menu (overlay TUI) that types the picked command into the invoking session — invoked from the right pane, the command lands in the left one. Nothing on the control surface could express "the split pane", and nothing told the fired script which pane it came from; this PR adds both halves.

## Design notes

- **Omitted `--pane` keeps the exact pre-pane behavior (main pane), NOT the focused/on-screen pane.** That's deliberately asymmetric with `session.text` (whose no-pane read is the on-screen surface): changing the injection default would silently redirect existing automation's keystrokes. Happy to flip it to `onScreenSurface` if you prefer consistency over compatibility here.
- `right` requires an existing split surface — the realize/`--select` path stays main-pane-only, since selecting never creates a split. No split → `session has no split pane` (same string as `session.text`).
- Pane validation is enforced server-side in `injectText` (raw socket clients can't bypass the CLI `validate()`), mirroring the `session.text` precedent.
- `SessionTextUITests.testSessionTextPaneSelectsCorrectPane` still feeds the right pane via the real keyboard on purpose (it also proves focus→first-responder routing); the new `SessionTypePaneUITests` covers the `--pane right` injection route.

## Keep-in-sync audit (four surfaces)

1. `ControlProtocol.swift` — reuses `ControlArgs.pane`, doc updated.
2. `ControlServer` — pane switch in `injectText` (`ControlServer+SurfaceIO.swift`).
3. `agtermctl` — `session type --pane left|right`, `validate()`-guarded.
4. Tests — round-trip (`ControlProtocolTests`), CLI mapping + rejection (`CommandsTests`), token/env symmetry incl. the new `AGT_PANE` (`CustomCommandTests`), e2e (`SessionTypePaneUITests`: `--pane right` reaches the split pane and only it, no-split errors, invalid pane rejected server-side).

Plus the doc mirrors: `agterm/Resources/agent-skill/` (SKILL.md, reference.md, examples.md — command count unchanged, no new command), README (token table + `session type` note), `.claude/rules/control-api.md` + `.claude/rules/keymap.md`.

## Verification

- `swift test`: 906 tests green.
- `make lint` (`--strict`): clean.
- `make build` (Debug): green.
- XCUITest: `SessionTypePaneUITests` (3/3) and `SessionTextUITests/testSessionTextPaneSelectsCorrectPane` pass locally.

### DIFF
--- a/.claude/rules/control-api.md
+++ b/.claude/rules/control-api.md
@@ -295,10 +295,25 @@ paths:
   (the last two also `validate()`-guarded); and round-trip + e2e (`testSessionNewWithCommandRunsAsProcess`,
   `testSessionNewWithName`, `testSessionNewWorkspaceNameCreatesThenReuses`) cover them.
   `session.type` injects into the target surface.
+  `args.pane` picks the pane like `session.text` (`left`|`right` only, no `other`):
+  omitted/`left` is the MAIN pane (omitted deliberately keeps the pre-pane behavior — always the main
+  pane, NOT the focused/on-screen one, so existing automation is unaffected);
+  `right` injects into the split surface, `session has no split pane` without one,
+  and an unknown value is an `invalid pane` error — all validated SERVER-SIDE in `injectText`
+  (mirroring the CLI `validate()`), so a raw socket client can't bypass it.
   Every session is realized eagerly (the deck mounts all at startup), so any session is normally typable
   WITHOUT `select`; `select:true` remains for the brief window before a just-created session is mounted
   (select, then a bounded poll, the `focusSplitPane` idiom), with `session not realized` the fallback
   if the surface still isn't up.
+  The realize/select path applies to the MAIN pane only — a split pane is never created by selecting,
+  so `pane:right` injects into the existing split surface or errors.
+  Four-point keep-in-sync audit for `session.type --pane`: (1) reuses `ControlArgs.pane` in
+  `ControlProtocol.swift` (no new field), (2) the pane switch in `injectText`
+  (`ControlServer+SurfaceIO.swift`), (3) the `session type --pane left|right` option
+  (`validate()`-guarded) in `agtermctlKit`, (4) round-trip in `ControlProtocolTests` + CLI mapping in
+  `CommandsTests` + the e2e (`testSessionTypePaneRightReachesSplitPane`,
+  `testSessionTypePaneRightWithoutSplitErrors`, `testSessionTypeRejectsInvalidPaneServerSide`) in
+  `SessionTypePaneUITests` (a `ControlAPITestCase` subclass like `SessionTextUITests`).
   `GhosttySurfaceView.inject(text:)` types via `ghostty_surface_key` keystrokes (printable runs as key-with-`text`,
   each `\n`/`\r`/`\r\n` as a Return keypress, keycode 36) — NOT `ghostty_surface_text`,
   whose bracketed-paste wrapping suppresses Enter and leaks `\e[200~`/`\e[201~` markers when fired rapidly.
--- a/.claude/rules/keymap.md
+++ b/.claude/rules/keymap.md
@@ -54,11 +54,16 @@ paths:
   in a text field) and rebuilds its matcher on `.agtermKeymapChanged`.
   The runner also exposes a public `run(_:)` for the palette items, which resolve context from the active
   session (no first responder to key off).
-  `CommandContext.pane` (the `{AGT_PANE}`/`$AGT_PANE` token, `left`|`right`, `left` without a split)
+  `CommandContext.pane` (the `{AGT_PANE}`/`$AGT_PANE` token, `left`|`right`)
   carries the fired-from pane: the keybind path derives it from the focused SURFACE's identity
   (`splitSurface === focusedSurface`), the palette path from `session.splitFocused` —
   so a script can feed it back as `agtermctl session type --pane "$AGT_PANE"` to type into the very
   pane the shortcut was pressed in.
+  It reflects the pane's PHYSICAL surface slot, not `hasSplit`: a promoted split survivor (the primary
+  pane exited and the split pane took over) reports `right` even though the session no longer shows a
+  split, because that survivor still lives in the `splitSurface` slot — which is exactly where
+  `session.type --pane right` reaches it, so the round-trip stays correct (a `left` there would name a
+  pane `session.type` can't type into, since the promoted session's `surface` slot is nil).
 - **Built-in override resolution is ORDER-INDEPENDENT, decided against the FINAL state (`resolveBuiltinOverrides`).**
   Overrides are NOT folded incrementally against a partially-built map (that was order-sensitive — it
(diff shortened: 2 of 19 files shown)


umputun (commented):
two things worth changing before merge:

1. **`{AGT_PANE}` reports `right` for a promoted single-pane session.** after a primary pane exits, `closePrimaryPane` leaves the survivor in `splitSurface` with `hasSplit=false`, `splitFocused=true`. So both derivations in `CustomCommandRunner` (palette via `splitFocused`, keybind via `splitSurface === focusedSurface`) report `right` for a session that has no split, which contradicts the `CommandContext.pane` "no split -> left" contract. Can you gate both on `session.hasSplit`, falling back to `left`? One line in each path.

2. **the `{AGT_PANE}` derivation has no e2e coverage.** the host-free tests prove a given pane value flows into env/expand, but nothing exercises the runner computing `left` vs `right`, which is the actual new logic on the keybind side. `KeymapUITests` has the pattern already (a `command "..." <chord>` firing a marker via `typeKey`). A `$AGT_PANE`-to-marker command fired from the main pane (expect `left`) and from a focused split (expect `right`) would cover it.

non-blocking:

- the `--pane right` inject returns `ok` in the brief window right after `session.split on` where the split surface object is mounted but its libghostty handle isn't realized yet (`inject` no-ops silently). Fine to leave as-is, it mirrors `session.text --pane right`'s no-poll behavior and the `{AGT_PANE}` feedback path fires from inside an already-realized split. Just flagging it as a known edge.

- Copilot flagged the `ghostty_surface_text` line in the `injectText` doc block (`ControlServer+SurfaceIO.swift:284`) as stale, and it's right, injection goes through `ghostty_surface_key` via `inject`. That line predates your PR (it's mine), so not yours to fix. I'll clean it up separately, or fix the one line while you're in that block if you like.

fkirill:
> two things worth changing before merge:
> 1. `{AGT_PANE}` reports `right` for a promoted single-pane session... Can you gate both on `session.hasSplit`, falling back to `left`?

traced this and the `hasSplit` gate has a hidden cost. after `closePrimaryPane` the survivor physically stays in the `splitSurface` slot (`surface` is torn down to nil). so on the injection side, `session type --pane right` reaches it, but `--pane left` hits the nil `surface` and returns `session not realized` — nothing ever remounts `surface` (the maximize branch only renders `splitSurface`).

so gating `{AGT_PANE}` to `left` there would emit a pane value whose own round-trip fails: `session type --pane "$AGT_PANE"` can't type into the promoted pane. that's the exact contract this feature is for, so `left` is worse than the imprecise doc — `right` is the *reachable* value.

the real mismatch is the "no split → left" wording vs. an implementation where a split-less session's only pane IS the split surface. two ways out:

- **now:** keep the behavior, reword the `CommandContext.pane` contract to "reflects the pane's physical slot — a promoted split survivor reports `right`, since that's where `session.type --pane` reaches it." round-trip stays correct.
- **follow-up (bigger):** migrate the survivor `splitSurface → surface` in `closePrimaryPane` (drop `splitFocused`), so `left` becomes both honest and reachable — and a re-split then opens the fresh pane on the right instead of shoving your work over. touches surface identity + the maximize render branch, wants its own test.

leaning toward the reword for this PR and filing the migration separately — ok with that, or do you want the migration in-scope here?

> 2. the `{AGT_PANE}` derivation has no e2e coverage.

i've got `CustomCommandPaneTokenUITests` locally that fires a `$AGT_PANE`-to-marker command from the main pane (expects `left`) and a focused split (expects `right`) on the keybind path, plus a palette case — just not pushed yet. pushing it with the rest. only gap left is the palette `left` case, i'll add it.

> Copilot flagged the `ghostty_surface_text` line...

already fixed that one line locally while i was in the block — it'll be in the push.

umputun (changes requested):
reword's the right call. Traced it too: after the primary exits, `left` would report a pane `--pane left` can't reach, so `right` is the honest and reachable value. Let's take the reworded contract now and keep the `splitSurface -> surface` migration as its own PR.

two asks:

1. Please add the promoted-survivor e2e. It's the one edge still uncovered and it's exactly the case we went back and forth on. Reproducible: split, `exit` the primary shell, fire the `$AGT_PANE` probe, expect `right`, and confirm `--pane $AGT_PANE` types into that pane.

2. The palette pane derivation (`CustomCommandRunner.swift:184`) reads `right` from bare `splitFocused`, so in the window right after `session split on` before `splitSurface` mounts it reports `right` while `--pane right` still errors "no split pane". Guard it on `splitSurface != nil`, the same way `Session.onScreenSurface` does (`splitFocused && splitSurface != nil`). One line, and it also fixes the selection read two lines up that has the same bare-`splitFocused` gap.

stale `ghostty_surface_text` line and the new pane e2e both look good, thx.

fkirill:
both done in `ca1ead9`.

1. **promoted-survivor e2e** — `CustomCommandPaneTokenUITests.testAgtPanePromotedSurvivorReportsRight`: split on, `exit` the primary shell so `closePrimaryPane` promotes the survivor, assert `$AGT_PANE=right`, then round-trip `session type --pane <reported>` and confirm the marker lands in that pane's own buffer.

2. **palette guard** — `run(_:)` now derives both the pane and the selection surface from `splitFocused && splitSurface != nil` (the `Session.onScreenSurface` idiom) instead of bare `splitFocused`. So the window right after `session split on`, before `splitSurface` mounts, reports `left` (reachable) rather than `right`-that-errors; a promoted survivor keeps `splitSurface` non-nil, so it still reports `right`. the keybind path was already nil-safe via the surface-identity check (`nil === focusedSurface` → left), so only the palette path needed it.

`swift test` + `swiftlint --strict` + `build-for-testing` are green. i haven't run the XCUITest suite locally (it grabs the screen), so the new e2e is build-verified not yet executed — happy to run it if you'd like.

`splitSurface → surface` migration left for its own PR, as agreed.

umputun (approved):
both addressed in `ca1ead9`/`2b461a7`: palette `{AGT_PANE}` gated on `splitFocused && splitSurface != nil` (pane + selection read), promoted-survivor e2e added. Ran the pane e2e locally: 7/7 green, including the survivor round-trip. Squashing.

PR #92, opened 2026-07-02
refactor: hoist theme catalog facts

melonamin:
Summary:
- Add a small host-free ThemeCatalog for theme ordering, default-row metadata, picker IDs, and control-name resolution.
- Rewire the macOS Settings/theme palette/control validation call sites to consume the shared catalog.
- Add focused core tests for ordering, ids, default behavior, and exact name matching.

Validation:
- cd agtermCore && swift test
- make lint
- make build

Note:
- Confirmed the earlier WindowAccessor.swift strong-capture build warnings are not from this PR: that file is unchanged in the branch.

### DIFF
--- a/agterm/AppActions+Palette.swift
+++ b/agterm/AppActions+Palette.swift
@@ -199,8 +199,9 @@ extension AppActions {
     /// the current one badged. Navigating a row previews it live (`onSelect`); Enter/click commits it.
     func paletteThemes() -> [PaletteItem] {
         let current = settingsModel?.settings.theme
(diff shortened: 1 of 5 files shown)


umputun (approved):
lgtm, one nit. `ThemeCatalog.names(in:)` has no production caller: `SettingsCatalog.themeNames()` still does its own `contentsOfDirectory` and only uses the catalog for sorting (`ThemeCatalog(names:).names`), so `names(in:)` is exercised by its test but never called in the app. Either route `themeNames()` through it (which drops the manual listing + empty-guard) or drop it. Not a blocker.

PR #96, opened 2026-07-02
fix: escape newlines in ShellEscape to prevent drag-drop command injection

### LINKED ISSUE #95: Command injection: a newline in a dropped filename executes shell commands
### Summary

Dropping a file whose name contains a newline character causes agterm to execute shell commands embedded in the filename. Drag-and-drop is meant to insert the file's path at the cursor (no trailing Enter), but embedded newlines are converted into real Return keypresses and submitted to the shell.

### Affected code

- `agtermCore/Sources/agtermCore/ShellEscape.swift` — the `significant` character set escapes spaces/quotes/`$`/`;`/`|`/tab/etc. but **not** `\n` or `\r`.
- `agterm/Ghostty/GhosttySurfaceView.swift` — `inject(text:)` deliberately converts each `\n`/`\r`/`\r\n` into a Return keypress (`sendReturn`).
- Drop path: `performDragOperation` → `dropText` → `pasteboardText` (calls `ShellEscape.path(url.path)`) → `inject(text:)`.

### Root cause

A dropped file URL's path is shell-escaped by `ShellEscape.path`, but that escaper does not handle line terminators. Any `\n`/`\r` in the path survives into `inject(text:)`, which submits it as a command boundary. Filenames may contain any byte except `/` and NUL, so newlines in names are legal.

### Safe reproduction (harmless — no arguments, nothing is deleted)

Note: `ShellEscape` escapes spaces, so a command *with arguments* actually breaks at the escaped space and won't run — only the newline survives unescaped. A **no-argument** command demonstrates the bug without any way to weaponize it here.

```sh
OUT="$(mktemp -d)/drop-poc"; mkdir -p "$OUT"
touch "$OUT/$(printf 'report.txt\ndate\n')"   # runs `date` after the name
touch "$OUT/$(printf 'harmless.txt\n')"        # trailing newline only -> submits the name
ls -1b "$OUT"
```

Drag `report.txt␊date` into an agterm terminal:

- **Expected:** the path is inserted at the cursor; nothing executes.
- **Actual:** `report.txt` is submitted (command not found), then `date` runs on its own line — a command that came from the filename, not the keyboard.

The `harmless.txt` file (single trailing newline) submits its own name as a bogus command, which alone proves the drop presses Enter when it should not.

### Impact

A maliciously-named file delivered via zip/tarball/AirDrop runs arbitrary no-argument commands on drop; a single trailing newline already causes an unintended command submission.

### Suggested fix

Add `\n` and `\r` to `ShellEscape.significant` (backslash-escaped they become a shell line-continuation, not a command terminator), or strip/reject control characters in the drop path. `ShellEscapeTests` currently has no newline case.

A fix + regression test is prepared in the linked PR.

vlondon:
Fixes #95.

### Problem

Dropping a file whose name contains a newline executed shell commands embedded in the filename. `ShellEscape.path` did not escape `\n`/`\r`, and the drop path feeds the escaped path through `GhosttySurfaceView.inject(text:)`, which converts a bare newline into a Return keypress — so any newline in the filename submitted the following text to the shell.

### Fix

Add `\n` and `\r` to `ShellEscape.significant` so a line terminator is backslash-escaped into a shell line-continuation instead of a command terminator. This is contained to the host-free `ShellEscape` and does not touch the shared `inject(text:)` newline→Return behavior that `session.type` relies on.

### Test

Added `escapesNewlinesSoADroppedNameCannotInjectACommand` to `ShellEscapeTests`, covering `\n`, `\r`, and `\r\n`.

### Verification

- `swift test` — all 899 tests pass.
- Longest changed line is 109 cols (limit 200); no file-size limits affected.

### Safe manual repro

`ShellEscape` escapes spaces, so only a no-argument command survives — this demonstrates the bug harmlessly:

```sh
OUT="$(mktemp -d)/drop-poc"; mkdir -p "$OUT"
touch "$OUT/$(printf 'report.txt\ndate\n')"
ls -1b "$OUT"
```

Before this change, dragging `report.txt␊date` in ran `date`; after, the path is inserted without executing.

### DIFF
--- a/agtermCore/Sources/agtermCore/ShellEscape.swift
+++ b/agtermCore/Sources/agtermCore/ShellEscape.swift
@@ -10,8 +10,9 @@ import Foundation
 /// wrap even a clean `/path/image.png` in quotes.
 public enum ShellEscape {
     /// The shell-significant characters that get a backslash prefix. Backslash is FIRST so the pass that
-    /// escapes it does not double-escape the backslashes added for the rest.
-    private static let significant: [Character] = Array("\\ ()[]{}<>\"'`!#$&;|*?\t")
+    /// escapes it does not double-escape the backslashes added for the rest. `\n`/`\r` are escaped so a
+    /// dropped filename with a newline can't inject a command via `inject(text:)` (bare newline → Return).
+    private static let significant: [Character] = Array("\\ ()[]{}<>\"'`!#$&;|*?\t\n\r")
 
     /// Backslash-escape every shell-significant character in `path`; a path with none is returned unchanged.
     public static func path(_ path: String) -> String {
--- a/agtermCore/Tests/agtermCoreTests/ShellEscapeTests.swift
+++ b/agtermCore/Tests/agtermCoreTests/ShellEscapeTests.swift
@@ -19,6 +19,14 @@ struct ShellEscapeTests {
         #expect(ShellEscape.path("a&b;c|d$e") == "a\\&b\\;c\\|d\\$e")
     }
 
+    @Test func escapesNewlinesSoADroppedNameCannotInjectACommand() {
+        // a filename may contain a newline; unescaped it would submit the rest as a shell command via
+        // inject(text:). Backslash-escaped, the line terminator can't terminate a command.
+        #expect(ShellEscape.path("report.txt\ndate") == "report.txt\\\ndate")
+        #expect(ShellEscape.path("a\rb") == "a\\\rb")
+        #expect(ShellEscape.path("a\r\nb") == "a\\\r\\\nb")
+    }
+
     @Test func emptyStaysEmpty() {
         #expect(ShellEscape.path("") == "")
     }


umputun (approved):
lgtm. Traced it end to end - the drop path is the only one that feeds `inject(text:)`, which converts `\n`/`\r` into a real Return keypress (it deliberately skips bracketed paste), so escaping the terminator in `ShellEscape` is the right layer. Backslash-first ordering still holds, and CR/CRLF are covered since `inject` normalizes them to `\n`.

checked the paste side too, since a bad filename can arrive that way. Cmd-V / middle-click routes through libghostty's native paste (`ghostty_surface_complete_clipboard_request`), which honors bracketed paste, so a pasted newline isn't auto-submitted - not the same vector. A file pasted from Finder goes through the same `pasteboardText` / `ShellEscape.path`, so this hardens that path too.

PR #98, opened 2026-07-02
fix: keep '#' inside single-quoted custom-command shell args

### LINKED ISSUE #97: Custom command silently truncated at '#' inside single quotes in keymap.conf
### Summary

A custom command in `keymap.conf` whose shell line contains a `#` preceded by whitespace and inside **single** quotes is silently truncated at parse time. The command is stored corrupted (unterminated quote), with **no diagnostic**, and only fails later at fire time with a generic failure banner.

### Affected code

- `agtermCore/Sources/agtermCore/Keymap.swift` — `stripComment` treats a `#` preceded by whitespace and outside a **double**-quoted span as an inline comment; it does not track single quotes.
- The doc comments (around `Keymap.swift:52-55` and `:278-281`) describe only double-quote protection.

### Reproduction

`keymap.conf`:
```
command "Commit" cmd+shift+c git commit -m 'fix #42'
```

Parsing this stores the command as `git commit -m 'fix` — everything from ` #42'` onward is dropped, leaving an unterminated single quote. `parseKeymap` emits **0 diagnostics**, so there is no feedback that the command was mangled. Pressing ⌘⇧C then runs a broken `/bin/sh -c` and shows only a generic failure banner.

Verified by feeding a `keymap.conf` through `parseKeymap` directly:

| line | stored `command` | result |
|---|---|---|
| `... git commit -m 'fix #42'` | `git commit -m 'fix` | truncated |
| `... git commit -m "fix #42"` | `git commit -m "fix #42"` | fine (double quotes already protected) |
| `... curl http://example.com/#frag` | `curl http://example.com/#frag` | fine |

### Precise trigger

The `#` is stripped only when it is **preceded by whitespace** and outside a double-quoted span. So the breaking cases are single-quoted content with a space before `#` (e.g. commit messages: `'fix #42'`, `'msg #123'`). A URL fragment like `http://h/#frag` (the `#` follows `/`) and `echo '#!/bin/sh'` (the `#` follows the opening quote) are **not** affected.

### Impact

Common, realistic shell lines — most notably `git commit -m 'msg #123'` — are silently corrupted, and the failure surfaces only at runtime with no pointer to the cause.

### Suggested fix

Track single quotes in `stripComment` as well (a `#` is a comment only when preceded by whitespace and outside **both** single- and double-quoted spans), and update the doc comments. Add `KeymapTests` cases for single-quoted and mixed-quote `#`.

A fix + regression tests are prepared in the linked PR.

vlondon:
Fixes #97.

### Problem

A custom command in `keymap.conf` whose shell line contained a `#` preceded by whitespace inside **single** quotes was silently truncated at parse time. `stripComment` only tracked double quotes, so:

```
command "Commit" cmd+shift+c git commit -m 'fix #42'
```

was stored as `git commit -m 'fix` (unterminated quote), **with 0 diagnostics** — the corruption was silent and surfaced only at fire time as a generic failure banner.

### Fix

Track single quotes in `stripComment` as well: a `#` is an inline comment only when preceded by whitespace and outside **both** single- and double-quoted spans. The two quote states are mutually exclusive (a `"` inside `'...'` is literal, and vice versa), which the implementation respects. Updated the grammar/`stripComment` doc comments to match.

### Tests

Added `KeymapTests` cases:
- `hashInsideSingleQuotedShellArgIsKept` — `'fix #42'` survives.
- `trailingCommentAfterSingleQuotedArgIsStripped` — a real `# comment` after a closed single quote is still stripped.
- `doubleQuoteInsideSingleQuotesDoesNotOpenAQuotedSpan` — a `"` inside `'...'` is literal, so a following `#` still starts a comment.

The existing `inlineCommentInsideQuotedNameIsKept` / `parseSkipsCommentsAndBlankLines` cases continue to pass.

### Verification

`swift test` — all tests pass (934). Longest changed line is 106 cols (limit 200).

### DIFF
--- a/agtermCore/Sources/agtermCore/Keymap.swift
+++ b/agtermCore/Sources/agtermCore/Keymap.swift
@@ -80,9 +80,10 @@ public struct KeymapStore: Sendable {
 ///
 /// Grammar (kitty-flavored): the file is line-based. Blank lines and lines whose first non-space
 /// character is `#` are ignored. A trailing inline comment is stripped when a `#` is preceded by
-/// whitespace AND lies outside any double-quoted span (so a `#` inside a `command "name"` or inside
-/// the shell line is kept) — this keeps `command "x" echo a#b` and `map cmd+a#` handling simple while
-/// allowing `map cmd+d toggle_split  # rebind`.
+/// whitespace AND lies outside any quoted span, single OR double (so a `#` inside a `command "name"`,
+/// inside a double-quoted shell arg, or inside a single-quoted one like `git commit -m 'fix #42'` is
+/// kept) — this keeps `command "x" echo a#b` and `map cmd+a#` handling simple while allowing
+/// `map cmd+d toggle_split  # rebind`.
 ///
 /// The first whitespace-token is the verb:
 /// - `map <chord> <action>`: `<chord>` is parsed via `parseKeybind` (a leader sequence, count > 1, is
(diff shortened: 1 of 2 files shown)


umputun (approved):
lgtm, two non-blocking notes:

1. test symmetry - you cover `"` inside `'...'` staying literal (`doubleQuoteInsideSingleQuotesDoesNotOpenAQuotedSpan`) but not the mirror: a `'` inside a double-quoted arg staying literal so a trailing `#` still cuts, e.g. `echo "a'b" # c` -> `echo "a'b"`. One more case would lock both directions.

2. fyi, not a change request: a lone unquoted `'` now opens a single-quote span that never closes, so a trailing `# comment` after it is no longer stripped (`echo don't # x` keeps `# x`). Matches POSIX shell (unterminated quote) and such a line is already broken shell, so probably fine - just flagging the behavior change vs the old double-quote-only logic.

PR #100, opened 2026-07-02
fix: single-quote-escape paths in show-image.sh overlay command

### LINKED ISSUE #99: show-image.sh: image path containing a single quote breaks the overlay command
### Summary

The bundled `show-image.sh` helper embeds the image path (and its own path) single-quoted with no escaping when it builds the overlay command. An image path containing a single quote breaks the command — the overlay fails to open (and it's a quoting-injection vector).

### Affected code

- `agterm/Resources/agent-skill/scripts/show-image.sh` (last line):
  ```sh
  agtermctl session overlay open "/bin/bash '$self' --emit '$abs'" --size-percent "$size"
  ```
  `$abs` (and `$self`) are wrapped in single quotes with no `'\''` escaping. `agtermctl session overlay open` runs the command via `sh -c` (`makeOverlaySurface` wraps it in `sh -c '…'`), so an embedded single quote terminates the quoting early.

### Reproduction

Simulating exactly how the overlay runs the command, with a path containing a single quote:

```sh
self="/path/to/show-image.sh"
abs="/tmp/it's a pic.png"
cmd="/bin/bash '$self' --emit '$abs'"
sh -c "$cmd"
# -> sh: -c: line 0: unexpected EOF while looking for matching `''
#    sh: -c: line 1: syntax error: unexpected end of file
```

The `'` in `it's` closes the single-quoted span early; the trailing `'` is then unmatched, so `sh -c` aborts and the image never displays.

### Fix

Single-quote-escape both `$self` and `$abs` with the POSIX `'\''` idiom before embedding, e.g.:

```sh
sq() { printf "'%s'" "$(printf '%s' "$1" | sed "s/'/'\\\\''/g")"; }
agtermctl session overlay open "/bin/bash $(sq "$self") --emit $(sq "$abs")" --size-percent "$size"
```

Verified: with the escape, a path like `/tmp/it's a "weird" pic.png` is passed through `sh -c` as a single intact argument.

A fix is prepared in the linked PR.

vlondon:
Fixes #99.

### Problem

`show-image.sh` built the overlay command by embedding `$self` and `$abs` single-quoted with no escaping:

```sh
agtermctl session overlay open "/bin/bash '$self' --emit '$abs'" --size-percent "$size"
```

`agtermctl session overlay open` runs the command via `sh -c`, so an image path containing a single quote (e.g. `/tmp/it's a pic.png`) closed the quote early and aborted with `unexpected EOF while looking for matching '`. The image never displayed.

### Fix

Add a POSIX single-quote helper and quote both paths through it:

```sh
sq() { printf "'%s'" "$(printf '%s' "$1" | sed "s/'/'\\\\''/g")"; }
agtermctl session overlay open "/bin/bash $(sq "$self") --emit $(sq "$abs")" --size-percent "$size"
```

### Verification

- `bash -n` — no syntax errors.
- Simulated the `sh -c` parse the overlay performs: a path like `/tmp/it's a "weird" pic.png` is now passed through as a single intact argument (`arg[/tmp/it's a "weird" pic.png]`), where before it aborted the command.

Only the single source of truth at `agterm/Resources/agent-skill/scripts/show-image.sh` is changed; installed copies are regenerated by Help ▸ Install Agent Skill.

### DIFF
--- a/agterm/Resources/agent-skill/scripts/show-image.sh
+++ b/agterm/Resources/agent-skill/scripts/show-image.sh
@@ -45,5 +45,10 @@ command -v agtermctl >/dev/null 2>&1 || { echo "show-image.sh: agtermctl not on 
 self=$(cd "$(dirname "$0")" && pwd)/$(basename "$0")
 abs=$(cd "$(dirname "$img")" && pwd)/$(basename "$img")
 
+# POSIX single-quote a string for safe embedding in the command below: wrap in '...' and replace each
+# embedded ' with the '\'' sequence, so a path containing a quote can't break out of the quoting. The
+# overlay command is run via `sh -c`, so an unescaped ' in the path would otherwise abort it.
+sq() { printf "'%s'" "$(printf '%s' "$1" | sed "s/'/'\\\\''/g")"; }
+
 # open an overlay that re-invokes this script in --emit mode inside the overlay's pty.
-agtermctl session overlay open "/bin/bash '$self' --emit '$abs'" --size-percent "$size"
+agtermctl session overlay open "/bin/bash $(sq "$self") --emit $(sq "$abs")" --size-percent "$size"


umputun (approved):
lgtm, thx. Escaping is the correct POSIX single-quote idiom and the `sh -c` round-trip checks out.

PR #105, opened 2026-07-03
refactor(control): add dispatcher foundation

melonamin:
Summary:
- Add a small agtermCore ControlDispatcher/ControlActions seam for routed control commands.
- Migrate only tree and sidebar visibility/mode/expand/collapse through the dispatcher.
- Keep all other ControlServer command arms inline as fallback and cover routed behavior with core mock-action tests.

Validation:
- cd agtermCore && swift test
- make lint
- make build
- xcodebuild focused UI tests for tree, sidebar show/hide, sidebar.mode, and sidebar expand/collapse


umputun (approved):

umputun:
this conflicts with #107 (setSidebar) and #108 (buildTree), both merged now, so it needs a rebase on master before it can go in. Approved on my side, so just the rebase.

one behavior FYI while you're in there: for `sidebar`/`sidebar.mode` the mode string now gets parsed before the `activeStore` check, so "no open window" + a bad mode returns the invalid-mode error instead of "no open window". Degenerate case, and the tests pin the new order, so fine to keep. Just flagging since the goal was to preserve behavior, your call.

PR #120, opened 2026-07-04
[codex] Route remaining session controls through dispatcher

melonamin:
## Summary

Routes the remaining group 3b session control families through `ControlDispatcher`:

- `session.type`
- `session.copy` (the existing selection-copy command)
- `session.overlay.open`, `session.overlay.close`, `session.overlay.result`

The dispatcher now owns host-free validation and exact control-response routing. The macOS adapter keeps target resolution, surface injection, overlay lifecycle, and other AppKit/process side effects on the app side.

Also updates `nonMigratedCommandFallsThrough` to probe `session.select`, per the prior review nit, so the fallthrough test stays outside this migration set.

## Validation

- `swift test --filter ControlDispatcherTests` — passed, 35 tests
- `swift test` in `agtermCore` — passed, 853 tests
- `make lint` — passed
- `make build` — passed
- Focused UI/control run:
  - `ControlAPIUITests` — passed
  - `ControlOverlaySplitUITests` — passed
  - `SessionTypePaneUITests/testSessionTypePaneRightReachesSplitPane` — failed with the same assertion on a clean `origin/master` worktree, so this appears pre-existing rather than introduced by this refactor

### DIFF
--- a/agterm/Control/ControlServer+SessionActions.swift
+++ b/agterm/Control/ControlServer+SessionActions.swift
@@ -28,6 +28,66 @@ extension ControlServer: ControlActions {
         collapseWorkspaces(window: window)
     }
 
+    func typeSession(_ target: String?, window: String?, options: ControlSessionTypeOptions) async -> ControlResponse {
+        // Resolve first (cross-window when no `args.window`), then realize-and-inject; the realize
+        // path is async (bounded poll), so this can't go through the synchronous `resolveSession`
+        // helper. The not-found / ambiguous error strings must stay in sync with `resolve(...)`.
+        switch resolver.resolveSessionTarget(target, window: window) {
+        case .failure(let response):
+            return response
+        case .success(let (store, id)):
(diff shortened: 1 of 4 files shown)


umputun (approved):
lgtm. Faithful 1:1 hoist - validation ordering, error strings, and response shapes all match the old inline arms. Core tests + swiftlint clean here too.

two trivial nits, neither blocking:

1. the comments you moved into `ControlServer+SessionActions.swift` got sentence-cased on the way over (`resolve first` -> `Resolve first`, `select the target` -> `Select the target`) and the `—` swapped for `--`. They were lowercase in the old ControlServer arms, worth keeping them that way to match the in-function comment convention.
2. `ControlSessionTypeOptions` / `ControlSessionOverlayOpenOptions` are inline in `ControlDispatcher.swift`, while the sibling option types live in `ControlModes.swift`. Moving them there would match the series, but keeping them next to the dispatch arm is fine too - your call.

PR #124, opened 2026-07-04
refactor(control): backfill dispatcher session workspace routes

melonamin:
## Summary
- route remaining simple session/workspace control commands through `ControlDispatcher`
- keep app-side resolver/store behavior in the `ControlActions` adapter
- add focused dispatcher tests for migrated success paths and exact guard errors

## Scope
Migrated `session.select`, `session.go`, `session.close`, `session.rename`, `workspace.new`, `workspace.select`, `workspace.rename`, and `workspace.delete`.

Left surface/UI/global commands inline: `session.text`, `session.background`, `session.search`, `quick`, `restore.clear`, and `window.*`.

## Validation
- `make lint`
- `swift test`
- `make build`

## Notes
This is intended as a pure refactor: validation order, response strings, response shapes, resolver calls, and store mutations are preserved.

### DIFF
--- a/agterm/Control/ControlServer+SessionActions.swift
+++ b/agterm/Control/ControlServer+SessionActions.swift
@@ -118,6 +118,78 @@ extension ControlServer: ControlActions {
         }
     }
 
+    func selectSession(_ target: String?, window: String?) -> ControlResponse {
+        resolver.resolveSession(target, window: window) { store, id in
+            store.selectSession(id)
+            return ControlResponse(ok: true, result: ControlResult(id: id.uuidString))
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


umputun (approved):
lgtm. Checked all 8 arms for behavior parity: error strings, resolver calls, and the workspace.rename trim all match the old inline bodies, and the fallthrough case-list stays exhaustive.

one optional nit, not a blocker: the workspace.rename guard in `ControlDispatcher.swift` open-codes `trimmingCharacters(in: .whitespacesAndNewlines)` + `!isEmpty`, but agtermCore already ships `String.trimmedOrNil` for exactly that. `guard let name = request.args?.name?.trimmedOrNil else { ... }` is identical and reuses the module normalizer.

PR #126, opened 2026-07-04
feat: pointing-hand cursor on link hover + validated link open (#125)

### LINKED ISSUE #125: Cursor doesn't change to a pointing hand when hovering a link (⌘-hover)
## Summary

When the mouse hovers a URL in the terminal (with ⌘ held, the macOS convention), the cursor does **not** change to a pointing hand — it stays the default arrow. Other macOS terminals (Terminal.app, iTerm2, Ghostty.app) show the pointing hand on ⌘-hover so it's clear the URL is clickable.

## Cause

agterm's libghostty action callback (`GhosttyCallbacks.action`) handles a curated set of `GHOSTTY_ACTION_*` and lets everything else fall through to `default: return false`. Three mouse/link actions libghostty actively sends are dropped:

- `GHOSTTY_ACTION_MOUSE_SHAPE` — "show the pointing hand over this link" (also the I-beam over the grid, crosshair/resize in the matching modes). Dropped → the cursor never changes.
- `GHOSTTY_ACTION_OPEN_URL` — the click-to-open. Dropped → libghostty falls back to opening it itself, with no scheme validation on agterm's side.
- `GHOSTTY_ACTION_MOUSE_OVER_LINK` — the hovered URL (not needed for the cursor; libghostty draws the underline itself).

## Expected

⌘-hover over a URL shows the pointing hand + link underline; ⌘-click (or double-click) opens it — matching the platform convention.

## Notes

Plain (no-modifier) hover is intentionally **not** in scope: libghostty hardcodes URL links to `hover_mods = ctrl/super` on macOS and the `link` config that could relax it is non-functional in the current build ("TODO: This can't currently be set!"), so plain-hover would need a fragile modifier-faking workaround. The standard ⌘-hover is the right behavior.

Have a fix ready.

vnazarenko:
Closes #125.

## What

Hovering a URL with ⌘ held now shows the **pointing-hand cursor** + link underline, and ⌘-click (or double-click) opens it — matching Terminal.app / iTerm2 / Ghostty.app. Before, agterm dropped the relevant libghostty actions, so the cursor never changed and the open path had no scheme validation.

## Cause

`GhosttyCallbacks.action` handled a curated set of `GHOSTTY_ACTION_*` and let the rest fall through to `default: return false`, dropping the whole mouse/link family: `MOUSE_SHAPE` (the cursor request), `OPEN_URL` (click-to-open), and `MOUSE_OVER_LINK`.

## Change (3 files, +80/-1)

- **`GhosttyCallbacks.swift`** — handle `GHOSTTY_ACTION_MOUSE_SHAPE` (→ apply an `NSCursor`) and `GHOSTTY_ACTION_OPEN_URL` (→ scheme-validated open).
- **`GhosttySurfaceView+Input.swift`** — `applyMouseShape` + `resetCursorRects` + an `nsCursor(for:)` map (pointer→`.pointingHand`, text→`.iBeam`, crosshair/grab/resize variants → their `NSCursor`, else `.arrow`); `openLink` opens only `http`/`https`/`mailto`/`ftp`/`file` via `NSWorkspace` (a terminal renders untrusted output, so an arbitrary custom scheme is never followed); and a `mouseExited` handler that reports negative coordinates so libghostty clears the hovered-link highlight when the pointer leaves the surface.
- **`GhosttySurfaceView.swift`** — a `mouseShape` property (`resetCursorRects` reads it) and `.mouseEnteredAndExited` on the tracking area.

## Scope note (plain hover deliberately excluded)

libghostty hardcodes URL links to `hover_mods = ctrl/super` on macOS, and the `link` config that could set `highlight = hover` is non-functional in the pinned build (Config.zig: "TODO: This can't currently be set!"). Making URLs react on **plain** (no-modifier) hover therefore needs an app-side ⌘-faking hack — which I prototyped and rejected: because a faked ⌘ never "releases", ghostty's link-underline clear (driven by a mods-change full repaint) never fires and the underline sticks. So this PR ships the standard ⌘-hover, which is the correct macOS-terminal convention anyway.

## Keep-in-sync

Pure libghostty-driven hover/render chrome — no `AppActions`/`AppStore` action behind it, nothing to drive over the control socket, so keep-in-sync **exempt** (like the other rendering gotchas).

## Testing

Build green, `swift test` (1075) green, swiftlint `--strict` clean. Cursor shape / link underline aren't accessibility-observable (Metal surface, not in the AX tree), so this is manually verified — confirmed in an isolated dev instance: ⌘-hover shows the hand + underline, moving off / releasing ⌘ / leaving the surface clears it, ⌘-click opens; text selection and a mouse-reporting TUI are unaffected. Not XCUITest-covered (consistent with the other libghostty rendering behavior, e.g. cursor focus).

### DIFF
--- a/agterm/Ghostty/GhosttyCallbacks.swift
+++ b/agterm/Ghostty/GhosttyCallbacks.swift
@@ -112,6 +112,21 @@ final class GhosttyCallbacks: @unchecked Sendable {
             let hidden = action.action.mouse_visibility == GHOSTTY_MOUSE_HIDDEN
             DispatchQueue.main.async { NSCursor.setHiddenUntilMouseMoves(hidden) }
             return true
+        case GHOSTTY_ACTION_MOUSE_SHAPE:
+            // libghostty requests a mouse cursor shape for this surface — the pointing hand over a detected
+            // link / OSC-8 hyperlink, the I-beam over the grid, resize/crosshair in the matching modes.
+            // Recover the surface and apply the mapped NSCursor via its cursor rects.
+            guard let view = surfaceView(from: target) else { return true }
+            let shape = action.action.mouse_shape
+            DispatchQueue.main.async { view.applyMouseShape(shape) }
+            return true
+        case GHOSTTY_ACTION_OPEN_URL:
+            // a click opened a detected link / OSC-8 hyperlink. Copy the URL out of the C string
+            // synchronously (only valid for this call), then open it scheme-validated on the main actor.
+            guard let view = surfaceView(from: target), let ptr = action.action.open_url.url else { return true }
+            let link = String(cString: ptr)
+            DispatchQueue.main.async { view.openLink(link) }
+            return true
         default:
             return false
         }
--- a/agterm/Ghostty/GhosttyCallbacks.swift
+++ b/agterm/Ghostty/GhosttyCallbacks.swift
@@ -115,16 +115,19 @@ final class GhosttyCallbacks: @unchecked Sendable {
         case GHOSTTY_ACTION_MOUSE_SHAPE:
             // libghostty requests a mouse cursor shape for this surface — the pointing hand over a detected
             // link / OSC-8 hyperlink, the I-beam over the grid, resize/crosshair in the matching modes.
-            // Recover the surface and apply the mapped NSCursor via its cursor rects.
+            // recover the surface and apply the mapped NSCursor via its cursor rects.
             guard let view = surfaceView(from: target) else { return true }
             let shape = action.action.mouse_shape
             DispatchQueue.main.async { view.applyMouseShape(shape) }
             return true
         case GHOSTTY_ACTION_OPEN_URL:
-            // a click opened a detected link / OSC-8 hyperlink. Copy the URL out of the C string
-            // synchronously (only valid for this call), then open it scheme-validated on the main actor.
-            guard let view = surfaceView(from: target), let ptr = action.action.open_url.url else { return true }
-            let link = String(cString: ptr)
+            // a click opened a detected link / OSC-8 hyperlink. copy the URL out of the LENGTH-DELIMITED C
+            // buffer synchronously (only valid for this call; `open_url.url` is NOT NUL-terminated — it is a
+            // Zig slice, so honor `.len` instead of `String(cString:)`, which would over-read into adjacent
+            // hyperlink storage), then open it scheme-validated on the main actor.
+            let openURL = action.action.open_url
+            guard let view = surfaceView(from: target), let ptr = openURL.url else { return true }
+            let link = String(decoding: UnsafeRawBufferPointer(start: ptr, count: Int(openURL.len)), as: UTF8.self)
             DispatchQueue.main.async { view.openLink(link) }
             return true
         default:
--- a/agterm/Ghostty/GhosttySurfaceView+Input.swift
+++ b/agterm/Ghostty/GhosttySurfaceView+Input.swift
@@ -195,12 +195,23 @@ extension GhosttySurfaceView {
         ghostty_surface_mouse_pos(surface, pt.x, pt.y, mods(event))
     }
 
+    /// The pointer entered the surface: restore libghostty's mouse position from the current point. Paired
+    /// with `mouseExited`'s `-1, -1` reset — without it, `scrollWheel` (which never sets `mouse_pos`) would
+    /// report a scroll at the stale `-1, -1` if you scroll right after re-entering, before any move, inside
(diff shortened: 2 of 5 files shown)


umputun (changes requested):
solid fix for the dropped actions, but a couple real bugs in the new code before this can merge:

1. **URL over-read** - `open_url` is length-delimited (`{ char* url; uintptr_t len }`, backed by a non-sentinel Zig `[]const u8`), so `String(cString: ptr)` reads past `len` into adjacent hyperlink storage until a stray NUL. Usually the link then fails to parse (garbage tail); worst case it opens a longer attacker-influenced URL or over-reads. Upstream Ghostty uses `Data(bytes:count:)` for exactly this field, while its `needle` (a bare `char*`) keeps `String(cString:)` - that asymmetry is the tell. Honor `len`:
```swift
let link = String(decoding: UnsafeRawBufferPointer(start: ptr, count: Int(action.action.open_url.len)), as: UTF8.self)
```

2. **`mouseExited` stale-position regression** - reporting `-1,-1` on exit without the paired `mouseEntered` reset introduces a bug: `scrollWheel` doesn't set `mouse_pos`, so scrolling right after re-entering the surface (before any mouse-move) reports the scroll at `-1,-1` inside a mouse-reporting TUI (vim/less/htop). Upstream pairs the exit with a `mouseEntered` that restores the position for exactly this reason. Add that override; optionally also upstream's `pressedMouseButtons != 0` guard so the `-1,-1` isn't emitted mid-drag.

3. **drop `file` from the allowlist** - `NSWorkspace.open` on a `file://` is the Finder-double-click path (LaunchServices), so a ⌘-click on an attacker-controlled `file:///…/X.app` or `.command` launches it. It's the only allowlisted scheme that hands off to a local executable/handler; http/https/mailto/ftp are fine. Drop `file`.

4. **hoist the scheme check + test it** - the allowlist predicate is pure Foundation. Pull it into `agtermCore` (e.g. `LinkPolicy.permittedURL(from:) -> URL?`) with a host-free unit test, leaving `NSWorkspace.open` app-side - same split as `ShellEscape`. A table test over raw inputs would also have caught #1's effect on parsing.

minor: the case-body comments at `GhosttyCallbacks.swift:118` and `:124` are capitalized; lowercase them to match the other arms.

vnazarenko:
Thanks — all four are real, fixed in 942f7eb.

1. **URL over-read** — you're right, `open_url` is length-delimited (Zig slice + separate `len`), and my `String(cString:)` over-read past `len`. Now:
   ```swift
   let link = String(decoding: UnsafeRawBufferPointer(start: ptr, count: Int(openURL.len)), as: UTF8.self)
   ```

2. **`mouseExited` stale position** — added the paired `mouseEntered` that restores `mouse_pos` from the entry point, so a scroll right after re-entering isn't reported at `-1,-1` in a mouse-reporting TUI. Also took your optional suggestion and guarded `mouseExited` with `pressedMouseButtons == 0` so it's not emitted mid-drag.

3. **`file://`** — dropped from the allowlist (it's the only scheme that hands off to a local executable via LaunchServices). Allowlist is now `http`/`https`/`mailto`/`ftp`.

4. **Hoisted + tested** — the predicate now lives in host-free `agtermCore.LinkPolicy.permittedURL(from:) -> URL?` (same split as `ShellEscape`), with a table test over raw inputs (web/mail permitted; `file://`, `javascript:`, custom schemes, schemeless, and trailing-garbage inputs rejected). App-side keeps only `NSWorkspace.open`.

Minor: lowercased the two case-body comments.

`swift test` 1079 green, swiftlint `--strict` clean, Debug build green. Ready for another look.

umputun (approved):
re-review - all four verified fixed:

1. `open_url` decodes to `.len` now (no over-read), and the string is copied out before the main-actor hop
2. paired `mouseEntered` restores mouse_pos + the `pressedMouseButtons == 0` guard, no more `-1,-1` scroll in a mouse-reporting TUI
3. `file://` out of the allowlist, single gated sink through `LinkPolicy.permittedURL`
4. predicate hoisted to host-free `LinkPolicy` with the table test

optional nits, none blocking:
- `LinkPolicy.permittedSchemes` has no caller outside `permittedURL`, could be `private`
- the `LinkPolicy` doc says "web/mail only" but the set still has `ftp`, one or the other
- `ftp` is mostly dead on modern macOS (no default handler), fine to drop if you're trimming

separate from this PR: the blanket `file://` reject means OSC-8 file links from dev tools (eza, cargo) still render clickable but no-op. I'll open a follow-up for a reveal-in-Finder path (`activateFileViewerSelecting`, non-executing), not something to add here.

lgtm, merging.

PR #132, opened 2026-07-05
refactor: Route window controls through dispatcher

melonamin:
## Summary
- route `window.rename`, `window.resize`, `window.move`, and `window.zoom` through `ControlDispatcher`
- keep `window.new`, `window.list`, `window.select`, `window.close`, and `window.delete` on the existing app-side path
- keep AppKit/window registry effects in `ControlServer` and add focused dispatcher tests for validation and response passthrough

## Validation
- `swift test --package-path agtermCore --filter ControlDispatcherTests`
- `swift test --package-path agtermCore`
- `make lint`
- `make build`
- `xcodebuild -project agterm.xcodeproj -scheme agterm -derivedDataPath build/DerivedData -only-testing:agtermUITests/ControlWindowUITests test` partially failed in unrelated header interaction tests; migrated command-specific tests in that class passed (`testWindowMove`, `testWindowResize`, `testWindowZoom`, targeting/list cases).

## Notes
- `window.new` and `window.list` behavior are intentionally unchanged.
- No comment-only lines were removed or changed in the production diff.

### DIFF
--- a/agterm/Control/ControlServer+WindowCommands.swift
+++ b/agterm/Control/ControlServer+WindowCommands.swift
@@ -75,10 +75,7 @@ extension ControlServer {
     /// Resolve a window id and resize its on-screen window to `width` x `height` points (frame size).
     /// The window must be open; a closed window errors (resize it after `window.select`). Control-native
     /// (no GUI surface — the native title bar already drags-to-resize).
-    func windowResize(_ target: String?, width: Int?, height: Int?) -> ControlResponse {
-        guard let width, let height, width > 0, height > 0 else {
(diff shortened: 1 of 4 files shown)


umputun (approved):
lgtm. Clean move of the four window commands into the dispatcher, guards and error strings preserved verbatim, and the invalid-input test asserting `actions.calls.isEmpty` is the right check for this kind of "move the guard" refactor.

one optional nit: the invalid-input coverage is a bit asymmetric. Missing-height, zero-width, and missing-y are tested, but not missing-width, `height:0`, or missing-x. Guards are symmetric so not a real gap, only worth adding if you want the mirror cases in.

PR #134, opened 2026-07-05
session new/move: --after/--before placement flags

olomix:
Placing a new session "right after the current one" over the control channel currently means appending it and then walking it up with repeated `session move --to up` — an O(N) shuffle that's visibly step-by-step in the sidebar and painful with dozens of sessions (discussion #123).

This adds reference-based placement: `--after <sid>` / `--before <sid>` on both `session new` and `session move`, so a session lands at the chosen slot in a single round-trip. Following the direction settled in the discussion, these are dedicated value flags rather than new `--to` values, keeping `--to` a clean closed `up|down|top|bottom` enum shared across `session.go`/`session.move`/`workspace.move`.

The anchor sid is resolved across the whole store, so it carries its own workspace — `--after`/`--before` are a self-contained placement mode, mutually exclusive with `--to` and with the workspace parameter, and cross-workspace placement (relocate + position in one shot) falls out for free with no "was the workspace set?" ambiguity. The headline case is `agtermctl session new --after active`.

It's dispatcher-first: arg validation and error strings live host-free in `ControlDispatcher`, the resolve-and-place side effect sits behind `ControlActions`, and the index math reuses the already table-tested `SidebarDrop` drop-index resolver (the after-this-row index plus the same-workspace post-removal off-by-one) and the existing clamped `AppStore.moveSession(_:toWorkspace:at:)`. No new command — the catalog stays at 50. `after`/`before` are new optional args, so existing clients and the `--to`/workspace forms are unaffected.

Covered by protocol round-trip, dispatcher, CLI-parse, and `SidebarDrop`/`AppStore` unit tests, plus XCUITest e2e over the socket (same-workspace, cross-workspace, create, and the error guards). The agent-skill docs, control-api rule, README, and site docs are updated to match.


umputun (approved):
lgtm, matches the #123 direction. Index math is right on the tricky cases (same-workspace off-by-one, cross-workspace, self/in-place no-op), and the e2e covers cross-workspace for both new and move.

one optional nit, not a blocker: the anchor-resolve block is near-identical between `createSession` and `placeSession` in `ControlServer+SessionActions.swift`. A small `resolveAnchorLocation` helper would DRY it. Fine as a follow-up.

PR #137, opened 2026-07-05
refactor: route remaining control commands through dispatcher

melonamin:
## Summary
- Route the remaining `ControlServer.dispatch` command arms through `ControlDispatcher`: `window.new/list/select/close/delete`, `session.search`, and `quick`.
- Keep AppKit, SwiftUI, window-cache, target-resolution, quick-terminal, and terminal-search side effects in the macOS app-side adapters.
- Preserve the nonisolated cache fast path for `window.list` and `tree --window` closed-window errors.

## Behavior Notes
- `session.search` still resolves target/window before validating `--to`, preserving prior error ordering and exact response strings.
- `quick` still delegates mode parsing to the app-side action, preserving `no open window` versus invalid-mode ordering.
- Window lifecycle/list commands call the same app-side helpers as before; the dispatcher only routes them.

## Validation
- `make test` (967 tests)
- `make lint`
- `make build`

## Note to maintainer

Looks like this is the last piece of my Linux port. Next, I'm going to bring the port up to speed with the latest UI/UX changes you've introduced over the last couple of days and report back if `agtermCore` needs more work to properly support the UI abstraction.

Thanks for bearing with me through this series of PRs!

### DIFF
--- a/agterm/Control/ControlServer+SurfaceIO.swift
+++ b/agterm/Control/ControlServer+SurfaceIO.swift
@@ -146,6 +146,19 @@ extension ControlServer {
     /// via libghostty's SEARCH_TOTAL callback; `searchTotal`/`searchSelected` are cleared before the query so
     /// the bounded main-actor poll waits for the FRESH count (not a stale prior needle's), then `count` + the
     /// "N of M" display string are returned in `text`.
+    func searchSession(_ target: String?, window: String?, text: String?, to: String?) async -> ControlResponse {
+        // Resolve first (cross-window when no `window`), then select + realize the surface; the realize
+        // path is async (bounded poll), so this can't go through the synchronous `resolveSession`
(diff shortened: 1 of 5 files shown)


umputun (approved):
lgtm. Checked the equivalence claims hold: `session.search` resolves the target before `--to` validation, and `quick` keeps the no-open-window check ahead of invalid-mode. Both hold because the dispatcher forwards the raw mode and `--to` instead of parsing them, so the app-side ordering stays put. The `window.list` fast path is untouched too, and the dispatcher suite is green (56 tests).

one non-blocker for a follow-up: the switch is exhaustive now and `dispatch` never returns nil, so `ControlServer.dispatch`'s trailing fallthrough arm is dead code and the `ControlResponse?` optional could drop. Fine to leave out of this PR, natural end of the series whenever you want it.

PR #141, opened 2026-07-05
Universal Binary Support for Intel and Apple Silicon

MrZoidberg:
This branch implements universal binaries (arm64 + x86_64) — a single build works natively on both Apple Silicon and Intel Macs.

• Build Pipeline: GhosttyKit and Xcode project now produce universal slices
• CI: Added Rosetta smoke check to verify Intel slice on Apple Silicon runners
• Release: Updated signing/notarization and Homebrew cask for both architectures  
• Docs: Updated README, website, and LLMs config to reflect universal support

⚠️ Testing Status

CI was not tested well and needs further testing — particularly:

• Building release from CI
• Homebrew installation verification on Intel

Let me know if you see any issues or concerns. Codex was used to implement this change.


umputun (commented):
appreciate the work here, but I'm going to pass on universal/Intel builds.

the main blocker is that I can't test or maintain it. I have no Intel hardware, so I can't verify the x86_64 app actually runs or renders. The CI "Rosetta smoke check" only runs `agtermctl tree` (the control CLI), never the app or its Metal rendering, so it doesn't really show the terminal works on Intel, and the PR itself notes CI wasn't tested well. I can't stand behind a slice nobody has confirmed works.

it also couples the GhosttyKit build (zig SHA + Metal toolchain, pinned deliberately) to a universal target I can't validate on every libghostty bump, and dropping `-Dxcframework-target=native` pulls in iOS/iOS-sim slices agterm doesn't use.

timing works against it too: macOS 26 Tahoe was the last macOS for Intel hardware, and Rosetta 2 is being wound down (warnings in 26.4, general removal in macOS 28), so the Intel base is capped and shrinking. Not worth the ongoing maintenance for a native GUI terminal I can't run on that hardware.

closing as not planned.

thx for the effort, the release-script lipo verification was a nice touch.

PR #144, opened 2026-07-06
fix: drop the two startup subshells from the zsh agent-status integration

aldobrynin:
`integration.sh` is sourced by every interactive shell inside an agterm session.
Computing `_ags_dir` via `$(cd "$(dirname …)" && pwd)` forks three processes
(two command substitutions plus the `dirname` exec) on every shell startup.

In the zsh branch this can be a pure parameter expansion: `${_ags_self:A:h:h}` —
`:A` absolutizes and resolves symlinks (same result as the `cd`-and-`pwd` dance),
`:h:h` strips `/shell/integration.sh`. The bash branch keeps the portable subshell
form, mirroring the existing shell-specific split used for `_ags_self` itself.

## Impact

Measured on an M4 Max (macOS 26.5, warm shell, interleaved medians of 15 runs):

| variant | startup median |
|---|---|
| stock integration | 43.7 ms |
| integration not sourced | 31.2 ms |
| patched integration | 31.1 ms |

The change recovers the entire ~12.5 ms — the two command substitutions are
effectively the file's whole startup cost.

## Change

```diff
--- a/agterm/Resources/agent-status/shell/integration.sh
+++ b/agterm/Resources/agent-status/shell/integration.sh
@@ -17,11 +17,12 @@
 # Locate agterm-agent-status.sh relative to this file (layout is preserved on
 # install: shell/integration.sh sits next to agterm-agent-status.sh's dir).
 if [ -n "${ZSH_VERSION:-}" ]; then
   _ags_self="${(%):-%x}"
+  _ags_dir="${_ags_self:A:h:h}"
 else
   _ags_self="${BASH_SOURCE[0]}"
+  _ags_dir="$(cd "$(dirname "$_ags_self")/.." >/dev/null 2>&1 && pwd)"
 fi
-_ags_dir="$(cd "$(dirname "$_ags_self")/.." >/dev/null 2>&1 && pwd)"
 : "${AGTERM_AGENT_BIN:=$_ags_dir/agterm-agent-status.sh}"
```

## Behavior

Unchanged:

- same `_ags_dir` on a standard install (verified old vs new expression byte-identical);
  `_ags_preexec`/`_ags_precmd` register as before
- `AGTERM_AGENT_BIN` / `AGTERM_AGENT_RE` overrides and the `AGTERM_SESSION_ID`
  guard untouched
- the installer's `AGTERMCTL` baking is unaffected (it patches the wrapper,
  not this file)
- `integration.fish` doesn't have this pattern

No changelog edit per repo convention (release-only).

### DIFF
--- a/agterm/Resources/agent-status/shell/integration.sh
+++ b/agterm/Resources/agent-status/shell/integration.sh
@@ -18,7 +18,7 @@
 # install: shell/integration.sh sits next to agterm-agent-status.sh's dir).
 if [ -n "${ZSH_VERSION:-}" ]; then
   _ags_self="${(%):-%x}"
-  _ags_dir="${_ags_self:A:h:h}"
+  _ags_dir="${_ags_self:a:h:h}"
 else
   _ags_self="${BASH_SOURCE[0]}"
   _ags_dir="$(cd "$(dirname "$_ags_self")/.." >/dev/null 2>&1 && pwd)"
--- a/agterm/Resources/agent-status/shell/integration.sh
+++ b/agterm/Resources/agent-status/shell/integration.sh
@@ -18,10 +18,11 @@
 # install: shell/integration.sh sits next to agterm-agent-status.sh's dir).
 if [ -n "${ZSH_VERSION:-}" ]; then
   _ags_self="${(%):-%x}"
+  _ags_dir="${_ags_self:A:h:h}"
 else
   _ags_self="${BASH_SOURCE[0]}"
+  _ags_dir="$(cd "$(dirname "$_ags_self")/.." >/dev/null 2>&1 && pwd)"
 fi
-_ags_dir="$(cd "$(dirname "$_ags_self")/.." >/dev/null 2>&1 && pwd)"
 : "${AGTERM_AGENT_BIN:=$_ags_dir/agterm-agent-status.sh}"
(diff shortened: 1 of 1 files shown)


umputun (commented):
checked the fork drop and confirmed bash never evaluates the zsh-only line. It's inside the `ZSH_VERSION` branch and macOS bash 3.2 sources the file clean.

one non-blocking suggestion: `${_ags_self:a:h:h}` (lowercase `:a`) instead of `:A`. `:A` resolves symlinks, so it diverges from the old `cd/pwd` form when integration.sh is sourced through a symlink. It then points `_ags_dir` at the symlink target's parent instead of the sourced location's parent, so `AGTERM_AGENT_BIN` can default to a different, un-baked `agterm-agent-status.sh`. The installer deep-copies so a standard install never hits this, but `:a` is byte-exact to the old result on every layout at the same zero-fork cost, so the "byte-identical" claim holds on symlinked setups too.

quick check: with `install/shell/integration.sh -> target/shell/integration.sh` sourced via the symlink, old and `:a` both resolve to `install` while `:A` resolves to `target`.

aldobrynin:
@umputun good catch, thanks for the review. applied the fix for symlinks

umputun (approved):

PR #146, opened 2026-07-06
fix: unreadable selected session label on light themes (light:/dark: theme form)

### LINKED ISSUE #145: Selected session label unreadable on light themes
**Problem**

In the sidebar, the currently selected session's label is drawn near-white on a light fill, so on a light theme it is unreadable — you can't tell which session is selected or read its name. Unselected rows render fine. On a dark theme the selected label reads fine.

**When it happens / configuration**

It appears when the theme is configured with the appearance auto-switch syntax `theme = light:X,dark:Y` in `ghostty.conf` and the system is in Light mode.

With a single theme set instead (e.g. `theme = GitHub Light Default`) the selected label is readable — even though that is a light theme. So the trigger is the `light:,dark:` syntax, not the theme's lightness.

Ghostty supports different theme configuration
> To specify a different theme for light and dark mode, use the following syntax: light:theme-name,dark:theme-name. For example: light:Rose Pine Dawn,dark:Rose Pine

https://ghostty.org/docs/config/reference#theme

**Environment**

agterm 0.7.1, macOS 26.5.1, no global ghostty config.

**Steps to reproduce**

1. In `ghostty.conf`: `theme = light:<light theme>,dark:<dark theme>` (e.g. `light:Alabaster,dark:Gruvbox Dark Hard`), system in Light mode.
2. Select a session in the sidebar.
3. Its label is invisible (near-white text on a light fill).

**Screenshots**

Selected (unreadable):

Image

Not selected, focus on other worktree (fine):

Image

bigspawn:
Fixes #145.

`resolveSelectionColors` used the raw `theme` value as a theme-file name, so the `light:X,dark:Y` auto-switch form never matched a file → selection colors nil → selected sidebar row fell back to white-on-white on light themes. Resolve the active variant (new host-free `ThemeName.resolved`) before the lookup.

Tests: `ThemeNameTests` (7 cases), `agtermCore` green, `swiftlint --strict` clean. Verified live with `theme = light:Alabaster,dark:Gruvbox Dark Hard` in Light mode.

Screenshot:

SCR-20260706-lomw

### DIFF
--- a/agterm/Ghostty/GhosttyApp.swift
+++ b/agterm/Ghostty/GhosttyApp.swift
@@ -295,9 +295,9 @@ final class GhosttyApp {
             ?? selectionBackground.map(Self.contrastingText(for:))
     }
 
-    /// Whether the active appearance is dark, for resolving the `theme = light:…,dark:…` form. Uses
-    /// `NSApp` when available; `resolveThemeColors` runs from `init` before `NSApp` is set, so it falls
-    /// back to the system setting rather than force-unwrapping the nil `NSApp`.
+    /// Whether the active appearance is dark, for the `theme = light:…,dark:…` form.
+    /// `resolveThemeColors` runs from `init` before `NSApp` exists, so fall back to
+    /// the system setting instead of force-unwrapping nil `NSApp`.
     private static func currentAppearanceIsDark() -> Bool {
         if let appearance = NSApp?.effectiveAppearance {
             return appearance.bestMatch(from: [.darkAqua, .aqua]) == .darkAqua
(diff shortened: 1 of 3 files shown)


umputun (approved):
lgtm, merging.

one known limitation, not for this PR: with `theme = light:X,dark:Y` the selection + bg/fg colors resolve only at load and on config reload, so a live macOS light/dark toggle updates the terminal but leaves the sidebar selection pill on the load-time variant until the next reload. It's pre-existing (bg/fg already work this way) and strictly better than the previous nil case, so I'll track it as a separate follow-up.

PR #151, opened 2026-07-06
fix(core): support agtermctl tests on Linux

melonamin:
## Summary

- Make the shared `agtermctlKit` Unix socket client compile on Linux by selecting Darwin or Glibc APIs at build time.
- Update socket-client tests to use the same platform abstraction.
- Relax shared config-path tests so they use an installed POSIX-like shell instead of assuming `/bin/zsh` exists.

## Why

The shared core package is host-free, but a few CLI support and test files still assumed Darwin APIs or a macOS shell layout.

This keeps the shared package buildable and testable on Linux while preserving the Darwin path.

## Validation

- `swiftlint lint --config .swiftlint.yml agtermCore/Sources/agtermctlKit/SocketClient.swift agtermCore/Tests/agtermCoreTests/ConfigPathsTests.swift agtermCore/Tests/agtermctlKitTests/SocketClientTests.swift`
- `git diff --check v0.8.2`

### DIFF
--- a/agtermCore/Sources/agtermctlKit/SocketClient.swift
+++ b/agtermCore/Sources/agtermctlKit/SocketClient.swift
@@ -1,4 +1,8 @@
+#if canImport(Darwin)
 import Darwin
+#elseif canImport(Glibc)
+import Glibc
+#endif
 import Foundation
 import agtermCore
 
@@ -40,13 +44,18 @@ struct SocketClient {
 
     /// Open and connect a `AF_UNIX` stream socket to `path`.
     private func connect() throws -> Int32 {
-        guard path.utf8.count < 104 else {
+        var addr = sockaddr_un()
+        let pathCapacity = MemoryLayout.size(ofValue: addr.sun_path)
+        guard path.utf8.count < pathCapacity else {
             throw SocketClientError("socket path too long (\(path.utf8.count) bytes): \(path)")
         }
+        #if canImport(Darwin)
         let fd = socket(AF_UNIX, SOCK_STREAM, 0)
+        #else
+        let fd = socket(AF_UNIX, Int32(SOCK_STREAM.rawValue), 0)
+        #endif
         guard fd >= 0 else { throw SocketClientError("socket() failed: \(String(cString: strerror(errno)))") }
 
-        var addr = sockaddr_un()
         addr.sun_family = sa_family_t(AF_UNIX)
         let pathBytes = path.utf8CString
         withUnsafeMutablePointer(to: &addr.sun_path) { dst in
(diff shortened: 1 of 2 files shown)


melonamin:
@umputun this one leaks Linux details into the repo. I can make it a bit better, for `zsh` just say that it is required on Linux for `agterm` development and drop it. For the socket problem we can wrap the whole macOS implementation in `#if darwin`, it will stop failing on Linux.

Thoughts?

melonamin:
@umputun I need your input please, this is blocking me from starting releasing Linux builds in sync with macOS.

umputun (commented):
on the two you asked about: keep the socket code as-is, and yes to your zsh idea.

one general thing first: Linux code creeping into the shared core isn't a pattern I want. Per #58 the Linux app is your fork, and I don't want this repo growing a second platform's plumbing it never compiles. This case is small and transparent to the macOS build, so I'll make an exception, but it's an exception, not the door opening for more.

**socket**: keep what you've got, don't wrap it in `#if darwin`. That wrap compiles but leaves the client a no-op on Linux, so `agtermctl` couldn't actually connect there, which defeats the point of a Linux build. The per-platform Darwin/Glibc selection you already wrote is the idiomatic shim and keeps it functional. So `SocketClient.swift` and the socket bits in `SocketClientTests.swift` stay as-is.

**zsh**: go with your other idea. Drop the `posixLikeShellPath`/`writeEditorRC` relaxation in `ConfigPathsTests.swift`, keep the hardcoded `/bin/zsh`, and just require zsh for Linux dev (note it in your fork's setup). Less test machinery for a macOS-identical result. Dropping it also clears the four Copilot notes on that file (`/bin/sh` being dash and not handling `-ilc`, the silent `guard ... return` skips).

**CI / ownership**: no Linux job here. My CI is macOS-only, so I won't add one just to compile these branches. That means I'm taking the portability as best-effort. If a later `agtermCore` change breaks the Linux `#if`, I won't catch it. You'll see it in your Linux build and fix it there. I'm hosting this sliver upstream only so you're not re-patching `SocketClient` on every pull, not signing up to keep Linux green.

one non-blocking thing Copilot flagged, on code you're keeping: the `< 104` path cap in `SocketClient.swift` is Darwin's `sun_path` size, Linux's is 108, so it's a touch over-strict on Linux. No macOS impact. Your call whether to make it per-platform, since the Linux side is yours to keep working.

revert the `ConfigPathsTests` change, leave the socket code, and I'll merge it.

melonamin:
Left only socket conditionals, Copilot's concerns addressed

umputun (commented):
does what we agreed: per-platform socket selection kept functional (not a `#if darwin` no-op), `ConfigPathsTests` reverted. Darwin path is byte-identical, and the `SocketClientTests` changes assert real behavior on both platforms rather than skipping on one. The `< 104` cap you left as my call is resolved too, `MemoryLayout.size(ofValue: addr.sun_path)` is now per-platform (104 Darwin / 108 Linux).

one nit, non-blocking:

`systemConnect` (`SocketClient.swift:201`) is a private free function called only from `SocketClient.connect()`. Per the methods-over-standalone rule it fits better as a `private static func systemConnect(...)` on `SocketClient` (called `Self.systemConnect(...)`). The three shims in the test file are fine as free functions since they're shared across the stub classes.

PR #158, opened 2026-07-07
feat: Add terminal zoom toggle

melonamin:
## Summary

- Add a host-free terminal zoom controller and default `cmd+shift+return` toggle.
- Render the active terminal surface full-window above sidebar/chrome for primary, split, scratch, overlay, and quick terminal surfaces.
- Expose stable `tree.surfaces[].id` values and a `surface.zoom` socket command, plus `agtermctl surface zoom [show|hide|toggle]`.
- Preserve split/layout state by keeping zoom as a rendering layer and by making active-surface toggle/hide clear the current zoom target without changing pane ratios, focus, sidebar state, or split/scratch visibility.
- Update the bundled agent skill with the CTL-driven surface zoom workflow.

Related #154

## Validation

- `make build`
- `git diff --check`


umputun (changes requested):
three things before merge, plus a rebase - it's behind #173/#174/#175 (and the reopen-recent merge) now and conflicting.

**1. zoom needs a `tree` read-back.** `surface.zoom` sets per-window state but nothing reports whether/which surface is zoomed. `tree.surfaces[].active`/`visible` are derived from session flags (overlay/scratch/splitFocused), identical zoomed or not (`ControlSurfaceNode` is just id/kind/active/visible, and `controlTree` builds it without consulting `TerminalZoomController`), so a script can't record-then-restore or check "is it already zoomed". That's the same write-only gap `session.overlay.resize` shipped with. Add a top-level `ControlTree.zoomedSurface: String?` through the existing `quickVisible` closure seam + a round-trip test. The `surfaces` addressing list isn't the state read-back.

**2. keeping zoom modal is fine, but grey out the gated items.** The `uiActionsEnabled` gate turns the gated File/View/Navigate items into silent no-ops while zoomed with no feedback. Add `.disabled(terminalZoomActive)` to them so they read as disabled, on top of the action-layer gate that also catches the keymap paths.

**3. drop the opaque backing in the zoom layer.** The zoom host's `.background(terminalColor)` (copied from the quick-terminal panel, `WindowContentView+Zoom.swift`) flips a translucency window from translucent to opaque on zoom and back. The deck behind is already at opacity 0, so no backing is needed - let the window backing show through so zoom matches the un-zoomed terminal.

minor:
- `control-api.md:220` still says "the public command count stays 53", should be 54.
- `ControlSurfaceNode.visible` reads false for panes behind a *floating* overlay (visually on screen) - fine since addressing uses id/kind, but worth a doc caveat.
- no e2e for the `--window`-scoped error paths (`no such surface`, `window not open`), only the happy path.

no lifetime/focus/deck-corruption issues - the value-type target resolved fresh by id self-heals when the surface goes away, and the stable-ZStack-placeholder holds the NSSplitView invariant (ratio preservation covers it).

melonamin:
Updated, all addressed

umputun (commented):
clean re-review. The three fixes are in and hold: traced the render-layer invariant (zoom never mutates split ratio, focus, sidebar, or scratch/overlay visibility), `zoomedSurface` reads back live per-window off the real controller, the menu grey-out also gates the keymap and NSEvent-monitor paths (`SessionSwitcher`/`PaneShortcuts`), and the opaque zoom backing is gone. Keep-in-sync is complete and the command count is consistent at 58.

one doc nit:

`SKILL.md:87` still says the tree carries "five read-only top-level fields" (`idleMs`/`autoFollowMs`/`sidebarVisible`/`sidebarMode`/`quickVisible`) and omits `zoomedSurface`. That contradicts `SKILL.md:140` (which documents `zoomedSurface`) and `reference.md`, both of which count six. Bump it to six and add `zoomedSurface`.

minor, your call: the new public API in `TerminalZoom.swift` (`TerminalZoomController`, `TerminalZoomRegistry`, `set`/`toggle`/`resolveTarget`/`isTargetValid`) carries no `///` docs, while this same PR documents every new `ControlProtocol` type. A doc pass on the controller would match.

fyi, not a finding: an adversarial pass flagged `testZoomedSurfaceTreeReadBackAndScopedErrorPaths` as failing on a `window.new`/`window.close` race near line 129. I ran it down through `resolveOpenWindow`/`windowClose`. The test already gates on materialization implicitly (the wrong-window assert at :124 only holds once the store is loaded, and with `continueAfterFailure=false` it can't reach :129 otherwise), `window.close` self-synchronizes, and the closed-window path returns a deterministic error. The empty read was an XCUITest environment artifact in my worktree, not a defect.

Issue #131, opened 2026-07-05
reveal file:// links in Finder instead of no-op (follow-up to #126)


umputun:
follow-up from #126, which deliberately dropped `file://` from the click-to-open allowlist.

**problem**: OSC-8 `file://` hyperlinks (eza `--hyperlink`, cargo, gh, some compilers) still render clickable, the pointing-hand cursor + underline come from libghostty's link detection, independent of our open allowlist. But the ⌘-click silently no-ops now, so a link looks clickable and does nothing.

**why #126 dropped it**: `NSWorkspace.open` on a `file://` goes through LaunchServices (the Finder double-click path), so a ⌘-click on an attacker-controlled `file:///…/X.app` or `.command` would launch it. A terminal renders untrusted output, so a blanket open on `file://` is a click-to-execute gadget.

**proposal**: for `file://` specifically, reveal in Finder via `NSWorkspace.activateFileViewerSelecting(url)` instead of opening. It selects the file in Finder without executing anything, so it keeps the launch gadget closed while restoring the dev-tool case. `http`/`https`/`mailto` keep the current open path.

implementation: `LinkPolicy` could return the parsed `file://` URL with a "reveal" disposition (vs "open"), and `openLink` branches on it. Keeps the policy host-free and unit-testable.

umputun:
@vnazarenko this follows on directly from your #126, want to take it as a continuation? No pressure, fine to leave it for anyone to pick up.

i-kozlov:
@umputun — an optional design question following up on this issue.

**Use case:** I'd like a ⌘-click on a `file://` link to *open* plain, non-executable documents (`.md`, `.txt`, `.pdf`, `.log`) in their default app, not just reveal them. Dev tools that emit OSC-8 file links (logs, docs, source listings) make this a common want; today the click no-ops.

This builds on your reveal-in-Finder plan here — I'd only add an **opt-in, default-off** *open* layer on top.

- **Reveal in Finder** = this issue: unresolved `file://` → `activateFileViewerSelecting` instead of today's no-op.
- **Opt-in open**, layered on: an agterm setting (like `newSessionDirectory`), off by default, listing extensions a ⌘-click may open. Guards, so it stays narrow rather than "safe":
  - resolve the final filesystem target first and open **only a regular, non-executable, non-package file**; symlinks/aliases/bundles/directories fall back to reveal (never `NSWorkspace.open`, so no resolve-to-`.app` launch);
[...]

PR #162, opened 2026-07-07
reveal file:// links in Finder instead of no-op (#131)

i-kozlov:
Follow-up to #126 / #131.

After #126 dropped `file` from the click-to-open allowlist, a ⌘-click on a `file://` OSC-8 hyperlink became a no-op (the link still shows the hand/underline). `eza`, GNU `ls --hyperlink`, and many compilers emit legitimate `file://` links, so this is a dead click in practice.

This makes a ⌘-click on a `file://` link **reveal the file in Finder** (`NSWorkspace.activateFileViewerSelecting`) instead of doing nothing. It never *opens* the file: opening goes through LaunchServices (the double-click path), so a click on an attacker-controlled `file:///…/X.app` or `.command` would launch it — reveal only selects it, executing nothing. Actually opening a file stays a separate, explicit action. `http`/`https`/`mailto` keep the existing open path.

Reveal-only, matching the scope agreed in #131 — no opt-in open layer, no settings, no allowlist.

### How
- `LinkPolicy` now maps a raw link to an `open` / `reveal` / `ignore` disposition (host-free, Foundation-only, unit-tested); `openLink` is thin AppKit glue that switches on it.
- Host handling for `file://`: an empty host (what `eza` emits) or this machine's own name (what GNU `ls --hyperlink` emits) → reveal; a non-local host → ignore, so a stray `file://server/…` can't trip a Finder network/SMB mount. A UNC-style `//`-path (`file:////server/share`) is ignored too — it hides a remote target in the path where the host check can't see it.
- This-machine names come from `gethostname()` (a pure syscall) rather than `Host.current()` / `ProcessInfo.hostName`, which resolve via mDNS/Bonjour and would trip the macOS "find devices on local networks" permission prompt on first click.

### Tests
`LinkPolicy` disposition matrix: web → open; local `file://` → reveal (empty host, localhost, own hostname, case-insensitive, trailing FQDN dot, single-slash `file:/`, `.app` → reveal-not-launch); non-local host + UNC path → ignore; other/schemeless/garbage → ignore; plus host-name expansion and a default-parameter smoke test. `swift test` green, `make lint --strict` clean, app builds.

### Docs
README + `site/docs.html` note the new behavior.


umputun (changes requested):
two changes worth making before merge, plus a couple of minor notes.

**1. `permittedURL(from:)` is dead code now.** `openLink` moved to `disposition(for:)`, and `disposition`'s `.open` arm already does what `permittedURL` did (same `permittedSchemes`, same `URL(string:)`). Nothing in production calls it anymore, only its own 4 tests. Can you drop it and fold the one case still worth keeping, `trailingGarbageDoesNotYieldAWebURL`, onto `disposition`? The web/scheme cases are already covered by `webSchemesOpen` / `otherSchemesIgnore`.

**2. reveal hands Finder the host-bearing URL.** `.reveal(url)` keeps the original `file://<host>/path` authority and passes it straight to `activateFileViewerSelecting`, so the reveal leans on Finder's host handling instead of the decision the policy already made. Two edges fall out of that:
- a local path can still name a network target the authority check never sees. `file:///net/<host>/share` (and `/Network/Servers/...`) classify as `.reveal`, and Finder walking that path can invoke autofs. Low severity (NFS only, `/net -hosts` is commented out in the default `auto_master`), but it's the same "stray link can't trigger a network mount" the docs promise.
- no test exercises a host-bearing reveal, so it's unconfirmed that a `file://<hostname>/...` link (what `ls --hyperlink` emits) reveals rather than no-ops.

normalizing to a hostless local path closes both. After a link classifies as local, return `URL(fileURLWithPath: url.path(percentEncoded: false))` and reject a path starting `/net/` or `/Network/` (and empty/relative paths). Then `activateFileViewerSelecting` only ever sees a plain local path. Worth a test for the host-stripped reveal URL and for the `/net` rejection.

**minor, optional:**
- `LinkPolicy.swift:44`, `String(cString: buffer)` on a `[CChar]` array hits the deprecated array overload and adds a new build warning. The rest of the tree passes pointers, so it doesn't warn there. `String(decoding:as: UTF8.self)` over the NUL-trimmed bytes matches the byte-buffer convention already used elsewhere and clears it.
- `LinkPolicy.swift:76`, the `// A UNC-style path...` comment is capitalized while the other inline comments in the file are lowercase.

i-kozlov:
Pushed fixes for all four review points (host-stripped + dot-normalized reveal that rejects empty/relative/UNC paths and the `/net`, `/Network`, `/home` automount roots after `..` normalization; removed the now-unused `permittedURL`; `String(decoding:as: UTF8.self)` over NUL-trimmed bytes; comment casing).

One residual edge worth flagging on point 2, kept as a deliberate trade-off rather than fixed here: `disposition` only dot-normalizes the path, it does not resolve symlinks (stays host-free, no `stat`). So a local symlink like `/tmp/link -> /net` is not caught by the automount guard — the "no stray link can trigger a network mount" guarantee holds for literal paths but not for a symlink indirection. Closing that would mean touching the filesystem, which is broader than this reveal policy and closer to the separate opt-in open-file work.

umputun (commented):
Round-1 points all addressed: `permittedURL` gone, reveal is host-stripped and normalized, `String(decoding:)`, comment casing. One gap left in the automount guard, worth closing first:

`isAutomountPath` denies `/net` / `/Network` / `/home`, but `/home` is a symlink and `auto_home` is actually mounted at `/System/Volumes/Data/home` (`mount` shows `map auto_home on /System/Volumes/Data/home`). So `file:///System/Volumes/Data/home/<user>/x` classifies as `.reveal`, and Finder walking that folder can still trigger the auto_home mount, the same network mount the `/home` entry exists to stop. It's a literal path (no symlink needed), so it stays host-free to close: add the `/System/Volumes/Data/...` canonical roots (at least `/home`; the `/net` and `/Network/Servers` equivalents too for symmetry) to the denylist, plus a test that `file:///System/Volumes/Data/home/<name>/...` is ignored. Real network impact is only on directory-bound Macs, but it's the exact guarantee the guard makes.

The symlink residual you flagged (`/tmp/link -> /net`) is fine to leave. Catching it needs a `stat`/`realpath`, which drops the host-free property; it needs local FS write to set up (stronger than a link click); and reveal selects the leaf in its parent rather than traversing the target. Keeping the classifier `stat`-free is the right call.

Otherwise lgtm.

i-kozlov:
Pushed the round-2/round-3 fixes plus a hardening pass. Two places where the implementation intentionally diverges from the literal suggestion — flagging so the reasoning is on record:

**1. Normalization is purely lexical, not `URL(fileURLWithPath: url.path(percentEncoded: false))`.** Building a file URL and standardizing it (`standardizedFileURL`) consults the filesystem — verified locally: it stats the path for a directory hint (existing dirs come back with a trailing slash, missing ones don't). That stat runs on the very automount path *before* the denylist check, and a stat inside autofs can itself trigger the mount the denylist exists to prevent. So the classifier now collapses `.`/`..` with a string-only normalizer (`lexicallyNormalizedAbsolutePath`) and never touches the filesystem before deciding. Host-stripping is kept — reveal still gets a plain hostless path.

**2. The denylist also covers the canonical Data-volume roots.** `/home` is a firmlink and `auto_home` actually mounts at `/System/Volumes/Data/home`, so a literal `file:///System/Volumes/Data/home/<user>` slipped past a plain `/home` entry. Added `/System/Volumes/Data/{home,net,Network/Servers}` — kept narrow (not the whole `/System/Volumes/Data`, which backs every real file, so `/System/Volumes/Data/Users/...` still reveals).

The symlink residual you were fine leaving is kept stat-free by design; there's now a regression test (`revealDoesNotResolveSymlinks`) that pins the reveal path to the link itself, not the resolved target (a swap to `resolvingSymlinksInPath()` makes it fail).

umputun (approved):
round-2 automount gap is closed and tested. The canonical `/System/Volumes/Data/{home,net,Network/Servers}` roots are denied now, and `/System/Volumes/Data/Users/...` still reveals so the guard stays narrow. Switching to the lexical normalizer is the right move: standardizing would stat the autofs path before the denylist runs.

ran the classifier through the bypass surface too (case, doubled slashes, `..` into /net, percent-encoded, non-local host as exact-match, UNC), all closed. Every guard runs on the decoded path and the same normalized string is what reaches Finder. The symlink residual is the one you flagged and the regression test pins it.

lgtm, thx

PR #166, opened 2026-07-07
docs: agent-status hooks report to the wrong session when AGTERM_* leaks through a daemon

paskal:
## What

Documents a footgun with the agent-status hooks: a long-lived process started from inside an agterm session (a tmux server is the common carrier — directly, or via a tmux-backed session manager like agent-deck) captures the spawning session's `AGTERM_*` environment and hands it to every child it ever creates. Status hooks in those children resolve `$AGTERM_SESSION_ID` to the session that started the daemon, so one session's glyph blinks while the work happens in a completely different one.

tmux makes it especially sticky: the variables land in the server's **global environment** (`tmux show-environment -g`), so every pane on that server inherits them regardless of which client attaches, and they survive session restarts.

## Changes

- `agent-skill/SKILL.md` — a warning in the "Am I inside agterm?" env section: the variables are inherited by daemons, scrub with `env -u` before starting one.
- `agent-skill/troubleshooting.md` — a new symptom entry, "The agent-status glyph updates the wrong session": diagnosis (`ps eww <pid>` for the agent's real env, `tmux show-environment -g`), the fix for a poisoned server (`tmux set-environment -g -r` for each var + pane respawn), and prevention.
- `docs/troubleshooting.md` — the user-facing mirror bullet under "Other common issues", per the keep-in-sync convention.

## How it was found

Hit in production: 17 agent-deck-managed Claude Code sessions all reported status to the single agterm session that had restarted the fleet — its glyph blinked on every turn of every agent, while their own sessions stayed idle. The tmux server had been auto-started from that session's shell, poisoning the global env; `env -u` on the CLI alone did not help because the server, not the CLI, was the carrier.


umputun (commented):
one thing before merge: the scrub example in SKILL.md lists 5 vars but the tmux loop in troubleshooting.md lists 6 - the SKILL.md one drops `AGTERM_PANE`.

`AGTERM_PANE` is injected per-surface by `SurfaceEnvironment.session(...)` and `agterm-agent-status.sh` forwards it as `--pane`, so it belongs in the scrub list too. Can you add `-u AGTERM_PANE` to the SKILL.md example so both match?

not a correctness bug (scrubbing `AGTERM_SESSION_ID` alone already no-ops the hook), just aligning the two lists.

paskal:
added in 5de0c3b — SKILL.md scrub example now carries `-u AGTERM_PANE`, both lists match at six vars.

PR #174, opened 2026-07-08
feat: reopen recently closed items

melonamin:
## Summary

Adds reopen support for accidentally or recently closed sessions and workspaces.

- Grace-period undo for GUI session/workspace closes, including Cmd-Z support and a sidebar toast.
- Persistent File > Open Recent entries for closed sessions and workspaces.
- A settings toggle to disable the grace-period undo while keeping recent reopen behavior.

## Technical notes

- Core owns the host-free pending-close and recent-closed state; SwiftUI/AppKit only render menus, toast UI, and keyboard monitoring.
- Recent session/workspace restore reuses the existing snapshot model so restored state matches window restore fidelity.
- GUI grace-period closes keep surfaces alive until finalization; finalized closes record persistent recent items.

## Demo

https://github.com/user-attachments/assets/2734a157-9e28-44b3-8750-97f18072a8b7

p.s. decided for features like this it it easier to start with PR right away, then discuss, @umputun I hope you are Ok with this approach

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,7 +309,7 @@ The bindable built-in action names are:
 new_window         rename_window      delete_window
 new_workspace      rename_workspace   delete_workspace
 new_session        open_directory     rename_session
-close_session      clear_status
+close_session      reopen_recent      undo_close         clear_status
 increase_font_size decrease_font_size reset_font_size
 toggle_split       toggle_scratch     toggle_search
 toggle_sidebar     toggle_flag        toggle_flagged_view
--- a/agterm/AppActions+Palette.swift
+++ b/agterm/AppActions+Palette.swift
@@ -28,7 +28,9 @@ extension AppActions {
             sidebarShowsFlaggedOnly: activeStore?.sidebarMode == .flagged,
             activeSessionFlagged: activeStore?.activeSession?.flagged == true,
             hasFocusedWorkspace: activeStore?.focusedWorkspaceID != nil,
-            activeSessionHasSplit: activeStore?.activeSession?.hasSplit == true
+            activeSessionHasSplit: activeStore?.activeSession?.hasSplit == true,
+            hasPendingClose: activeStore?.pendingCloseSummary != nil,
+            hasRecentClosed: !library.recentClosedItems.isEmpty
         )
     }
 
@@ -47,6 +49,8 @@ extension AppActions {
         case .renameSession: renameActiveSession()
         case .renameWorkspace: renameActiveWorkspace()
         case .closeSession: closeActiveSession()
+        case .reopenRecent: openLatestRecentClosed()
+        case .undoClose: undoClose()
         case .clearStatus: clearActiveSessionStatus()
         case .previousSession: selectPreviousSession()
         case .nextSession: selectNextSession()
--- a/agterm/AppActions.swift
+++ b/agterm/AppActions.swift
@@ -1,5 +1,6 @@
 import agtermCore
 import AppKit
+import SwiftUI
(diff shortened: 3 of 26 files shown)


hackmajoris:
Could you also add the standard reopening tab/session shortcut: `cmd+shift+t`.?

umputun (commented):
approach is fine, keep both the grace-undo and the persistent Open Recent, and on-by-default is good. A few things to sort out first.

**bugs**

1. `restoreRecentClosed` rebuilds the session/workspace from the snapshot with the *same* UUIDs and inserts without checking they aren't already live. If the post-reopen `recentClosedStore.remove(id)` fails (it swallows write errors) or you crash between finalize and removal, Open Recent can reinsert an item that's already in the tree, giving you duplicate Identifiable ids that corrupt selection/lookup and SwiftUI diffing. Guard it: skip the entry (or select the existing one) if any of its ids already exist in the target store.

2. `softCloseSession` removes from the tree and `cancelPendingSave()` but doesn't save, so a crash during the 3s grace recovers the session from disk, which is the right direction. But if an unrelated `save()` runs in that window (open a new session -> `addSession` -> `save()`), the tree is flushed without the soft-closed session, which then only lives in memory and isn't in recent-closed yet. A crash there loses it outright. Recording to recent-closed at soft-close time and removing it on undo closes the gap and unifies the two close paths.

3. multi-close only keeps the latest live-undoable. Close A then B within 3s and ⌘Z/toast only brings back B. A finalizes to Open Recent on its own timer, so its live process and scrollback are gone. Not a leak, teardown is fine, but the undo quietly degrades for rapid closes. Either promote the next pending record into `pendingCloseSummary` after undo/finalize, or accept it, but add a test either way, it's untested right now.

**shortcut**

`openLatestRecentClosed` / `reopenLatestRecentClosed` are wired all the way through but nothing calls them. Bind that to ⌘⇧T for persistent reopen and keep ⌘Z for the in-window grace undo. That's the standard reopen-closed-tab gesture and covers @hackmajoris's ask.

**tests**

- no `RecentClosedTests.swift`. The dedup-by-snapshot-id, the 20-item cap, the version gate, and the disk round-trip in the store aren't tested directly (also the one-test-file-per-source convention).
- every soft-close test uses `grace: 60`, so the scheduled finalize task never actually fires. Add one that lets the timer run and asserts the undo-then-timer-fires no-op.

**minor**

- recent-workspace restore `append`s while pending-workspace restore inserts at the original index. Probably intentional, but drop a comment, it reads as a bug.
- `PendingCloseSummary.expiresAt` and `cancelPendingClose(forSession:)` look dead, set-never-read and unreachable.
- the toast has a `.transition` but no `withAnimation` around `pendingCloseSummary`, so it pops in and out instead of animating.
- `undo_close` is a new rebindable keymap action but it's not in the README keybindings or the agent-skill action list. One line each, non-blocking.

umputun (approved):
round 2, all addressed - verified each:

- dedup guard (selects the existing / undoes the pending close instead of reinserting) + record-at-soft-close with remove-on-undo + multi-close promotion, all with tests
- ⌘⇧T reopen-recent and ⌘Z grace-undo (skips text fields so native undo survives), covers @hackmajoris's ask
- `RecentClosedTests` for dedup/cap/version-gate/round-trip, plus the grace-timer-fires and promotion tests
- toast animates now, dead code gone, README + agent-skill updated

reopen/undo stay GUI-only (the grace window isn't scriptable), so no control-API coverage by design - fine.

lgtm

PR #178, opened 2026-07-08
fix: clamp restored window frames to visible screens

melonamin:
## Summary
- Clamp saved window frames before applying them during relaunch.
- Move the pure restore-geometry decisions into `WindowGeometry` so they are host-free and unit-tested.
- Keep `WindowAccessor` responsible only for converting `NSWindow`/`NSScreen` values into core geometry types.

## Bug
`agterm` saves each window frame and reapplies it on launch. If the last saved frame was on an external display that is no longer connected, the app restored that exact frame with only a width/height check. The window could reopen entirely off-screen, leaving the user with an apparently missing window.

## Fix
The restore path now normalizes the saved frame first. Core geometry code picks the current display with the largest overlap, reports no-overlap as a disconnected-display case, clamps oversized frames to the display, and fully contains the restored frame on the chosen display. The full-containment step avoids the vertical edge case where a frame saved on a monitor above the current display would restore with only its bottom strip visible and the titlebar still off-screen.

## Validation
- `swift test --package-path agtermCore --filter WindowGeometryTests`
- `./scripts/test.sh`
- `make build`
- `swiftlint lint --strict --quiet agterm/Views/WindowAccessor.swift agtermCore/Sources/agtermCore/WindowGeometry.swift agtermCore/Tests/agtermCoreTests/WindowGeometryTests.swift`

### DIFF
--- a/agterm/Views/WindowAccessor.swift
+++ b/agterm/Views/WindowAccessor.swift
@@ -231,7 +231,50 @@ struct WindowAccessor: NSViewRepresentable {
             guard let saved = UserDefaults.standard.string(forKey: Self.frameKey(windowID)) else { return }
             let frame = NSRectFromString(saved)
             guard frame.width > 0, frame.height > 0 else { return }
-            window.setFrame(frame, display: true)
+            window.setFrame(Self.constrainedRestoredFrame(frame, for: window), display: true)
+        }
+
+        /// Normalize a persisted frame before restore. A saved `NSRect` may come from a display that is
+        /// no longer attached; applying it directly can put the relaunched window entirely off-screen.
+        /// Keep the saved size/position when possible, but clamp it to a visible strip on a current screen.
+        private static func constrainedRestoredFrame(_ frame: NSRect, for window: NSWindow) -> NSRect {
+            guard let screen = bestScreen(for: frame, fallback: window.screen ?? NSScreen.main) else { return frame }
+            let maxSize = screen.visibleFrame.size
+            let size = WindowGeometry.clampSize(
+                WindowGeometry.Size(width: Double(frame.width), height: Double(frame.height)),
+                min: WindowGeometry.Size(width: Double(window.minSize.width), height: Double(window.minSize.height)),
+                max: WindowGeometry.Size(width: Double(maxSize.width), height: Double(maxSize.height))
+            )
+            let origin = WindowGeometry.clampOrigin(
+                WindowGeometry.Point(x: Double(frame.origin.x), y: Double(frame.origin.y)),
+                windowSize: size,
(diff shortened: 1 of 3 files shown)


umputun (commented):
lgtm on the fix. It's conservative too, a fully on-screen window clamps to an identical frame so only the gone-display / partially-off-screen cases change.

one thing: `bestScreen`, `intersectionArea`, and the size+origin composition in `constrainedRestoredFrame` are pure geometry but live app-side with no tests. That's the same logic `WindowGeometry` already owns host-free (and unit-tests). Worth pushing the pure part down there, e.g. a `bestDisplayIndex(for:among:)` over `[Rect]` plus a `constrain(frame:min:displayFrame:)`, leaving only the `NSScreen` / `NSWindow` -> `Rect` conversion here. Then multi-screen selection, the no-overlap fallback, oversize-shrink, and off-screen-origin all get real coverage in `WindowGeometryTests`.

not a blocker, but restore-geometry edge cases are worth having tests on.

melonamin:
updated

umputun (commented):
geometry hoist + tests are in, all green locally (swift test + swiftlint).

round 1 missed one edge case: the disconnected-display fallback can still leave the titlebar off-screen when the saved frame came from a monitor arranged *above* the current one (e.g. external mounted above a laptop, then undocked). `bestDisplayIndex` returns nil, so `constrain` clamps the original y to `maxY - visibleMargin` and only the window's bottom 80pt strip stays on-screen, with the titlebar above the top edge, not grabbable. The horizontal case is fine since the strip is part of the full-width titlebar, so it's vertical-only. Might be masked by AppKit's own `constrainFrameRect` on setFrame, I haven't confirmed that on-device. Not a blocker, the primary off-screen bug is fixed either way, but for an automatic restore fully containing the frame (or a titlebar-reachable clamp) would be safer than reusing the `window.move` partial-strip behavior.

minor, tests: the two `constrain` cases each exercise only one clamp. `constrainShrinksOversizedFrameToDisplay` has an in-range origin and `constrainKeepsOffScreenOriginAtLeastPartiallyVisible` has an in-range size, so nothing pins that the origin clamps against the *shrunk* size. A combined case covers it: origin (5000,100) size 5000x4000 on the 1920x1080 display, expect origin (1840,100) size (1920,1080).

minor: `WindowGeometry.Size.init(_ cg:)` and `Rect.init(_ cg:)` are now duplicated file-private in WindowAccessor and WindowRegistry. Could share one conversions file, but that's a maintainer cleanup, not on you.

umputun (approved):
round 3 addressed both, full-containment on restore (strip clamp kept for move/resize) plus the combined + above-display tests, lgtm, thx

PR #179, opened 2026-07-08
feat: support multi-session batch actions

melonamin:
## Summary

Adds multi-select support to the sidebar so batch commands and drag/drop can operate on multiple sessions at once.

## What changed

- Adds host-free sidebar selection state in `agtermCore` for ordered multi-selection, pruning, and context-menu target resolution.
- Enables multi-row sidebar selection with Shift/cmd-click while keeping the active terminal selection separate.
- Preserves multi-selection on right-click when the clicked row is already selected; right-clicking another session narrows the command to that row.
- Applies batch actions for selected sessions: close, move, flag/unflag, and clear status.
- Enables dragging selected session groups between workspaces or within the same workspace, preserving tree order and precise drop placement.
- Keeps batch drag/drop index math host-free in `SidebarDrop`.
- Groups multi-session grace-period closes into one undoable pending close, with undo/reopen restoring every selected session.
- Adds focused tests for selection normalization, hidden-row pruning, batch move/flag, grouped close undo/finalize, recent-closed restore, and batch drag/drop index math.

## Validation

- `swift test --package-path agtermCore --filter "SidebarDropTests|SidebarSelectionTests"`
- `swift test --package-path agtermCore`
- `xcodebuild -project agterm.xcodeproj -scheme agterm -configuration Debug -destination "platform=macOS" build CODE_SIGNING_ALLOWED=NO`
- `swiftlint lint --quiet` on touched Swift files
- `git diff --check`
- `codex-review --full-access --parallel-tests "swift test --package-path agtermCore --filter SidebarDropTests|SidebarSelectionTests"`: clean, no accepted/actionable findings

## Screenshots

Screenshot 2026-07-08 at 19 40 27


umputun (changes requested):
few things before this can go in.

**control API coverage is missing.** the keep-in-sync rule says a new user action isn't done until it's drivable from the socket, and this adds five batch actions while touching no `ControlProtocol`/`ControlServer`/`agtermctl` file.

not all five are equal though:

- clear-status adds no store method, `menuClearStatus` just loops the existing `setAgentIndicator`. A shell loop over `agtermctl session status idle --target $s` gives a byte-identical model, so it's exempt on the merits.
- flag/unflag is a pure plural, `setFlag(_:forSessions:)` is the singular in a loop with one `save()`. Also loop-equivalent.
- `session.move` loses index atomicity. A script looping `--after` has to recompute anchors after every call, and there's no way to express "remove all N first, then insert the block at the post-removal slot", which is exactly what `SidebarDrop.resolveSessions` computes.
- close is the real gap. One grace timer, one undo restoring N sessions, one summary toast, none of that is reachable by looping.

so pls add multi-target `session.close` and `session.move` (command case, dispatcher arm, `agtermctl` subcommand, round-trip + e2e tests), and call out flag/clear-status explicitly as loop-equivalent exemptions. The rule requires an exemption to be stated and right now nothing is.

one thing worth knowing before you do the close one: `ControlServer+SessionActions.swift:174` calls `store.closeSession(id)`, the hard close. Control has never reached the grace/undo path at all, so multi-target close means first deciding whether it should.

**`clearFlags()` doesn't prune the selection.** AppStore.swift:715-724. This PR is what introduces the invariant that a flag mutation prunes `sidebarSelection`. You added the prune to `setFlag(_:forSession:)`, which had none before, and to the new `setFlag(_:forSessions:)`. `clearFlags()` is the one flag path left without it, and it's reachable from the menu bar, the palette, and `ControlServer+SessionActions.swift:456`.

repro: flagged mode, select A and B, Clear Flagged, switch back to tree. A and B are still multi-selected. Unflagging one session drops it from the selection, unflagging all of them doesn't. Not dangerous, `syncSelection:618` re-applies the selection so the rows come back visibly highlighted, just inconsistent. `pruneSidebarSelection()` before the `save()`, plus a test.

**`SidebarSelectionController.targets(forContextSession:fallbackSessionID:)` is dead code.** SidebarSelection.swift:40. Zero production callers, the only references are the three assertions at SidebarSelectionTests.swift:15-17. The live context-menu path calls `AppStore.sidebarSelectionTargets(forContextSession:)` (AppStore.swift:891), which reimplements the same rule with a visibility filter and no `fallbackSessionID`. Two copies of one rule, and the tested one is the unused one. Drop whichever you don't want.

**the flag context menu left the AppActions seam.** WorkspaceSidebar+ContextMenu.swift:172. It used to call `actions.toggleFlag(node.id)`, now it calls `store.setFlag(...)` directly. The direct call is the more correct of the two, `AppActions.store` is `library.activeStore` so the old path looked the session up in the frontmost window's store and silently no-oped on a background window's sidebar. But this skips the seam rather than fixing it. `AppActions.toggleFlags(_:in:)` mirroring the `closeSessions(_:in:)` you already added would fix this and the View-menu path together.

**two questions.**

`allowsEmptySelection = false` (WorkspaceSidebar.swift:56) is new here, why? `syncSelection` calls `outline.deselectAll(nil)` at :599 when the active session has no visible row, and AppKit documents `deselectAll` as a no-op when empty selection is disallowed. That would leave a stale highlight where `.claude/rules/sidebar.md:169-174` says the sidebar should show no selection at all (focus a workspace that doesn't hold the active session, the rule calls that state "intentional, not a bug"). I haven't run it so I may be wrong about the AppKit side, but those two lines look in tension.

grouped undo vs Reopen Closed: `pendingCloseID(containingSessionID:)` now matches the whole `.sessions` record (AppStore+RecentClosed.swift:79), so batch-closing A, B, C and then picking only B from Reopen Closed inside the grace window restores all three and clears all three from the list. It matches how `.workspace` already behaves and it's fine once the record finalizes. Intended? If so a one-line comment there would save the next reader.

**docs.** `.claude/rules/sidebar.md` is stale in four spots now: line 47 still describes the single `-1` post-removal adjustment for session drags, lines 49-54 list only `resolveSession`/`resolveWorkspace` as the entry points, line 105 omits `setFlag(_:forSessions:)`/`moveSessions`/`softCloseSessions`, and line 109 still routes the row context menu through `AppActions.toggleFlag`. There's also no selection model described anywhere in it. README.md and `site/docs.html` want a sentence on shift/cmd-click too.

no XCUITests for any of the new surface, and README:125 claims close/move/drag are covered by them. Shift-click, cmd-click, right-click narrowing and multi-row drag are all AppKit-level, the host-free tests structurally can't reach them.

**optional, your call.** `PendingCloseRecord.session` and `.sessions` could collapse into one array-carrying case, one-element array for the single close. That kills five duplicate switch arms (`undoPendingClose`, `finalizePendingClose`, `summary(for:)`, both `pendingCloseID` lookups) and the `selectRestored:` bool, and it matches the shape `PendingWorkspaceClose` already has. Not a blocker, current code is correct and tested.

umputun (changes requested):
round 1 is all addressed: batch close/move are on the socket with the exemption stated at `.claude/rules/control-api.md:265-267`, `clearFlags()` prunes, the dead `targets(forContextSession:fallbackSessionID:)` is gone, the row context menu is back on the `AppActions` seam with the store threaded through, and `allowsEmptySelection` is `true` now so my `deselectAll` worry was wrong. You took the `PendingCloseRecord` collapse too, and it holds up for the single-session path. `swift test` 1343/1343, swiftlint clean, Debug and Release both build.

three things block.

**1. batch requests put `"active"` in the wire target.** `SessionCommands.swift:72` and `:158`. `singleTarget` is nil once `targets.count > 1`, so the real targets ride only in `args.targets` and the top-level target falls back to `"active"`. `ControlRequest` has no version field and synthesized `Decodable` drops unknown keys, so a server built before this PR never sees `args.targets` and runs its old arm against `request.target`. `agtermctl session close --target A --target B` then hard-closes the *active* session and prints `ok`. Same shape for `session.move`.

that skew is reachable. `/usr/local/bin/agtermctl` is an `ln -sf` into the bundle (`CLIInstall.privilegedInstallCommand`), so replacing the bundle repoints the CLI at the new binary while the already-running process keeps serving the socket with old code. `make deploy` and `brew upgrade --cask agterm` both produce exactly that state, and the running app keeps the old code until someone quits it.

fix is `target: target.targets.first ?? "active"` in both spots. It is inert on a current server, since every batch branch in `ControlDispatcher.swift:197-255` reads `args.targets` and never touches `request.target`. On an old one it degrades to acting on a session the user actually named rather than one they didn't.

**2. `agtermctl session close --target A --target B` prints `2 diagnostic(s)`.** `SocketClient.formatResponse` maps any non-nil `result.count` to the keymap-reload string (`SocketClient.swift:145-147`), and both batch arms return `ControlResult(count: ids.count)` (`ControlServer+SessionActions.swift:193`, `:538`, `:590`). Only `text` and `exitCode` sit ahead of it. I ran `formatResponse` against `ControlResult(count: 2)` to confirm. Nothing catches this: the e2e sends raw JSON over the socket and `SocketClientTests.swift` is untouched.

pls give the batch commands their own result field instead of reusing `count`, and have it report sessions actually affected rather than targets requested. Today `session move --target A --target B W` with `A` already in `W` answers `2` while only `B` moved.

**3. the test guarding the `clearFlags()` prune doesn't guard it.** `clearFlagsPrunesRowsHiddenInFlaggedMode` (`SidebarSelectionTests.swift:107`) asserts `sidebarSelectionIDs == []` right after `clearFlags()` in flagged mode. Nothing is visible at that point, so the read-time filter inside `sidebarSelectionIDs` returns `[]` whether or not the raw list was ever pruned. I deleted `pruneSidebarSelection()` from `clearFlags()` and the full suite still passed, 1345/1345. Its sibling `batchFlagChangePrunesRowsHiddenInFlaggedMode` (`:91`) is real because it re-flags and asserts a second time. `clearFlags` needs that same step.

two more, not blocking but I'd like them in this PR.

**control batch close ignores `closeGraceUndoEnabled`.** `ControlServer+SessionActions.swift:190` calls `softCloseSessions` unconditionally while the GUI branches on the setting (`AppActions.swift:242-253`), and `ControlServer` already holds `settingsModel` at `ControlServer.swift:23`. A user who turned grace-undo off still gets a 3s grace and a toast from a script. It also means `ok` from the batch form doesn't say the surfaces are gone, while `ok` from the single-target form does. The comment at `:188-189` ("that is the GUI behavior this API makes scriptable") is wrong as written, since the GUI behavior is conditional.

**grouped undo focuses the wrong session.** Batch close writes one recent entry per session (`AppStore+PendingClose.swift:124`) and Open Recent renders a button for each (`agtermApp+Menus.swift:121-128`), but `pendingCloseID(containingSessionID:)` matches the whole record. Inside the 3s grace, clicking `beta` restores alpha+beta+gamma, clears all three rows, and focuses `close.selectedSessionID` instead of `beta`. Nothing is lost, it over-restores. Keep the grouped restore, just select the session that was picked.

smaller: `moveSessions` and `moveSession` disagree on a one-element list. `AppStore.swift:454` filters out members already in the target workspace, `:438` removes and re-appends. `ControlDispatcher.swift:248-252` routes any non-nil `targets` to `moveSessions`, so a raw client sending `{"targets":["a"],"workspace":"W"}` with `a` already in `W` gets a no-op where `{"target":"a"}` would move it to the end. The CLI can't reach it today because `singleTarget` collapses one target onto the top level, but the wire contract shouldn't fork on array length.

optional, your call. `SidebarSelectionController` is 41 lines wrapping one `[UUID]`, and since `sidebarSelectionIDs` re-normalizes on read, both the stored order and the pruning are redundant. No test drives the type directly, they all go through `AppStore`. A `private var sidebarSelectionRaw: [UUID]` plus the existing computed accessor would be the same thing. Also `AppStore.swift` is at 979 lines against the 1000 cap, worth knowing, not yours to fix here.

the drag/drop index math I checked closely and it's right, including dropping a selection onto a row inside itself. Same for the grace-timer races: undo after finalize, overlapping batches, and closing a session already inside a pending batch all no-op the way they should. `resolveBatchSessions` failing all-or-nothing before it mutates anything is the right call.

umputun (commented):
round 2 is addressed:

- wire target carries `targets.first` on both `session.close` and `session.move` (`SessionCommands.swift:72`/`:158`), covered by `sessionCloseMultipleTargets` and the move round-trips.
- `result.affected` is its own field and renders as `1 session`/`N sessions`, covered at 0, 1 and N by `formatResponseAffectedSessions` (`SocketClientTests.swift:186`). `session move --target A --target B W` with `A` already in `W` reports 1.
- `clearFlags()` prune is now really guarded by `SidebarSelectionTests.swift:120`.
- control batch close branches on `closeGraceUndoEnabled` (`ControlServer+SessionActions.swift:189`), grouped undo focuses the picked session (`AppStore+PendingClose.swift:281`), one-element `moveSessions` matches `moveSession` (`AppStore.swift:457` plus the dispatcher route at `ControlDispatcher.swift:281`).

`swift test` 1349/1349, swiftlint clean.

**one blocker: the prune fix is tested in 2 of 5 places.**

these three call sites can be deleted with the full 1349-test suite still green:

- `AppStore.swift:618`, `setSidebarMode`
- `AppStore.swift:627`, `setFocusedWorkspace`
- `AppStore.swift:680`, `setFlag(_:forSession:)` (the singular)

only `:696` (plural `setFlag`) and `:733` (`clearFlags`) are guarded. It is the same shape you already fixed once: `sidebarSelectionIDs` filters on read, so asserting `== []` while the rows are hidden passes whether or not the raw array was pruned. `batchFlagChangePrunesRowsHiddenInFlaggedMode` is real precisely because it re-flags and asserts a second time. The three above need that second step.

the singular `setFlag` matters most of the three, it is what `session.flag` and the menu `toggleFlag` call.

---

two non-blocking, would be nice to have both in this PR.

**the flag exemption is stated wrong.** `.claude/rules/control-api.md` says Flag/Unflag is "loop-equivalent to repeated `session.flag on|off|toggle --target <id>`". `AppActions.toggleFlags` (`AppActions.swift:474`) computes `allSatisfy(\.flagged)` and applies one uniform value to the set. A loop of `session.flag toggle` flips each row independently, so a mixed selection diverges: GUI flags both, the loop unflags the flagged one. `on`/`off` are loop-equivalent, `toggle` is not. The exemption still stands, the sentence doesn't.

**`resolveBatchSessions`' all-or-nothing has no test.** `session.close --target <valid> --target <bogus>` errors without closing the valid one (`ControlServer+SessionActions.swift:628`), which is the right call and the thing most worth pinning. Same for an ambiguous prefix mid-batch.

---

smaller, your call on all of these.

`closeSessions` and `placeSessions` disagree on the empty result: `:200` errors `"no sessions closed"`, `:602` returns `ok` with `affected: 0`. The close branch is also dead, `softCloseSessions` cannot return false once every id resolved. Pick one, I'd keep the `placeSessions` shape.

`session close --target a --target a` dedupes to one id and takes the hard-close path, so a duplicated target silently loses the grace window that `--target a` alone also loses. Consistent with the documented single-target hard close, just a sharp edge.

`SidebarSelectionTests.swift` has no matching source file now that `SidebarSelection.swift` is gone (that collapse was my suggestion, so this is my mess). Everything it drives lives in `AppStore`. Folding it into `AppStoreTests.swift` would match the one-test-file-per-source convention.

`AppStore.swift` is at 983 lines against the 1000 cap. Not yours to fix here, just so you know before the next thing goes into it.

`AGTERM_UITEST_ALLOW_CLOSE_CONFIRMATION` (`ContentView.swift:56`) is production code that only a UI test reads. It follows the `isUITestLaunch` precedent two lines up and it only narrows the bypass, never widens it, so I'll take it. Flagging it so it's a deliberate yes rather than something that slid through.

---

two things that are not yours.

reopening a batch after the grace window expires restores one session per click, because `softCloseSessions` writes one recent entry per member (`AppStore+PendingClose.swift:123`). A workspace close writes one entry that restores the whole workspace forever. The comment at `AppStore+RecentClosed.swift:77` scopes the grouping to "while the record is still pending", which is accurate and is what I asked for in round 1. Three multi-selected sessions aren't a durable container the way a workspace is, so N entries is the right model. No change wanted, noting it because "matching workspace close behavior" only holds until finalize.

undoing a pending session close and then a pending workspace close for the same workspace produces two workspaces with the same UUID. `restorePendingSession` recreates the missing workspace (`AppStore+PendingClose.swift:266`), then `restorePendingWorkspace` inserts it again unconditionally (`:298`). This reproduces on the single-session `softCloseSession` path too, and neither function is touched here, so it predates this PR. Your grouped record widens the door, any member's Recent entry now reaches the group undo, but doesn't open it. I'll file it separately, don't fix it here.

the drag/drop index math holds up: same-workspace up and down, non-contiguous selection, dropping onto a row inside the dragged block, cross-workspace, and mixed sources already in the target. The grace-timer races are fine too, and `resolveBatchSessions` failing before it mutates anything is the right shape.

melonamin:
All addressed per request

umputun (approved):
round 3 is addressed. The prune-guard blocker is genuinely fixed: all five prune tests now hide-then-re-show and assert a second time, so deleting `pruneSidebarSelection()` from any of the five sites (including the singular `setFlag`, which is the one `session.flag`/`toggleFlag` hit) fails the second assert. The two non-blocking items are in too, the `control-api.md` flag exemption now correctly says on/off is loop-equivalent while `toggle` isn't, and `resolveBatchSessions` all-or-nothing is pinned by `testSessionCloseBatchIsAllOrNothing`. Tests/lint/Debug build green, and I found no new correctness, wire-contract, or host-free-boundary issues.

one optional nit: `AppStore.swift:908` `replaceSidebarSelection(with:)` is `public`, but every caller is inside `agtermCore` (AppStore plus the two extensions) and its sibling `pruneSidebarSelection` is already `internal`. Drop the `public`.

not for this PR: `AppStore.swift` is at 994/1000 and `AppStoreTests.swift` at 1894/2000. The sidebar-selection block would fit an `AppStore+SidebarSelection.swift` whenever the next thing goes into either.

PR #180, opened 2026-07-09
feat: Improve sidebar drag and reveal actions

melonamin:
## Summary

- Add Finder folder drops on the sidebar to create sessions rooted at the dropped directories.
- Add `Reveal in Finder` for the active session and session context menu.
- Seed directory-picking panels from the active session cwd when available.
- Export public text for sidebar drags while keeping reorder semantics on private UUID pasteboard types.
- Spring-open collapsed workspaces while dragging over them.

## Notes

The AppKit work stays in the app target. `agtermCore` is unchanged; sidebar reorder math still goes through the host-free `SidebarDrop` resolvers.

## Validation

- `xcodebuild -project agterm.xcodeproj -scheme agterm -configuration Debug -destination "platform=macOS" build`
- `make lint`
- `git diff --check`
- `cd agtermCore && swift test --filter SidebarDropTests`

Full `cd agtermCore && swift test` currently reports `AppStoreTests.scheduledFinalizeTearsDownAfterGrace` failing in the timer/grace-period path. This PR does not touch `agtermCore`, and the focused sidebar host-free tests pass.


umputun (commented):
i have concerns with this one, and they start above the code.

**the premise.** i'm not sold on Finder folder drop. Getting to a directory from a terminal is a solved problem, there are endless CLI tools built for exactly that, and we already have the open-directory-for-new-session dialog, which on macOS should already accept a folder dropped onto it. So the gain over what's there looks small to me, and it isn't behavior i've wanted from a terminal.

**already doable.** "Reveal in Finder" is a one-liner as a custom command: `open -R "$AGT_SESSION_PWD"`. `keymap.conf` is the extension surface, and when something falls out of it in one line, that's where it lives, not in `AppActions`. The sidebar context-menu variant (acts on the clicked row, not the active session) is the only part config can't express.

**scope.** five unrelated changes in one commit: the folder drop, reveal, the `NSOpenPanel` cwd defaulting, the public drag-text export, and spring-loading. The drag-text export isn't in the title or the description at all, and it's the riskiest hunk here.

reveal is pure UI, so i won't ask for the full control-API surface on it, but that skip should be called out rather than left silent.

**bugs found:**

1. `DirectoryPanelDefaults.existingDirectoryURL` never returns nil for a non-empty path. It walks up to `/` and returns that, so `url(paths:)` always takes the first candidate and both the second candidate and the home fallback are dead code. `url(paths: "/Volumes/gone/proj", "/Users/me")` gives `/Volumes`.
2. a spring-opened workspace never re-collapses. `outlineViewItemDidExpand` inserts into `expandedWorkspaceIDs` unconditionally, and `suppressExpansionPersist` only gates the store write. Abort a drag and it stays open for the rest of the run.
3. folder drop in flagged mode: the drag source is gated to `.tree`, the drop target isn't. `workspaceNode(forID:)` returns nil, the drop still validates `.copy`, and `addSession` selects an unflagged session that mode can't show.
4. `plainTextPasteboardType` ("public.utf8-plain-text") is the same UTI as `.string`, so the item sets it twice. The comment calling `.string` a legacy type is wrong.
5. a session row dragged into a terminal pane now pastes its cwd, since the item writes `.string` and the surface registers `.string`. That makes the comments at `GhosttySurfaceView+Input.swift:10-11` and `:26-27` false.
6. external drags went from forbidden (`[]`) to `.copy` and now carry `focusedCwd` as text. A drag's pasteboard is readable by the destination before you release, so dragging a row over another app exposes an absolute path.
7. a multi-folder drop spawns one shell per directory with no cap, and `autoUnfocusIfOutsideFocus` can silently drop you out of focus mode.

no tests added. `DirectoryPanelDefaults` is pure Foundation, so it belongs in `agtermCore` where `swift test` reaches it, and four assertions there would have caught the first bug.

on the note in your description: `scheduledFinalizeTearsDownAfterGrace` is a real load-dependent flake on master, not something you introduced. I reproduced it in 1 of 4 full-suite runs under CPU load, while it passes 10/10 in isolation. I'll fix it separately.

i'd rather not take this as-is. The one piece worth keeping on its own is seeding `NSOpenPanel.directoryURL` from the active session cwd, so the folder pickers open near where you're working instead of wherever AppKit last left them. That one needs `DirectoryPanelDefaults` moved to `agtermCore` with tests. For the rest, let's settle the premise first.

melonamin:
> scope. five unrelated changes in one commit:

Should I open PR per feature? All of theses are kinda-sorta around sidebar, that's why I bundled them

> the premise. i'm not sold on Finder folder drop
> already doable. "Reveal in Finder" is a one-liner as a custom command

Agree with already doable, but that is not the point. As a long time Mac user I expect that most of the things are dragable. If I can drag a folder into terminal a get the path pasted, why can't I drag it into session list and get a new session with that folder as cwd? 

Reveal in finder is the same, of course I can make a custom shortcut or just type `open .` in terminal, in my mind this doesn't mean I shouldn't have the same affordance for mouse. 

Implementation quirks aside I think this makes the app better and more native.

umputun (commented):
you changed my mind on the premise, both halves of it.

the terminal pane already registers `.fileURL` (`GhosttySurfaceView.swift:210`) and `GhosttyCallbacks.pasteboardText` shell-escapes the dropped path. So "if I can drag a folder into the terminal" isn't an analogy, it's shipped behavior, and the sidebar is the missing half of it. There's also no Finder-to-session route anywhere in the app today: no `application(_:open:)`, no `CFBundleDocumentTypes`.

on reveal, my `open -R "$AGT_SESSION_PWD"` was just wrong. `CustomCommandRunner.swift:230` binds `sessionPWD` to `session.effectiveCwd`, and that's `currentCwd ?? initialCwd` (`Session.swift:291`), the primary pane. You use `focusedCwd` (`Session.swift:264-267`), which honors `splitFocused`. In a split with the right pane focused my one-liner reveals the wrong directory, and there's no `AGT_` token for the focused cwd at all. On top of that the app already reveals in Finder for `file://` link clicks (`GhosttySurfaceView+Input.swift:441`), so "that lives in keymap.conf, not `AppActions`" contradicts what we already ship.

both features stay. Sorry for the runaround.

**scope.** Keep it as one PR and fix in place. The hunks are separable, but not so entangled that splitting buys us anything at this size, and you'd pay for it in rebases.

**four things I got wrong.**

1. the flake. I said `scheduledFinalizeTearsDownAfterGrace` is a real flake on master and I'd fix it separately. It was already fixed, in 7c0d95a, which swapped the flat 30ms sleep for a bounded poll on `surface.teardownCount`. Your branch is 3 commits behind and still has the old version. Rebase and it's gone, nothing for me to do.

2. bug 1's wording. "the second candidate and the home fallback are dead code" is wrong. They run whenever an earlier candidate is nil or empty, which is the default (`customDirectory` is nil until set, `SettingsView.swift:175-178`). The bug is narrower: `existingDirectoryURL` never returns nil for a non-empty path, so a set-but-stale `newSessionCustomDirectory` skips the fallback instead of falling through to the session cwd. I ran the function: `/Volumes/gone/proj` gives `/Volumes`, `/nope/nope/nope` gives `/`. Return nil once you've walked past the last real component and let the next candidate have a turn.

3. bug 5 is deliberate and I wrote it up as an accident. Your comment at `WorkspaceSidebar.swift:74-75` says terminals may take `.copy` from the public text representation. What's actually broken is that `GhosttySurfaceView+Input.swift:10-11` and `:26-27` still claim a sidebar row drag is ignored, and they're false now.

4. bug 6 I overstated. The cwd ends up on the pasteboard because you write text to it, not because the external mask went from `[]` to `.copy`. The mask only decides whether a drop can complete. And "any app you drag over can read it before you release" is inference on my part, I didn't verify cross-process `draggingEntered` delivery. The version I'll stand behind: external drags now export the session cwd, and workspace rows export their name, as plain text, with no opt-out and no mention in the description.

**pls drop the drag-text export.** It isn't load-bearing for anything else here. The folder drop is a drop destination, the text export is a drag source, they share no code. Reverting the local mask to `.move` and not writing text kills the terminal-paste behavior, the pasteboard question, and the duplicate-UTI bug in one cut, and Finder folder drops keep working because Finder is an external source with its own mask. If you want row drags to mean something later, `public.file-url` for the cwd is the honest payload: Finder would accept it, the terminal would get an escaped path. Raw text is the one representation that makes a workspace row a lie, since it exports the workspace *name*, which refers to nothing.

while you're in there: `.string` **is** `public.utf8-plain-text`, so `plainTextPasteboardType` sets the same UTI twice on the same item, and the comment calling `.string` legacy is wrong. `public.text` is the only genuinely distinct addition.

**reveal owes a control command, and I let you off too easy.** I wrote "reveal is pure UI, so i won't ask for the full control-API surface." CLAUDE.md's exemption for click-routing `reveal` names `AppActions.reveal(windowID:sessionID:pane:)` (`:817`), the notification routing action, not a Finder reveal. `agtermctl session reveal --target <id>` is something a script would plausibly want, so it needs the four points: `Command` case, dispatcher arm, `agtermctl` subcommand, round-trip and e2e tests.

**still standing from round 1:** bugs 2, 3, 4 and 7.

on 2, worth narrowing: `sidebar.md` already says programmatic expansion doesn't persist, and it doesn't here either. The real complaint is that Finder's spring-loading re-collapses when you abort the drag and this doesn't, because `cancelSpringLoadedExpansion` only cancels a pending work item. Track whether the row was opened by spring-load and collapse it on exit or cancel.

on 7, it's worse than I said. `autoUnfocusIfOutsideFocus` fires on a **single** folder too: drop into empty sidebar space, `targetWorkspaceID(for: nil)` resolves `currentWorkspaceID`, and if the active session is stranded outside the focused workspace the drop silently clears focus.

few more:

- `DirectoryPanelDefaults` is pure Foundation sitting in the app target, so `swift test` can't reach it. That's exactly why the `/Volumes` behavior shipped untested. Move it to `agtermCore` and give it the four assertions.
- no docs. README, `site/docs.html` and `agterm/Resources/agent-skill/` are all untouched, and this adds two user-visible features.
- `directoryURLs(from:)` calls `fileExists` per URL and `validateDrop` runs on every mouse move, so a multi-select from a network volume will hitch the drag. Resolve once per dragging sequence.
- `revealSessionInFinder` returns silently when the cwd isn't a directory (`AppActions.swift:145-148`), but the menu item is only disabled on `activeSession == nil`. Enabled item, nothing happens.

umputun (commented):
the harden commit + rebase clears everything from last round. All the round-2 items check out against the current code (text export dropped, the `session.reveal` four points, `existingDirectoryURL` nil-return, re-collapse on abort, flagged-mode drop gating, no focus-clear, `DirectoryPanelDefaults` in agtermCore with its four assertions, docs across all five surfaces, the per-drag cache, the enabled-vs-noop menu guards). Two new things from this pass.

**the abort-collapse fix over-corrected: a cross-workspace session move re-hides its own target.** `acceptDrop`'s `defer { finishDraggingSequence() }` (`WorkspaceSidebar+DragDrop.swift:63`) runs `cancelSpringLoadedExpansion()` on success too, not just on cancel. A Finder folder drop self-heals, `addSession` sets a fresh `selectedSessionID` so `syncSelection` re-expands the workspace. But `moveSessions` never touches the selection, so `syncSelection`'s reveal (gated on `selectionChanged`, `WorkspaceSidebar.swift:645`) never fires: drag a session onto a collapsed workspace, it spring-opens, you drop, and it re-collapses with the moved session hidden until you expand it by hand. The `draggingExited` / cancelled `draggingSession:endedAt:` paths are the ones that should keep collapsing. Fix: on a successful accept, clear the cache and pending work item but leave an already-spring-opened workspace open. Keep the collapse only on the exit/cancel paths.

**`.claude/rules/control-api.md:231` still says 58.** The catalog header (204) and mirror note (1160) both went to 59 with `session.reveal`, but the debug.appearance line still reads "the public command count stays 58", should be 59.

non-blocking: `AppActions.swift` is at 992 lines now, 8 under the strict limit, not yours to fix here, just a note before the next edit trips it. `README.md:159` "capped at 20 folders" reads like it truncates to 20, but over 20 rejects the whole drop (docs.html has it right). And the >20-rejection boundary plus the spring-load/drop happy path have no coverage, the AppKit drag is impractical to drive in XCUITest, so fair to skip.

umputun (commented):
lgtm. All three round-3 items are addressed: the cross-workspace session move keeps the spring-opened workspace open now, abort/cancel/leave still collapse (traced every path), and the count fix + README wording are both good. Build, `swift test`, and lint green.

PR #189, opened 2026-07-11
fix(sidebar): re-tint row text from the row view's live selection state

melonamin:
## Bug

Sidebar row text can render invisible — same foreground color as the pill/background behind it. Shows up as an unreadable selected row (white text on the white selection pill) or a blank-looking unselected row.

Reported with `Ghostty Default Style Dark` (video below), but it's not specific to that theme.

## Root cause

The sidebar draws its own selection, split across two mechanisms that can disagree:

- The **selection pill** (`SidebarRowView.drawBackground`) reads `NSTableRowView.isSelected` live, every redraw.
- The **row text/icon color** (`SidebarCellView.setColors`) is applied imperatively, only at three moments: cell build, `refreshSelectionAppearance()` on selection-did-change, and `.agtermAppearanceChanged`.

When a re-tint is missed, the pill keeps tracking reality while the text keeps the *other* state's color. On the ~1/3 of bundled themes using the inverted-selection idiom (`selection-background == foreground`, e.g. the default-style dark themes, Black Metal family, Ayu), that stale color is exactly invisible: white text on a white pill, or the previously-selected row left dark-on-dark.

Two ways to miss the re-tint:

1. The cell builder computes `selected` from `selectedRowIndexes.contains(row(forItem:))`. During a `reloadData` or an expand/collapse animation `row(forItem:)` transiently returns `-1`, so the selected row's cell gets built with unselected (white) colors and nothing heals it until the next selection change. Sessions driven by an agent tick their OSC title constantly (spinner glyphs), so `reconcile` fires `reloadItem` on those rows continuously — plenty of chances to land mid-animation.
2. Inline rename paints the edit field white, and `restore` re-applies the row color via `row(for: field)`. If that lookup misses (row reloaded/scrolled during the edit), white sticks.

Measured across the 487 bundled themes: 153 (31%) render the stale state below 1.5:1 contrast in at least one direction, 94 in both.

## Fix

Make the row view the single source of truth for the cell's selection tint — the same `isSelected` the pill already draws from:

- `SidebarRowView.isSelected.didSet` re-tints the hosted cell on every selection flip.
- `SidebarRowView.didAddSubview` tints a cell the moment it attaches, so a cell materialized into an already-selected row picks up the correct color regardless of what the builder's `row(forItem:)` returned.
- Rename `restore` reads the row view's live `isSelected` instead of recomputing via `row(for:)`.

The build-time tint and `refreshSelectionAppearance` (theme changes) stay as-is; this just closes the desync windows.

## Testing

Build + `swift test` (1109 tests) + `make lint --strict` all green.

Text color isn't accessibility-observable, so there's no UI-test assertion for it (same as the disclosure-triangle and cursor solid/hollow cases) — verified by eye on two side-by-side dev instances (pre-fix vs post-fix) under `Ghostty Default Style Dark`, stressing per-row reloads against sidebar collapse/expand. Pre-fix flickers/sticks invisible; post-fix stays readable.

Repro on the running app: select a session, keep a title ticker running in it (`while true; printf '\e]2;tick %s\a' $RANDOM; sleep 0.2; end`), then repeatedly collapse/expand its workspace in the sidebar. Old build: the selected row's label blinks out. New build: stays readable.

## Bug video

https://github.com/user-attachments/assets/281c7afa-6456-4032-99bd-0f9f726a9392

### DIFF
--- a/.claude/rules/libghostty.md
+++ b/.claude/rules/libghostty.md
@@ -103,6 +103,20 @@ paths:
   The row draws the themed pill in `drawBackground(in:)` for every state,
   and the Coordinator's `refreshSelectionAppearance()` repaints the pills + re-tints the row text on
   selection change (AppKit won't redraw rows on its own with `.none`) and on `.agtermAppearanceChanged`.
+  **The row view is the single source of truth for the cell's selection tint.**
+  The pill reads `isSelected` live at draw time, but the text/icon color is applied imperatively
+  (`SidebarCellView.setColors`), so the two can desync when a re-tint event is missed —
+  the cell builder's `row(forItem:)` can return -1 mid-reload/expand-collapse animation
+  (the constant OSC-title `reloadItem` ticks make this window easy to hit),
+  and on the ~1/3 of themes using the inverted-selection idiom (`selection-background == foreground`,
+  e.g. `Ghostty Default Style Dark`) a stale tint renders the row text fully INVISIBLE
+  (white-on-white pill, plus the previously-selected row dark-on-dark).
+  `SidebarRowView` therefore re-asserts `setColors` from its own live `isSelected`:
+  its `didSet` re-tints the hosted cell on every selection flip,
+  and `didAddSubview` tints a cell the moment it attaches (superseding the builder's build-time guess).
+  Rename `restore` reads the same row-view `isSelected` instead of recomputing via `row(for:)`.
+  Text color is not accessibility-observable, so this is verified by eye — like the disclosure-triangle
+  and cursor solid/hollow cases.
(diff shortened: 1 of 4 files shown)


umputun (approved):
lgtm. Poked at the parts of a view-lifecycle change that usually bite: the `isSelected` didSet is re-entrancy-free (`setColors` only writes text/tint color, never `isSelected`), `didAddSubview` runs after the build-time tint so it wins, and the `cell.superview as? NSTableRowView` cast in rename `restore` is safe since the cell is a direct subview of the row view in a view-based outline. Build/test/lint green locally too, same as CI.

Issue #190, opened 2026-07-11
Add Intel Mac build support


vshuraeff:
Hi! 

Homebrew installation on Intel Macs currently fails:

```text
Error: agterm: This cask depends on hardware architecture being one of [{type: :arm, bits: 64}], but you are running {type: :intel, bits: 64}.
```

Would you consider publishing an Intel build as well—either a universal DMG or separate arm64 and x86_64 DMGs?

mvitaly:
Opened #191 with universal-binary (Apple Silicon + Intel) build support, so нищеброды like me who still haven't upgraded all my macs to arm64 can run agterm too 🙂

Verified on an Intel Mac (Xcode 26.3): Debug/Release builds, fat `agterm` + `agtermctl`, tests and lint clean. Once it lands and a release ships, the `umputun/homebrew-apps` cask would also need its `depends_on arch: arm64` restriction removed for `brew install` to work on Intel.

umputun:
went through this in detail on #191 (an implementation of exactly this) and #141 before it. The short version: I'm passing on Intel/universal builds.

it's not a code problem - #191 even verified on real Intel hardware. The blocker is that a universal release couples the deliberately-pinned libghostty build to an x86_64 slice I have no Intel hardware to re-verify on each bump, and the Intel base is shrinking anyway (Tahoe is the last Intel macOS, Rosetta 2 being wound down). Not worth the ongoing maintenance for a native GUI terminal I can't run on that hardware.

full reasoning is on #191, closing as not planned.

PR #191, opened 2026-07-11
feat: build universal binaries (Apple Silicon + Intel)

mvitaly:
Addresses #190 — makes both source builds and (from the next release) the shipped DMG work on Intel Macs by going universal.

## What changed

- `scripts/setup.sh` builds libghostty with `-Dxcframework-target=universal`, so the xcframework carries a fat `macos-arm64_x86_64` slice. Zig cross-compiles the non-native slice, so this works from either arch of build machine.
- `project.yml` sets `ARCHS: $(ARCHS_STANDARD)`. Debug keeps `ONLY_ACTIVE_ARCH=YES`, so dev builds stay single-arch and fast; Release/dist emit both slices. This is also what un-breaks source builds on Intel — the previous `ARCHS: arm64` pin produced binaries that can't run there.
- The `Bundle agtermctl CLI` post-build script passes matching `--arch` flags to SwiftPM (both slices when the app is universal, native-only in Debug), so a universal app doesn't ship a single-arch CLI.
- Docs synced per the keep-in-sync conventions: README (install + build-from-source), `site/index.html` + `site/docs.html` claims, the release-notes template in `release.sh`, and the CLAUDE.md quote of the zig command.

## Verified (on an Intel Mac, Xcode 26.3, macOS 15.7)

- Debug builds single-arch x86_64; app launches and works (rendering, control socket, `agtermctl tree`).
- Release build produces fat `agterm` and `agtermctl` (`lipo -info`: `x86_64 arm64`), so the release WMO path is exercised on both slices.
- `swift test` (1366 tests) and `make lint --strict` clean.

## Maintainer notes / trade-offs

- The universal zig build roughly doubles the one-time libghostty build (and the CI `build` job's setup step). Upstream's `universal` target also emits iOS slices agterm never links; the xcframework grows to ~540 MB on disk (gitignored).
- README/site now claim universal releases, which becomes true when the next release is cut — probably best merged alongside one. `release.sh` itself needed no arch changes, only its notes text; signing/notarizing the universal DMG is untested here since that flow is maintainer-only.
- The Homebrew cask in `umputun/homebrew-apps` will need its `depends_on arch` arm64 restriction dropped at the same release (that's the actual failure in #190).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -214,11 +214,14 @@ The app must build, `swift test` must stay green, and `make lint` must pass afte
 ## GhosttyKit.xcframework
 
 - Source: **built from upstream `ghostty-org/ghostty` source** by `scripts/setup.sh`,
-  pinned to the `GHOSTTY_REV` SHA (`zig build -Demit-xcframework=true -Dxcframework-target=native …`
+  pinned to the `GHOSTTY_REV` SHA (`zig build -Demit-xcframework=true -Dxcframework-target=universal …`
   with zig 0.15.2).
   Self-owned: the only inputs are upstream ghostty at a pinned commit and the zig/Metal toolchains —
   no third-party fork, no daily-build release that can be pruned.
   Bump `GHOSTTY_REV` deliberately when adopting a newer libghostty.
+  The `universal` target emits a fat `macos-arm64_x86_64` slice (plus iOS slices agterm never links),
+  so the app builds universal (`ARCHS: $(ARCHS_STANDARD)`) and Release/dist run on Apple Silicon and Intel;
+  zig cross-compiles the non-native slice from either arch of build machine.
 - **The pin is a pre-regression commit on purpose.**
   A libghostty `main` renderer regression introduced after `4dcb09ada` (2026-04-30) blanks the scrollback
   on a font-size *increase* (decrease is fine); it is NOT an agterm bug and no app-side change fixes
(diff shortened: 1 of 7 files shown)


umputun (commented):
thx for picking this up, and the real-Intel-hardware verification is exactly what #141 was missing - that was my main blocker there, I couldn't stand behind a slice nobody had run. Dropping the meaningless CI smoke check and keeping the diff this small are both improvements over that attempt.

I'm still going to pass though, for the reasons that aren't about whether it works today.

`-Dxcframework-target=universal` couples the GhosttyKit build to a slice I can't validate. You verified the current `GHOSTTY_REV` on Intel, but I bump libghostty regularly and have no Intel hardware to re-check the x86_64 slice on each bump, so every bump reintroduces an unverified slice. The universal target also pulls in iOS/iOS-sim slices agterm never links and grows the xcframework to ~540 MB, as your own notes point out.

and the timing hasn't moved since #141: Tahoe is the last macOS for Intel, Rosetta 2 is being wound down (warnings in 26.4, general removal in 28), so the Intel base is capped and shrinking. Not worth the ongoing per-bump maintenance for a native GUI terminal I can't run on that hardware.

the `ARCHS: arm64` -> `$(ARCHS_STANDARD)` half that unbreaks *source* builds on Intel is the reasonable part, but I don't want to ship or claim universal *releases* I can't verify, and it's not worth carrying just that on its own.

closing as not planned, same as #141. appreciate the effort.

PR #195, opened 2026-07-12
feat: per-session file-tree panel

jokius:
## Summary

- Add a per-session file-tree panel as a third column right of the terminal — an `NSOutlineView` rooted at the active session's cwd, toggled by `cmd+ctrl+e` (rebindable via `toggle_file_tree`) and persisted per session.
- Interact via double-click (open in a configurable editor, or expand a folder), Space for Quick Look, and a context menu (Open / Reveal in Finder / Copy Path / Insert Path).
- Auto-refresh the tree on filesystem changes via FSEvents; the refresh button re-roots to the session's current cwd.
- Drag the divider for a per-window panel width (clamped, persisted), mirroring the sidebar width.
- Add Settings ▸ General ▸ Files to choose which app files open in (System Default or any `.app`).
- Add a `session.filetree [on|off|toggle|refresh|reroot <path>]` socket command with `fileTreeVisible`/`fileTreeRoot` tree read-back, plus the `agtermctl session filetree` subcommand.

## Notes

The host-free model (`FileTree`) and control logic stay in `agtermCore` with unit tests; the app target owns the AppKit panel (`FileTreePanel`) and the filesystem side effects (enumeration, FSEvents). Panel width is per-window (mirrors `sidebarWidth`); the editor choice is a GUI-only setting (keep-in-sync exempt, like other settings). Rebased on current `master`.

## Validation

- `make build`
- `make lint`
- `git diff --check`


umputun:
going to decline this. A file tree doesn't belong in agterm.

agterm is a terminal, not a file manager. The README is explicit about staying minimal: it covers the core use cases and stops there, and anything past the defaults is built over the control API and CLI, not shipped as a built-in. I'm keeping it that way rather than growing it into a do-everything app.

file browsing is already covered without this: a TUI file manager (yazi/lf/nnn) in a floating overlay, plus reveal-in-Finder and Quick Look from the shell. There's a screenshot of exactly that in the README. A built-in tree just duplicates a workflow that already works and adds a permanent AppKit surface to maintain.

there's also the process. No issue, no discussion, straight to a PR across 36 files. A change this size needs the scope settled first, and the scope answer here is no.

PR #208, opened 2026-07-13
feat: add Pi agent-status support to Install Agent Status Hooks

taras-mrtn:
## What

Adds agent-status support for **Pi** (the `@earendil-works/pi-coding-agent` runtime) to **Help ▸ Install Agent Status Hooks…**. When Pi is set up on the machine, the installer drops a lifecycle extension that reports Pi's state onto its session's sidebar row — `active` while Pi works, `completed` when it settles — the same auto-wiring agterm already does for Claude Code and Codex.

## Why

agterm's whole point is telling which of several running agents needs you, and it already auto-wires status for Claude Code and Codex. Pi is another coding agent in the same class, so a Pi user running it inside agterm got a blank row where the others glow. This closes that gap, following the exact pattern the Codex lifecycle-hooks work (#194) established: drive status from the agent's own lifecycle events rather than scraping output.

There's no issue requesting this — it's an extension of the actively-maintained agent-status subsystem in its established shape (cf. #56 fish, #194 Codex, #130 pane-aware).

## How

**Bundled extension** (`agterm/Resources/agent-status/pi/agterm-status.ts`, a new `project.yml` Resources folder mirroring the rest of `agent-status/`):
- `agent_start` → `active --blink`; `agent_settled` → `completed --auto-reset` (settled = after Pi's automatic retries, compaction retries, and queued continuations, so it doesn't fire mid-turn).
- It shells out to the **same `agterm-agent-status.sh` wrapper** the Claude/shell hooks use — one status-reporting path, not a parallel one — and **no-ops when `AGTERM_SESSION_ID` is unset**, so it's harmless in a non-agterm Pi session. Errors are swallowed so status reporting can never interrupt Pi's agent loop.
- Uses a **type-only import** of the Pi SDK, so it needs no installed package at runtime.

**Installer** (`installPiExtension`, app target) mirrors the Codex path:
- **Gated on `~/.pi/agent` existing** (like the `~/.codex` gate), so a non-Pi user's home isn't seeded.
- Copies to `~/.pi/agent/extensions/agterm-status.ts` (Pi's auto-discovered global dir), **symlink- and posix-mode-preserving**.
- **Ownership-marker protected**: a same-named extension without agterm's `// agterm-pi-status-extension` marker is treated as user-owned and left untouched; a managed one refreshes in place. No backup for the managed refresh (the extension carries no user state).
- Idempotent; the result is surfaced in the install alert (installed / already-current / user-owned / unreadable / no-Pi).

The host-free path/marker/ownership logic lives in `agtermCore.AgentHooksInstall`; the app target owns the filesystem glue — same split as the rest of the installer.

### Keep-in-sync

Control-API **exempt**: this drives the existing `session status` command, so no new `Command`/dispatcher/CLI/round-trip and no command-count change (same as the Codex/fish installer additions). The doc surfaces that *do* apply are updated: README, `.claude/rules/control-api.md`, agent-skill `troubleshooting.md`, `docs/troubleshooting.md`, and the website (`docs.html`, `index.html`, `llms.txt`).

### Known limitation

Pi status is **`active` → `completed` only, never `blocked`**. Pi has no native permission-prompt or structured-question event, so the extension can't detect "waiting for you" and deliberately does not infer `blocked` from assistant prose (a waiting Pi agent won't glow amber). Documented in the README and troubleshooting. This mirrors how #194 scoped what it could and couldn't detect.

## Test plan

- `cd agtermCore && swift test` — **1516/1516 pass**, incl. 2 new cases: Pi extension path resolution and the ownership-marker overwrite policy.
- `swiftlint --config .swiftlint.yml` on the changed Swift files — 0 violations.
- `make build` — succeeds; verified the bundled `pi/agterm-status.ts` lands in `Contents/Resources/agent-status/pi/`.
- Validated the website JSON-LD parses (fixed a trailing-comma the FAQ edit introduced).
- Manual: with `~/.pi/agent` present, ran the installer → extension copied to `~/.pi/agent/extensions/`; re-run is a clean no-op; an unmarked same-named file is left untouched with a warning. With no `~/.pi/agent`, the Pi step is skipped and the alert says so.


umputun (commented):
looks right overall, follows the Codex path. Four things to fix before merge:

**1. Pi step aborts the whole install instead of degrading** (`agterm/AgentHooksInstaller.swift`, `installPiExtension` / `install`)

`installPiExtension()` runs last in `install()`, and its errors propagate uncaught to `run()`, which shows only "Install Failed". So a Pi-only problem hides that the Claude, Codex, and shell integrations already installed. Every other integration degrades to a result value (`settingsSkipped`, `CodexResult`); Pi is the only one that aborts. The two build-integrity throws are effectively unreachable (the folder copy already guards them), so the realistic trigger is a filesystem error on `~/.pi/agent/extensions` (a non-directory path, or an unwritable dir). Fix: catch the FS errors in `installPiExtension` and return a `PiResult` warning case instead of throwing/propagating, so it degrades like its siblings.

**2. shell integration headers not updated for Pi** (`agterm/Resources/agent-status/shell/integration.sh:9`, `integration.fish:9`)

README now says "Claude Code, Codex, and Pi are excluded by default", but both bundled headers still read `Claude Code and Codex are intentionally NOT in the default list`. Update both to include Pi.

**3. inline existence probe re-implements `readExistingConfig`** (`installPiExtension`, ~line 247)

`(try? fm.attributesOfItem(atPath:)) != nil` collapses any non-ENOENT stat error to "absent", which then bypasses the marker gate. The sibling `mergeClaudeSettings` / `mergeCodexConfig` go through `readExistingConfig(at:)` (nil=absent / throw=unreadable / contents=present). Reuse it here so the exists-but-unreadable case folds into `.unreadable` uniformly. Not a security issue (local install of the user's own file), just consistency with the sibling paths.

**4. `PiResult` missing the doc comments `CodexResult` has** (`agterm/AgentHooksInstaller.swift`, ~line 48)

`CodexResult` documents the enum and `isWarning`; `PiResult` documents neither, despite being a deliberate mirror. Add matching comments.

**fyi, separate from the above (not this PR's job):**

- agtermctl's socket client (`SocketClient.readLine`) has no read deadline and the wrapper runs `agtermctl` as a child, not `exec`, so a hung agterm socket can block a caller. Pre-existing and shared by the Claude/Codex hook callers; the Pi extension's 1s timeout actually makes it the only bounded one, so nothing worse here. Worth hardening the client on its own sometime.
- three Pi-SDK behaviors the code leans on but I couldn't confirm from the tree: extension auto-discovery in `~/.pi/agent/extensions/` with no registration step, `pi.exec` inheriting env to the wrapper, and `timeout: 1_000` being milliseconds. Presumably covered by your manual test, just flagging.

taras-mrtn:
Addressed all four in the follow-up commit:

**1.** `installPiExtension` now degrades like its siblings: the directory-create + write are caught and mapped to a new warning-level `PiResult.writeFailed` (surfaced as a Pi-only alert line), so a filesystem problem on `~/.pi/agent/extensions` no longer aborts the whole install and hides that the Claude/Codex/shell steps completed. The two build-integrity throws stay as throws — a missing bundled file or marker is a broken build, which should be "\Install Failed'".

**2.** Both shell-integration headers now read "Claude Code, Codex, and Pi are intentionally NOT in the default list" ("hooks/extensions", since Pi's is an extension), matching the README.

**3.** The inline existence probe is gone — `installPiExtension` now reads the destination through `readExistingConfig(at:)` like `mergeClaudeSettings`/`mergeCodexConfig`, so a non-ENOENT stat error folds into `.unreadable` instead of reading as "absent" and bypassing the marker gate.

**4.** `PiResult` and its `isWarning` now carry doc comments matching `CodexResult`.

On the FYIs: agreed the socket-client read deadline is pre-existing and out of scope here. On the three Pi-SDK behaviors: two are documented in Pi's extension docs (packages/coding-agent/docs/extensions.md) — global auto-discovery is `~/.pi/agent/extensions/` with no registration step, and `pi.exec`'s `timeout` is milliseconds (the doc example uses `timeout: 5000`). Env inheritance to the child isn't explicitly documented; it held in my manual test and the extension no-ops on a missing env either way.

`swift test` 1516/1516, `swiftlint --strict` clean, `make build` succeeds.

umputun (approved):
all four addressed correctly: the degrade-to-`.writeFailed` on a destination FS error, the Pi line in both `integration.sh`/`integration.fish` headers, the `readExistingConfig` reuse for the destination read, and the `PiResult` doc comments. `swift test` green (1516), swiftlint clean, mergeable. Also confirmed the three Pi-SDK behaviors from the round-1 fyi against Pi's extension docs: global auto-discovery under `~/.pi/agent/extensions/`, `pi.exec` timeout in ms, and the `agent_start`/`agent_settled` events all check out.

two minor things, neither blocking:

- `AgentHooksInstaller.swift:245`: `String(contentsOf: source)` is outside the new `.writeFailed` do/catch, so a read failure on the just-copied source still escapes to "Install Failed" after the Claude/Codex/shell steps ran, the same partial-success masking the do/catch closed for the destination write. Near-unreachable in practice (source is agterm's own file, copied microseconds earlier in the same run), so folding it into a degrade path with its own message is optional polish, not something I'd block on.

- the ownership-marker gate is skipped for a *dangling* symlink at the destination: `readExistingConfig`'s `fileExists` reports a dangling link as absent, so `mayOverwritePiExtension` returns true and the write follows the link. Negligible: a dangling link has no target to clobber, and it's the same shared-helper behavior the Claude/Codex paths already have, the reuse in #3 just routes Pi onto that path too. Flagging only for completeness.

good to merge.

PR #229, opened 2026-07-15
fix: make site navigation responsive on mobile

Hormold:
## Problem

The site header only exposed its desktop navigation, so links overflowed the viewport on phones. The documentation pages also kept their fixed-width sidebar and two-column layout, squeezing the article into an unreadable strip.

## Fix

- Add a compact mobile menu while keeping the existing desktop navigation unchanged.
- Keep the header pinned at the top while scrolling on mobile.
- Turn the documentation sidebar into an off-canvas contents drawer with an overlay and automatic close after section navigation.
- Switch documentation content to a single-column mobile layout and size the drawer to the dynamic viewport height.
- Allow long inline command tokens and error labels to wrap on narrow screens instead of causing horizontal overflow.

The implementation stays within the existing static site: no dependencies or script bundle were added.

## Validation

- Manually tested `/`, `/docs`, and `/commands` in a real browser at 320px, 390px, 720px, 721px, and 1280px widths.
- Verified mobile menu open/close behavior, sticky header behavior, drawer scrolling, overlay close, and section-link navigation.
- Verified zero horizontal overflow on `/commands` at 320px and 390px.
- Verified the desktop navigation and sticky documentation sidebar remain unchanged above the mobile breakpoint.
- Verified no browser console errors.
- `git diff --check`

## Screenshots

**Commands - Before / After**

comparison-commands-mobile

**Docs - Before / After**

comparison-docs-mobile

**Landing Page - Before / After**

comparison-home-mobile

### DIFF
--- a/site/commands.html
+++ b/site/commands.html
@@ -133,7 +133,14 @@
           <details class="site-nav-mobile">
             <summary>menu</summary>
             <nav aria-label="Mobile navigation" onclick="this.parentElement.open=false">
-              <label for="doc-drawer">contents</label>
+              <button
+                type="button"
+                class="doc-drawer-trigger"
+                aria-controls="doc-navigation"
+                onclick="document.getElementById('doc-drawer').checked=true; document.querySelector('#doc-navigation a').focus()"
+              >
+                contents
+              </button>
               <a href="/">← home</a>
               <a href="/docs">docs</a>
               <a href="https://github.com/umputun/agterm">GitHub ↗</a>
@@ -143,7 +150,12 @@
       </div>
 
       <input class="doc-drawer-toggle" type="checkbox" id="doc-drawer" />
-      <label class="doc-drawer-overlay" for="doc-drawer" aria-label="Close documentation navigation"></label>
+      <button
+        type="button"
+        class="doc-drawer-overlay"
+        aria-label="Close documentation navigation"
+        onclick="document.getElementById('doc-drawer').checked=false; document.querySelector('.site-nav-mobile summary').focus()"
+      ></button>
 
       <!-- ===== LAYOUT ===== -->
       <div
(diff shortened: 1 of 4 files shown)


umputun (commented):
one real accessibility issue, the rest is minor. Not a blocker.

**docs drawer isn't keyboard/screen-reader operable** (`site/docs.html`, `site/commands.html`, `site/style.css`)

the off-canvas contents drawer is opened by `<label for="doc-drawer">contents</label>` and closed by the overlay `<label>`, but `#doc-drawer` is `display: none`, so the checkbox is out of the tab order and out of the accessibility tree. The `<label>`s have no keyboard/AT activation of their own, so a keyboard or VoiceOver user on a narrow viewport can't open the section drawer (WCAG 2.1.1). The top nav dropdown is fine, native `<details>`/`<summary>` is keyboard-operable, so this is only the docs drawer. Content stays reachable by scrolling, so it's degraded nav rather than lost content, but the toggle is interactive UI that isn't keyboard-operable.

fix: make the open trigger and the overlay real `<button>`s with an inline `onclick` toggle (consistent with the close-on-select `onclick` you already use), or visually-hide the checkbox (sr-only clip) instead of `display: none` so it stays focusable. Esc-to-close would be a nice add.

**minor**

- `span[style*="white-space: nowrap"]` in `site/style.css` couples to the exact inline-style string, so it silently stops matching if a token's style is written differently (e.g. no space after the colon). Fine as a pragmatic fix for the inline-styled pages, just flagging the fragility.
- `overflow-x: clip` is added on `index.html` only; the `docs.html`/`commands.html` roots have no `overflow-x` guard. Defensible (their sticky nav already worked and the drawer overflows left), but a matching `overflow-x: clip` there is cheap insurance against future right-overflow on the exact pages this targets.

the PR body still has the "screenshots will be attached here" placeholder, worth attaching before merge.

Hormold:
The screenshots were added a couple of minutes after the PR was opened, so the ai reviewer may have caught an earlier version. The accessibility issue and the other minor points are addressed in the latest commit :-)

umputun (approved):

PR #231, opened 2026-07-16
Return to the previously-active session when the active one is closed

olomix:
Closing the active session reselects a neighbor by position, which makes for unintuitive jumps: open a session, close it, and you land next to where it was instead of back where you came from. As agreed in discussion #147, the close now returns to the most-recently-active surviving session, reusing the recency stack Ctrl-Tab already maintains, with no setting.

The pick is scoped to the closing session's own workspace (and to the flagged set in flagged sidebar mode) so a close never yanks you into another workspace. When the close leaves that scope empty — the workspace was emptied, or no other session there is flagged — the MRU widens instead of jumping positionally into the first workspace. An empty recency stack falls back to the existing positional reselection, kept within the flagged set in flagged mode so the sidebar never ends up with no selected row. All three close paths (hard close and both undoable soft closes) share one host-free helper in agtermCore, covered by unit tests.


umputun (approved):
lgtm. Built and ran the branch tests, all green, and swiftlint --strict clean.

checked the parts most likely to bite and they hold: the soft-close paths keep the closing session in `sessionRecency` for undo, but the MRU pick still excludes it because `scope` comes from the pruned tree rather than the recency stack. The `nearestInScopeTarget` index math stays in pruned-tree coordinates so the end slices are safe, and `workspaces[location.workspaceIndex]` can't go stale because closing a session never removes its workspace.

PR #232, opened 2026-07-17
feat: optional notification sound (Settings > Notifications)

ZUBOV-ILLIA:
## What

Settings ▸ Notifications gains a **Notification sound** picker: a system sound (default **None**) attached to each delivered desktop notification — both the terminal OSC 9/777 path and the control channel's `notify`.

The motivation is the workflow agterm is built around: an agent finishing a turn in a background session already raises a banner, bumps the badge, and (opt-in) bounces the Dock icon — but silently, so with agterm in the background you only notice when you look. This adds the audible cue other terminals give (e.g. iTerm2's notification sound), opt-in and off by default.

## Behavior

- The sound **rides the banner**: it is attached as `UNNotificationSound` on the notification content in BOTH delivery paths (the OSC `notify` and the control `send` / `agtermctl notify`), so it follows the banners toggle, the macOS notification authorization, and Do Not Disturb / Focus — a banners-off user is never audibly interrupted (reworked per review, was raw `NSSound` at first).
- The OSC path's focus suppression still applies: the pane you're typing in never sounds.
- `start()` requests `.sound` alongside alert + badge, and `willPresent` includes `.sound`, so a banner for a session you're NOT looking at dings even while agterm is frontmost.
- The picker previews the selection via `StatusSoundPlayer` (a deliberate click always sounds), the way the Agent Status blocked-sound picker does.
- `AppSettings.notificationSoundName: String?` — nil/empty = silent (the default); picking None maps back to nil so `settings.json` stays minimal. NOT a ghostty key. The `NotificationManager.notificationSoundName` mirror is pushed by `SettingsModel.applyNotificationSound` alongside the other two `NotificationManager` mirrors and read on the next notification (no re-render needed).
- Testing heads-up: since playback routes through the OS, the sound needs the app's notification permission — if macOS didn't prompt on first run, make sure System Settings ▸ Notifications ▸ Agterm has **"Play sound for notifications"** enabled:

System Settings, Notifications, Agterm — the Play sound for notifications toggle

## Keep-in-sync

- README + `site/docs.html` notification/settings paragraphs updated; `.claude/rules/notifications.md` + `settings.md` notes added; CHANGELOG untouched (release-only).
- Control coverage: a GUI-only settings picker, keep-in-sync EXEMPT per the `dockBounce`/blocked-sound precedent (only `theme.set`/`config.reload` touch settings over the socket); the per-call audible nudge remains `session.status --sound`. A per-call `notify --sound` flag would compose naturally as a follow-up if wanted.

## Tests

- `AppSettingsTests.notificationSoundNameRoundTripsAndIsNotAConfigLine` — host-free round-trip, legacy decode to nil, not-a-config-line.
- `SettingsUITests.testNotificationSoundPickerPersists` — picker persistence, mirroring the dock-bounce test.
- `swiftlint --strict` clean; manually tested by a human on a Debug build — including a re-test after the review rework: the sound arrives together with the banner, banners off → full silence, Do Not Disturb → silence, picker preview unaffected.

## Screenshots

Notifications settings with the Notification sound picker

The notification-sound picker open, listing the system sounds


umputun (commented):
reviewed the branch, build + host-free tests + swiftlint --strict all pass. One change worth making before merge, plus a design question and a minor note.

**Requested: gate the sound to the banner toggle.** In both delivery paths (the OSC `notify()` and the control `send()` in `NotificationManager.swift`), `playNotificationSound()` runs before `guard bannersEnabled else { return }`, so the sound fires even when banners are switched off. The badge and the Dock bounce are deliberately banner-independent, but a sound is an active interruption rather than a glance-level cue, so a user who turned banners off has effectively asked not to be interrupted. Move the call below the `guard bannersEnabled` in both paths.

**Open question: playback path and Do Not Disturb.** The sound plays through `StatusSoundPlayer` (raw `NSSound`), which bypasses Do Not Disturb / Focus and the per-app notification-sound setting in System Settings. Setting `content.sound = UNNotificationSound(named:)` on the `UNMutableNotificationContent` you already build would route it through the OS instead: DND-respecting, and inherently gated to the banner (so it also covers the change above). Keeping raw `NSSound` is consistent with `session.status --sound` and the blocked sound, so it's not a blocker, but worth weighing alongside the banner-gating.

**Minor: shared sound throttle.** `notificationSoundName` and `blockedStatusSoundName` / `session.status --sound` share the single 200ms name-keyed throttle on `StatusSoundPlayer.shared`. If both are set to the same sound and a status change plus a notification land within 200ms, the second is de-bounced. Narrow, and coalescing two identical dings is probably desirable, so no change needed. Flagging it only so the cross-feature throttle behavior stays a conscious choice.

**Optional, pre-existing:** `StatusSoundPlayer`'s type godoc still says it's "owned by ControlServer" and its `standardNames` doc says the names are "used only to suggest valid values in the unknown sound error." Both now understate reality: the Settings pickers use `standardNames`, and `NotificationManager` becomes a third `.play` caller with this PR. The file isn't in the diff, but it's a natural moment to touch up those two comments.

ZUBOV-ILLIA:
Thanks for the thorough review — much appreciated! All four points are addressed in 228fe98:

**Banner gating + the playback path** — went with the `UNNotificationSound` route you suggested, which nicely solves both at once: the sound is now attached to the banner's content in both delivery paths, so it follows the banners toggle, the macOS notification authorization, and Do Not Disturb / Focus — a banners-off user is never audibly interrupted. `start()` now requests `.sound`, and `willPresent` returns `[.banner, .list, .sound]` so a foreground banner (for a session you're not looking at) dings too. The stored name maps to a sound file — `.aiff` is assumed when there's no extension (the system sounds' format), and `default`/`beep` map to `UNNotificationSound.default`.

**Shared throttle** — became moot for the notification sound with the rework: the OS owns playback now, and same-pane repeats already coalesce by request identifier. `SoundThrottle` stays what it was — a status-sound concern only.

**StatusSoundPlayer comments** — touched up both: the godoc now describes the shared singleton and its actual callers (ControlServer + the Settings pickers' previews), and `standardNames` mentions it is also the pickers' option list.

Everything was re-tested manually by a human on a Debug build after the rework: the sound arrives together with the banner, banners off → full silence, DND → silence, and the picker preview is unaffected. The README / site / rules docs were updated to match the new behavior.

One testing note: since playback now routes through the OS, the sound needs the app's notification permission. If macOS didn't prompt on first run, check System Settings ▸ Notifications ▸ Agterm and make sure **"Play sound for notifications"** is enabled — here's where it lives:

System Settings, Notifications, Agterm — the Play sound for notifications toggle

umputun (commented):
the banner-content rework holds up: the sound follows the banners toggle and DND without a separate guard, in both paths.

two things before merge:
- now that `willPresent` returns `.sound`, please confirm by ear that an unconfigured banner (sound = None, the default) stays silent in the foreground. It should, since `content.sound` is nil, but it's the one default-path behavior the rework touches.
- `AppSettings.notificationSoundName`'s doc comment says "resolved by `NSSound(named:)`", but the delivered sound goes through `UNNotificationSound(named:)` (with the `.aiff` suffix); only the picker preview uses `NSSound`. Worth fixing so the comment matches the delivery path; looks copied from `blockedStatusSoundName`.

ZUBOV-ILLIA:
Thanks for the second look - glad the banner-content rework held up!

Both points are handled:

- **Silent default confirmed by ear.** I (Ilya) tested it personally on a Debug build: with **Notification sound = None** (the default), a notification delivered while agterm is frontmost shows the banner with no sound - `content.sound` stays nil, so `willPresent` returning `.sound` changes nothing for the default path. As a control, setting it to Glass and triggering a background banner dings as expected.
- **Doc comment fixed** (`61fe2e7`): `AppSettings.notificationSoundName` no longer claims `NSSound(named:)` or banner-independence - it now describes the actual delivery (`UNNotificationSound(named:)` with the `.aiff` suffix, riding the banner, gated by notifications + DND), and notes that only the picker preview uses `NSSound`.

umputun (commented):
both resolved. The doc comment now matches the actual `UNNotificationSound` delivery, and your silent-default test lines up with the code (nil `content.sound`, so `willPresent`'s `.sound` is a no-op on the default path). Nothing else from me.

PR #233, opened 2026-07-17
feat: add inline '+' button on workspace rows to create new sessions

wievtsal:
## Summary

Currently there is no single-click way to start a new session in a workspace — the only options are the bottom-bar menu (two clicks: open menu → New Session) or the right-click context menu (two clicks: right-click row → New Session). This adds a `+` button directly on each workspace row so a new session is one click away.

- Adds a small `+` button to the trailing edge of each workspace row in the sidebar, between the workspace name and the status/badge area
- Clicking it creates a new session in that workspace — identical to right-click → New Session, using the same `addSession(toWorkspace:cwd:)` path including the new-session-directory setting
- Guards `handleSingleClick` so clicking `+` never also triggers the expand/collapse toggle; `addSessionButtonClicked` additionally cancels any pending deferred toggle
- Button tint tracks the row's theme color (selected vs unselected) via `setColors`

## Test plan

- [x] Build succeeds (`make build`)
- [x] SwiftLint clean (`make lint`)
- [x] `SidebarUITests/testInlineAddSessionButtonCreatesSession` passes — clicks the inline button, verifies workspace gains a second session
- [ ] Manual: click `+` on a workspace row → new session appears and is selected
- [ ] Manual: button tint matches row text color on selection and theme changes

Screenshot 2026-07-17 at 13 22 58

### DIFF
--- a/agterm/Views/SidebarRowViews.swift
+++ b/agterm/Views/SidebarRowViews.swift
@@ -19,6 +19,47 @@ final class SidebarCellView: NSTableCellView {
     /// against toggling expansion when the click lands on this button.
     var addButton: NSButton?
 
+    /// Width of `addButton`, toggled between 0 (hidden) and the glyph width by `setAddButtonVisible`
+    /// — the same collapse-the-slot convention as `StatusIconView.widthConstraint`, so an idle row's
+    /// name reclaims the space. Set by the cell builder alongside `addButton`.
+    var addButtonWidthConstraint: NSLayoutConstraint?
+
+    private static let addButtonWidth: CGFloat = 16
+    private var hoverTrackingArea: NSTrackingArea?
+
+    /// Reveals or collapses the inline "+": the Finder/Xcode hover convention, so an idle row shows
+    /// no button and the roll-up badge keeps its slot. Driven by `mouseEntered`/`mouseExited` and
+    /// reset hidden when a reused cell is reconfigured in the row provider. No-op for session cells.
+    func setAddButtonVisible(_ visible: Bool) {
+        guard let addButton, let addButtonWidthConstraint else { return }
+        addButton.isHidden = !visible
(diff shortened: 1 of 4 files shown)


umputun:
the code path here is fine, the blocker is the design.

an always-visible `+` on every workspace row is more UI than agterm wants on an idle row. Creating a session in a workspace is an infrequent action that already has a few entry points (right-click → New Session, the bottom-bar session menu, ⌘N, the command palette), so a permanent button on every row costs more than the one click it saves. It also sits right where the roll-up unseen badge shows on a collapsed workspace, and it trims the name's width on every row.

what I'd take is the same button revealed on hover: show the `+` only while the pointer is over the workspace row (the Finder/Xcode sidebar convention), so idle rows stay clean and it stops competing with the badge.

the pieces that fit the existing code:
- add an `NSTrackingArea` to `SidebarCellView`, same options as `GhosttySurfaceView+Tracking.swift` (`.mouseEnteredAndExited, .activeInKeyWindow, .inVisibleRect`) with an `updateTrackingAreas()` override, and toggle the button from `mouseEntered`/`mouseExited`.
- collapse the button to zero width when hidden, the way `StatusIconView` already does with its width constraint, so a hidden `+` reserves no space and the name reclaims the slot. Reset it hidden when a reused cell is reconfigured in the row provider.
- the UI test then needs to `hover()` the row before the button is hittable.

happy to merge it once it works on hover.

umputun (commented):
the hover-reveal resolves the design blocker, and the tracking-area / collapse / reuse-reset pieces all check out.

one non-blocking nit: if a workspace row gets reconfigured while the pointer is already over it (a `reloadData` mid-hover), the `+` resets hidden and won't come back until you move off the row and back on, since the reveal only fires on `mouseEntered`. Minor, fine to leave.

PR #234, opened 2026-07-17
Duplicate session context menu

dimetron:
Add "Duplicate Session"

Duplicates a session into a fresh shell rooted at the source's directory, inserted directly after it in the same workspace. Only the directory carries over — it's a plain New Session seeded with the source's focusedCwd (the cwd the sidebar row shows and "Reveal in Finder" opens), not a state clone: auto basename, no split/scratch/status, unflagged, no initial command.

Exposed on all three callers of the AppActions/AppStore seam, per the keep-in-sync convention.

Surfaces

- Sidebar — right-click a session row → Duplicate Session (single-selection only, next to Rename). A background window's menu acts on its own row, not the frontmost store's.
- Control API — new session.duplicate command. Takes no options (the target names both the workspace and the cwd); focuses the duplicate when it lands in the frontmost window and returns its id. Read-back is tree — the new session appears right after its source carrying the same cwd.
- CLI — agtermctl session duplicate --target <id>, echoes the new session id.

Implementation

- Host-free logic in agtermCore: AppStore.duplicateSession(_:) (AppStore+Duplicate.swift), the Command.sessionDuplicate case (ControlProtocol), and routing (ControlDispatcher).
- App target stays a thin adapter: AppActions.duplicateSession(_:in:) (select + focus), ControlServer arm, sidebar context-menu item.

Docs (keep-in-sync surfaces)

- Bundled agent skill (SKILL.md/reference.md/examples.md, command count 60 → 61)
- site/commands.html, site/docs.html, README.md
- .claude/rules/control-api.md, .claude/rules/sidebar.md

Tests

- AppStoreDuplicateTests.swift — cwd carry-over, insertion position, nil-on-miss
- Dispatcher round-trip + ControlProtocol coding + agtermctl CLI parse tests
- Extracted a shared MockControlActions for the dispatcher tests

Verification: swift test green (1263 tests), make lint --strict clean, Debug app builds.

Screenshot 2026-07-17 at 3 39 02 PM


umputun (changes requested):
needs the e2e before merge, rest is minor.

the `session.duplicate` control command has no end-to-end test. The host-free layers are covered (`AppStoreDuplicateTests`, dispatcher round-trip, protocol coding, CLI parse), but the app-side arm in `agterm/Control/ControlServer+SessionActions.swift` (resolver wiring, the `store === library.activeStore` focus branch, the "could not duplicate session" error path, the id echo) is hit by nothing. Every sibling create/mutate command has a case in `agtermUITests/ControlAPIUITests.swift` (`testSessionNewAndClose`, `testSessionMoveToAnotherWorkspace`); this one needs the same. Add a `testSessionDuplicate` that runs `session duplicate --target <id>` and asserts via `tree` that the new node is inserted directly after its source with the source's cwd and gets focused. That also exercises the read-back point below.

minor:

- read-back wording overstates for split sessions. The duplicate seeds `initialCwd` from `session.focusedCwd` (focus-aware) at `agtermCore/Sources/agtermCore/AppStore+Duplicate.swift:16`, but `tree` reports each node's `cwd` from `session.effectiveCwd` (always the primary pane) at `agtermCore/Sources/agtermCore/AppStore.swift:207`. So when the source is a split focused on its non-primary pane in a different dir, the source node's `tree.cwd` (primary) and the duplicate's `cwd` (focused pane) differ, and a script comparing the two hits a mismatch. Holds for every non-split case. The "carries the same cwd" read-back is repeated in `.claude/rules/control-api.md`, `.claude/rules/sidebar.md`, the agent skill (`SKILL.md`/`reference.md`/`examples.md`), `site/commands.html`, and the `duplicateSession` doc comment, so pls qualify it (the duplicate carries the source's focused-pane cwd, which for a split focused off the primary differs from the source node's `tree.cwd`).

- the `focusedCwd`-over-`effectiveCwd` seed choice isn't pinned by a test. Swapping `AppStore+Duplicate.swift:16` to `effectiveCwd` still passes all four `AppStoreDuplicateTests`: `duplicateSessionCopiesOnlyTheDirectory` sets `isSplit`/`hasSplit` but never `splitFocused`/`splitCwd`/`splitSurface`, so `focusedCwd` never takes its split branch. Add a case with `splitFocused = true`, `splitSurface = FakeSurface()`, and `splitCwd` different from the primary cwd, asserting `dupe.initialCwd` equals the split cwd.

- `AppActions.duplicateSession(_:in:) -> Session?` has no consumer: `menuDuplicate` discards it and the control path calls `store.duplicateSession` directly. Optional, drop it to `Void` to match `newSession`/`openDirectory`. Your call.

umputun:
one more, bigger than the nits above: Duplicate Session is wired only to the sidebar context menu, `session.duplicate`, and the CLI. New/Close/Rename Session are also in the menu bar, the ⌃P command palette, and keymap-bindable; Duplicate is in none of those (thinner even than Reveal in Finder, which at least has a menu item). For a first-class session action it's worth wiring across the board:

- add a `BuiltinAction.duplicateSession` case (`agtermCore/Sources/agtermCore/BuiltinAction.swift`) so it's `map`-bindable in `keymap.conf` and the palette can show its shortcut hint. No default chord needed, `renameSession` has none either.
- add a `PaletteCommand.duplicateSession` case plus its arm in `runPaletteCommand` (`agterm/AppActions+Palette.swift`) so it shows in the ⌃P command palette.
- add a menu-bar `Button("Duplicate Session")` in `agterm/agtermApp+Menus.swift` next to Rename Session / Reveal in Finder.

the palette, menu, and keymap all act on the ACTIVE session, so they need an active-session entry point (`duplicateActiveSession()`) alongside the existing row-scoped `duplicateSession(_:in:)`, the same split as `renameActiveSession()` vs the sidebar's rename. And document the new `duplicate_session` action wherever the bindable actions are listed (README keybindings plus the keymap docs).

dimetron:
Addressed the review in 4d5b628:

**e2e (blocker):** Added `testSessionDuplicate` in `ControlAPIUITests` — runs `session duplicate --target <id>` and asserts via `tree` that the new node lands directly after its source, carries the source's cwd, and becomes active (exercises the ControlServer arm: resolver wiring, the frontmost-store focus branch, and the id echo). Passes.

**Cross-the-board wiring** — Duplicate is now a first-class session action:
- `BuiltinAction.duplicateSession` (`duplicate_session`, keyless like `rename_session`) so it is `map`-bindable and the palette can show its shortcut hint
- `PaletteCommand.duplicateSession` + arm in `runPaletteCommand` (⌃P command palette)
- menu-bar **Duplicate Session** button next to Rename Session / Reveal in Finder
- `AppActions.duplicateActiveSession()` active-session entry point, mirroring `renameActiveSession()` vs the row-scoped `duplicateSession(_:in:)`
- documented `duplicate_session` in the README keybindings list, `site/docs.html`, and the agent skill

**Minor:**
- Pinned the `focusedCwd`-over-`effectiveCwd` seed with a split-focused-off-primary unit test (`duplicateSessionSeedsFromFocusedSplitPane`) — swapping the seed to `effectiveCwd` now fails there.
- Qualified the read-back wording everywhere it appeared (control-api, sidebar, SKILL/reference/examples, commands.html, and the `ControlServer`/`AppStore` doc comments): the duplicate carries the source's focused-pane cwd, which equals the source node's `tree.cwd` except when the source is a split focused off its primary pane (then `tree.cwd` reports the primary and they differ).
- Dropped `AppActions.duplicateSession(_:in:)` to `Void` (no consumer used the returned id).

Verified: `swift test` green (1277 tests), Release build clean, `swiftlint --strict` clean, `testSessionDuplicate` passes.

umputun (approved):
re-review of 4d5b628 - all round-1 points addressed, no new issues. Good to go.

- e2e: `testSessionDuplicate` exercises the `session.duplicate` ControlServer arm (resolver, focus branch, id echo) and checks insert-after / cwd carry / active off `tree`.
- split-cwd read-back wording qualified across `control-api.md`, `sidebar.md`, the agent skill, `commands.html`, and the `duplicateSession` doc comment.
- `focusedCwd`-over-`effectiveCwd` seed pinned by `duplicateSessionSeedsFromFocusedSplitPane` (fails if swapped to `effectiveCwd`).
- `AppActions.duplicateSession(_:in:)` dropped to `Void`.
- `BuiltinAction`/`PaletteCommand`/menu/`duplicateActiveSession` wiring all in and routed.

one optional nit, not blocking: `testSessionDuplicate` seeds only the source, so its order assertion alone can't tell insert-after from append - the position is really proven by the three-session `duplicateSessionInsertsAfterSourceInSameWorkspaceAndSelects`. A trailing sibling in the e2e would make it self-contained. Your call.

PR #246, opened 2026-07-18
feat: expose terminal surface to accessibility as an editable text area

pbldbl:
## Problem

Voice dictation tools don't work over the terminal. MacWhisper's hold-to-dictate widget never appears in agterm (it does in NSTextView/webview-based terminals), and the same applies to other AX-driven assistive tools.

Root cause: the Metal-backed `GhosttySurfaceView` is absent from the accessibility tree. Dictation/assistive tools probe `AXFocusedUIElement` for a focused *text field* before engaging; over agterm they find no focused element at all (verified: querying the focused UI element of the agterm process returns nothing), so they never show their input UI.

It is **not** Secure Keyboard Entry — there is no `EnableSecureEventInput` call in agterm or the bundled ghostty, and no secure input was active during testing.

## Fix

A new `GhosttySurfaceView+Accessibility.swift` reports the interactive surface as the minimal shape of an editable text field:

- `isAccessibilityElement`, role `.textArea`, label "Terminal";
- `isAccessibilityFocused` = key window + first responder — the signal dictation tools use to pick their target;
- an empty, settable `accessibilityValue`: scrollback is deliberately **not** mirrored into AX (that would pull VoiceOver into reading the whole GPU-rendered grid — a much larger undertaking); the element only needs to *look* like a focused editable field for dictation to engage;
- `setAccessibilityValue` routes AX-based insertion into the existing `NSTextInputClient.insertText` path — the same path the keyboard and IME already use, so text reaches libghostty unchanged (tools that synthesize keystrokes already flow through `keyDown`);
- everything is gated on `!viewOnly`, so non-interactive deck/preview surfaces stay out of the a11y tree exactly as before.

Additive, single-file change; no existing code touched.

## Testing

- `make build` (Debug) and `make release` build clean; `swift test` green (1570 tests).
- Manually verified on macOS 26 (arm64): with this change MacWhisper's hold-to-dictate widget appears over the terminal and dictated text lands at the prompt; before the change the widget never appeared. `viewOnly` dashboard cells remain out of the AX tree.


umputun (commented):
the dictation gap is real and worth closing. Before getting into the code, the bigger thing to settle is the *approach*, because the current shape has a cost beyond the dictation win.

**the approach question**

reporting the surface as an editable `.textArea` with a permanently-empty value is a trade for one class of AX client (dictation) at the expense of another. Before this change the Metal surface is absent from the AX tree, so VoiceOver and other AX clients skip it. After it, they land on a focusable "Terminal" text area that reads back empty while the screen holds content, which misrepresents the terminal to screen readers and to AX-driven automation.

upstream ghostty went the other way for 1.2: a *read-only* AX integration that mirrors real content for screen readers and AI tools, gated behind the accessibility permission so it is opt-in. That read-only model does not solve dictation (their voice-control dictation issue is still open), so "just take upstream's" is not the answer here. But two of its choices are worth deciding on before merging:

- **opt-in vs always-on.** this exposes every interactive surface unconditionally. Upstream put its AX behind a permission for a reason (cost, and not touching non-AT users). Worth deciding whether this should sit behind a setting.
- **empty-editable vs accurate.** the empty-value shape is what makes VoiceOver worse than before. Accept that as a known trade-off for the dictation use case, or gate the whole thing so non-dictation AX clients aren't affected.

my read: decide the shape (unconditional empty-editable shim vs. an opt-in/gated version) before polishing the code, since it changes how much of the below matters.

the two issues below are real regardless of which approach wins.

**major - background sessions become writable AX targets**

`isAccessibilityElement()` (`GhosttySurfaceView+Accessibility.swift:29`) and the setter (`:59`) gate only on `!viewOnly`. But `viewOnly` is true only for dashboard cells - every ordinary deck pane, both panes of a split, scratch, and overlay surfaces are `viewOnly == false`, and the eager deck mounts every session's surface at once. So a window ends up with one editable "Terminal" `.textArea` per realized surface, all overlapping, all but one invisible. Two consequences:

- VoiceOver navigation surfaces N identical empty "Terminal" text areas over the visible one.
- `setAccessibilityValue` is not focus-gated. AppKit does not gate the setter by focus, so an AX client that enumerates the window's text areas, or caches the focused element across a session switch, can inject dictated text into a *background* session's pty. Plain text goes to that shell's prompt; with an embedded newline it can run there (see the next point).

dictation via `AXFocusedUIElement` is safe because `isAccessibilityFocused()` (`:37`) is properly focus-gated, which is why MacWhisper works. The setter and the element advertisement just need the same scoping. The codebase already owns the right gate: `deckVisible` is the "only the on-screen pane" flag used for drag registration and cursor writes. Gate `isAccessibilityElement()` (ideally the whole extension) on `deckVisible`, and require live focus (`window?.isKeyWindow == true && window?.firstResponder === self`, same predicate as `isAccessibilityFocused`) inside `setAccessibilityValue` before inserting.

**minor - the settable-value contract is not honored**

the value is advertised settable (`:67`) but `accessibilityValue()` (`:44`) always returns `""` and `setAccessibilityValue` (`:59`) *appends* via `insertText` rather than replacing, and drops an empty set. So a client that sets the full value on each streaming update, reads back to verify (sees `""`, concludes failure, retries), or does select-all-replace duplicates or garbles text; an empty set can't clear. MacWhisper inserts incrementally so it dodges all of this, but the contract as advertised is a lie to any replace-semantics client.

multiline is the sharper edge: a set runs outside `keyDown`, so `insertText` takes the `currentKeyEvent == nil` branch and sends the whole string (newlines included) as one `ghostty_input_key_s`, with neither the newline->Return split that `inject()` does nor the bracketed-paste wrap that `insertPasted()` does. `inject` splits `\n` deliberately, so this raw path is untested for embedded newlines and could submit commands. Either don't advertise the value settable until replacement is honored (or emit only the actual edit delta), and route a multiline set through the `inject`/paste semantics instead of the raw key-event path.

**tests**

this is XCUITest-observable, unlike the cursor/color cases that are verified by eye - the AX role, label, focus, and the `viewOnly`/`deckVisible` exclusion can all be asserted. A test that the visible focused pane is a `.textArea` labelled "Terminal" and that a background/dashboard surface is not an AX element would lock the gating down and satisfy the new-functionality-needs-tests rule.

**nits**

- the header says the setter uses "the very same path the physical keyboard and IME already use" (`:18`), which isn't exact for a bulk/multiline set, see the multiline point above.
- "adapted for agterm" (`:1`) - adapted from where? Worth citing the source.

pbldbl:
Thanks for the thorough review — this was exactly the right framing. Pushed c3ab8af addressing it.

## Shape decided: always-on, but gated to the on-screen focused pane

I went with the unconditional shim rather than a setting, but fixed the scoping so the VoiceOver-regression cost you flagged mostly goes away: the whole bridge is now gated on `deckVisible` (the existing "on-screen pane, not split-focus-gated" flag drag/cursor already use), not just `!viewOnly`. So only the visible pane(s) are ever an AX text area — the eagerly-realized background sessions fall back to `super` and stay out of the tree exactly as before. A VoiceOver user lands on at most the visible pane, not N overlapping empty fields.

## The two issues

**major — background writable targets.** Closed by the `deckVisible` gate above, plus `setAccessibilityValue` now requires live focus (`isKeyWindow && firstResponder === self`, the predicate you suggested), so an AX client that enumerated text areas or cached a focused element can't inject into a background pane's pty.

**minor — settable-value contract + multiline.** Single-line sets take the keyboard/IME `insertText` path; a multiline set now routes through `insertPasted` (bracketed paste) instead of the raw key path, so embedded newlines land as literal text rather than typing Return — and I clear any in-flight IME preedit first. I documented the value as **append-at-cursor** (a terminal has no addressable value to replace) rather than pretending replace semantics work.

**tests.** Added `AccessibilityDictationUITests`: the visible focused pane is a `.textArea` labelled "Terminal" reading back empty; background sessions and view-only dashboard cells are **not** exposed; a visible split exposes both panes. The count assertions realize every surface first and then poll-then-hold, so the gating is what's under test, not realization timing.

**nits.** Fixed the "same path as the keyboard" wording (not exact for the paste branch) and the provenance line (original to agterm; noted the divergence from upstream 1.2's read-only AX).

## Three things I'd like your call on

1. **`isKeyWindow` in the setter.** I kept the exact predicate you asked for. One trade-off worth naming: a dictation tool whose own panel becomes the *key window* would have its AXValue writes dropped (agterm no longer key). MacWhisper's hold-to-dictate HUD is non-activating so agterm stays key and it works, but if you'd rather be defensive I can relax the write guard to `firstResponder === self` alone (still blocks background panes, since only the visible focused pane is first responder).

2. **Split exposes two "Terminal" text areas.** Because `deckVisible` isn't split-focus-gated, both visible split panes advertise as editable; only the first-responder one accepts a write. Focus-anchoring clients (MacWhisper/Dictation via `AXFocusedUIElement`) target correctly; a label-anchoring client could pick the unfocused pane and have its write silently dropped. I documented it as a known limitation and pinned the two-element contract with a test — happy to narrow to the focused pane instead if you prefer.

3. **Multiline no-submit.** `insertPasted` carries the same mode-2004 caveat as ⌘V and drop (a raw prompt with bracketed paste off still submits a trailing newline). Consistent with the existing paste behavior, so I left it — flagging so it's a conscious choice.

One caveat on verification: `make build` and `swift test` (1570) are green and the XCUITests compile, but I wasn't able to run the XCUITest target locally (automation-permission sandbox on my machine), so the new UI tests still want a real macOS run before you rely on them.

umputun (commented):
the write-path gating from round 1 holds up, I traced every `TerminalView` construction site and the setter really can only reach the focused pane of the key window. One real gap in the multiline fix though, plus a batch of comment/doc accuracy items.

**major - CRLF bypasses the multiline branch**

`GhosttySurfaceView+Accessibility.swift:110` - `text.contains("\n") || text.contains("\r")` never matches CRLF, because `"\r\n"` is a single grapheme cluster equal to neither. So CRLF text takes the `insertText` branch, the raw CR reaches the pty, ICRNL maps it to NL and the shell runs the line. This path isn't gated by mode 2004 at all, so unlike ⌘V it submits unconditionally, and a single line with a trailing `\r\n` is enough to trigger it.
Fix: `text.contains(where: \.isNewline)`, or test `unicodeScalars`. Same normalization `KeystrokeSegments.swift:11-13` and `WorkspaceSidebar+DragDrop.swift:316` already do.

**minor**

`GhosttySurfaceView+Accessibility.swift:119-122` - AXValueSettable is advertised on `axExposed` alone while the setter also needs key-window focus, so an unfocused exposed pane claims writability and then drops the write with no error. Fix: document the semantics, or gate the advertisement on the first-responder term only. Not on `isKeyWindow`, that would flip settable to NO whenever agterm isn't key and could break a client that probes settability at discovery time.

`GhosttySurfaceView+Accessibility.swift:104-114` - `unmarkText()` clears the preedit but never calls `inputContext?.discardMarkedText()`, so a CJK composition survives an AX insert and re-commits on the next keystroke. Fix: `discardMarkedText()` on both branches.

`GhosttySurfaceView+Accessibility.swift:7,12-13,31` - the summary lines say exposure is focus-gated. It isn't, the file says as much itself at lines 44-51, and `testVisibleSplitExposesBothPanes` asserts 2. Fix: "the on-screen pane(s)", with only the write path focus-gated.

`GhosttySurfaceView+Accessibility.swift:75,105` - the live-focus predicate is copied twice, `GhosttySurfaceView.liveFocus` already owns it. Fix: drop `private` from it and reuse, same way `currentTrackingArea` is internal so `+Tracking.swift` can reach it.

`GhosttySurfaceView+Accessibility.swift:61-62` - the doc says everything below falls through to `super`, but `isAccessibilityElement()` returns a hard `false`. Fix: reword it. Changing the code there would be a behavior change, not a comment fix.

`GhosttySurfaceView+Accessibility.swift:95` - "(the shape the maintainer asked for)" states provenance instead of the rule. Fix: drop the parenthetical.

`AccessibilityDictationUITests.swift:70-89` - the dashboard case can't fail on the `!viewOnly` half. No host mounts a viewOnly surface with `deckVisible: true`, so deleting that term keeps all four tests green. Fix: say it pins the `deckVisible` exclusion.

`AccessibilityDictationUITests.swift:173-178` - `realizeAllSurfaces` is described as bringing surfaces into the AX tree, but the deck mounts every session eagerly and `axExposed` is a live read, not a latch. The poll-then-hold is what keeps the assertions non-vacuous. Fix: re-attribute it, or drop the helper.

`WindowContentView.swift:694` - the comment says the Metal-backed `QuickTerminalPane` is not in the a11y tree. That's false as of this PR: `QuickTerminalPane` never assigns `deckVisible`/`viewOnly`, so the defaults expose it. Behavior is right, the comment isn't. Fix: correct the comment. Don't set `deckVisible = false` there, that would drop dictation in the quick terminal.

`README.md` - dictation is a user-visible capability shipping with no user-facing docs. Fix: one line, dictation tools that probe for a focused text field now engage over the terminal, only the on-screen pane(s) are exposed, scrollback is not mirrored to screen readers. `site/docs.html` is mine to mirror, skip it.

**your three questions**

1. keep `isKeyWindow`. AppKit's first responder is per-window and doesn't resign when a window loses key, so `firstResponder === self` alone would make every background window's focused pane writable again, which is the round-1 issue at window granularity. It also costs nothing for the use case: when agterm isn't active `NSApp.keyWindow` is nil, so `AXFocusedUIElement` returns nothing and the tool never anchors in the first place.
2. keep both split panes exposed. Narrowing to the focused pane hides the other from screen readers, and the write is focus-gated regardless.
3. agreed, leave it. Same residual ⌘V and drop already carry, and closing it belongs to the separate unsafe-paste work. The CRLF case above is the part that's actually new.

ran the four UI tests here, all pass in ~34s.

pbldbl:
Thanks — and for running the UI tests. Pushed 4431d06 with all of it.

**Code**

- **AXValueSettable** now gated on the first-responder term only (not `liveFocus`): an exposed-but-unfocused split pane no longer advertises writable and then drops the write, and a discovery-time settability probe made before agterm is key still sees YES.
- **CJK/IME** — `discardMarkedText()` on both write branches, so a live composition can't survive the AX insert and re-commit on the next keystroke. This was the genuinely new one; good catch.
- Reused `GhosttySurfaceView.liveFocus` (dropped `private`, like `currentTrackingArea`) for `isAccessibilityFocused` and the write guard instead of the two copied predicates.

**Comments / docs**

- Reworded the summary lines: exposure is on-screen-pane(s), only the write path is focus-gated.
- `axExposed == false` returns a hard `false` from `isAccessibilityElement` — fixed the "falls through to super" wording.
- Dropped the maintainer-provenance parenthetical.
- Corrected the `QuickTerminalPane` a11y comment in `WindowContentView`: it *is* in the tree now (default `deckVisible`/`viewOnly`), which is intended — dictation works in the quick terminal — so I noted "don't set `deckVisible = false` here" rather than changing behavior.
- `README` gained a short Accessibility section (left `site/docs.html` to you).

**Tests**

- Dropped `realizeAllSurfaces` — you're right that the deck mounts eagerly and `axExposed` is a live read, so the poll-then-hold is what keeps the assertions non-vacuous; re-attributed the comments accordingly.
- Re-attributed the dashboard test to the `deckVisible` exclusion (the `!viewOnly` half can't fail there).

**Your three answers** — all taken as-is: `isKeyWindow` stays in the write guard (the per-window-first-responder point is convincing), both split panes stay exposed, and the multiline mode-2004 caveat is left as the same residual ⌘V/drop carry.

`make build` + `swift test` (1570) green; the four UI tests still compile — I still can't run the XCUITest target locally (automation-permission sandbox), so thanks for exercising them on your side.

pbldbl:
Ran another pass over the round-2 edits. Pushed eb01c22.

**Fixed**

- **Authoritative settable gate.** `isAccessibilitySelectorAllowed` now returns the setter's answer directly (`axExposed && firstResponder === self`, an explicit `false` off the focused pane) instead of delegating to `super`. Since the class implements both `accessibilityValue()` and `setAccessibilityValue(_:)`, AppKit's default could report the getter/setter pair settable and leak `AXValueSettable = YES` onto the unfocused split pane — which would defeat the round-2 gate. Now it can't.
- **Header comment** corrected: the settable advertisement is focus-gated too, not just the write path (it contradicted the body doc and invited a "simplify back to axExposed-only" regression).

**Two things I want to flag rather than change unilaterally**

1. **Settable vs. key-window asymmetry.** You asked for the settable advertisement to be first-responder-only (not `isKeyWindow`), so a discovery-time probe isn't mis-cached when agterm isn't key. The consequence: with two windows open, the focused pane of the *visible-but-not-key* window advertises settable, but its write is dropped by the `liveFocus` guard (that window isn't key). It's harmless for a focus-anchoring client (`AXFocusedUIElement` is nil when agterm isn't key, so it can't reach that pane) — only a role/label-enumerating client hits it, the same class as the split case we already accepted. I kept your predicate and documented the asymmetry. If you'd rather kill it, gating the advertisement on `liveFocus` too closes it, at the cost of the discovery-probe case you called out — your call.

2. **`discardMarkedText()` on both branches** is in as you asked. One reviewer note worth a real-IME check: for some input methods `discardMarkedText()` *commits* the in-flight conversion rather than dropping it, and since the AX path runs with `currentKeyEvent == nil`, a synchronous `insertText` callback would send the half-composed text into the pty. I believe `discardMarkedText` is the canonical discard (not commit) path, but I can't exercise a CJK IME here — worth a quick manual check on your side if you have one set up.

**One to confirm, not fix**

The `testBackgroundSessionsAreNotAccessibleTargets` guard now relies on the settled count being 3 under the un-fixed (`!viewOnly`-only) code. That holds only if an eagerly-mounted but never-focused background surface actually shows up in XCUITest's `app.textViews` (which matches your "eager mount + live read" point). Since you have the runner set up: if you temporarily revert the gate to `!viewOnly` alone and that test goes red, the guard is proven non-vacuous. I couldn't run that red-check locally (automation-permission sandbox).

`site/docs.html` left to you as you said. `make build` green.

umputun (commented):
the round-3 items are in. One thing from round 2 is not, and it was reported as done.

**critical - CRLF still bypasses the paste branch (the round-2 major, never fixed)**

`GhosttySurfaceView+Accessibility.swift:121` - `text.contains("\n") || text.contains("\r")` is byte-identical to the original commit. `git show 4431d06` applied every other round-2 item and left this line alone, and the reply said "Pushed 4431d06 with all of it" while addressing only the mode-2004 caveat, which round 2 had listed as a separate point.

Measured:

```
"ls -la\r\n".contains("\n")  ->  false
"ls -la\r\n".contains("\r")  ->  false
"ls -la\r\n".count           ->  7
```

`"\r\n"` is a single grapheme cluster equal to neither, so a CRLF payload takes the `insertText` branch. The AX call runs outside `keyDown`, so `currentKeyEvent` is nil and `+Input.swift`'s `insertText` sends the whole string as `ke.text` through `ghostty_surface_key`. `insertPasted` is never entered, so there is no bracketed-paste wrap at all: the raw CR reaches the pty, ICRNL maps it to NL, the line runs.

This is not the documented mode-2004 caveat. Mode 2004 governs the paste path and the paste path is skipped, so the submit is unconditional, which is not how ⌘V and drop behave. One line with a trailing `\r\n` is enough, and CRLF is exactly what a browser `<textarea>` value carries per the HTML spec.

Three places assert the property the code doesn't have: the header at :28-30, the setter doc at :108-111, and the README paragraph. Same defect, not separate items.

Fix as in round 2: `text.contains(where: \.isNewline)`, which covers LF, CR, CRLF, VT, FF, NEL, U+2028 and U+2029.

**minor - the routing predicate has no test, which is why this survived two pushes**

`AccessibilityDictationUITests.swift:13` says the write path isn't reachable from XCUITest. That holds for the AppKit halves, but the broken part is a pure-Swift predicate over a `String`, and that belongs in `agtermCore` under `swift test`. `KeystrokeSegments.split` is the same shape, already host-free, and its tests already cover `"one\r\ntwo"`. Four UI tests green and `swift test` green at 1570, while a reported major in the pty write path went through two follow-up commits untouched.

Extract the newline decision into a host-free helper (or reuse `KeystrokeSegments`) and cover `\n`, `\r`, `\r\n` and the no-newline case.

**minor - `axExposed` doesn't consult window visibility**

`deckVisible` comes only from model state: `TerminalView.swift:32`, assigned at `WindowContentView.swift:408/425/441/450/482/542` from `deckInteractive && isActive && ...`. Nothing there reads `NSWindow.isVisible` or miniaturization, so minimizing a window leaves its pane `deckVisible == true` and still reporting role `.textArea` with label "Terminal", and AppKit keeps minimized windows in the AX tree. The write path is unaffected since `liveFocus` needs key, so this is exposure only - but the comment above `axExposed` says "True only for the pane that is actually on screen", and that isn't the case. Either add a visibility term to the gate or narrow the comment, your pick.

**settable vs key-window asymmetry - keep it as it is**

flagging it rather than changing it unilaterally was the right call. Keep the first-responder-only advertisement: a discovery-time probe seeing NO because agterm isn't key is the worse of the two failures, and the enumerating-client case is the same class as the split asymmetry already accepted. Nothing to change here, the doc you added covers it.

pbldbl:
Round 4 — pushed 1e63335.

**CRLF (critical) — confirmed and fixed.** You are right, and my round-2 reply was wrong: `git log -L 121,121` shows the line untouched since the original commit, and the behaviour reproduces exactly as measured — `"ls -la\r\n".contains("\n")` and `.contains("\r")` are both false, count 7, because CRLF is one grapheme cluster equal to neither. So the payload took `insertText` with no bracketed-paste wrap at all, the raw CR reached the pty, and ICRNL ran the line. Not the mode-2004 caveat, agreed.

The decision now lives in `agtermCore/Sources/agtermCore/AccessibilityInsert.swift` as `needsPasteRouting(_:)` using `text.contains(where: \.isNewline)`, and `setAccessibilityValue` calls it. I gave it its own helper rather than reusing `KeystrokeSegments`: `split` normalizes and segments for the newline→Return path, which is the opposite semantics from the one this needs (`session.type` WANTS Return; the paste path must not) — reusing it would have meant asking "does the segment list contain a `.returnKey`", which couples the AX router to a type-injection detail. The doc on the new helper spells out why the naive pair fails, so it does not come back.

Your second point is the reason it survived two pushes, so: `AccessibilityInsertTests` covers no-newline, `\n`, `\r`, `\r\n` (asserting the two `contains` false as the regression anchor), and VT/FF/NEL/U+2028/U+2029. `swift test` 1575 tests, was 1570. Updated the header at the UI-test file to say the routing decision is host-free and unit-tested, and that only the two AppKit insert calls rest on the manual check.

**`axExposed` visibility (minor) — added the gate**, `!viewOnly && deckVisible && window?.isVisible == true`. `NSWindow.isVisible` is false exactly for miniaturized/hidden/not-yet-ordered-in and true for merely occluded or off-Space, which is the semantics the comment claimed. The comment now records why the term is needed (`deckVisible` is pure model state) and that this is exposure only. All four `AccessibilityDictationUITests` still pass with it in.

**Doc claims:** fixed the header at :28-30 and the setter doc at :108-111, both of which now name the CRLF case explicitly. The README paragraph does not need a change — it says only that dictated text lands at the prompt and that exposure is limited to the on-screen pane(s); it makes no no-submit/newline claim.

**settable vs key-window:** left as is.

`swift test` green (1575), Debug build green, the four a11y UI tests green. `swiftlint` is not installed on this machine, so lint is unverified locally — CI will cover it.

umputun (commented):
round-3 items check out. `needsPasteRouting` is real in the tree this time, `AccessibilityInsertTests` would have failed against the old predicate, and the `window?.isVisible` term is in. Your `discardMarkedText` question is answered at the end.

one new problem, same class as the CRLF one and the same root cause: the routing predicate is narrower than the guarantee built on top of it.

**major - control characters bypass the paste path and can submit a command**

`GhosttySurfaceView+Accessibility.swift:137` and `AccessibilityInsert.swift:19` - `needsPasteRouting` tests `Character.isNewline` only, so a payload with no line break takes `insertText`. That goes to `ghostty_surface_key` with `ke.text` set, and libghostty's legacy key encoder ends at `writer.writeAll(utf8)`, so the bytes reach the pty unfiltered.

`insertPasted` is the branch that sanitizes: `ghostty_surface_text` to `completeClipboardPaste` to `input.paste.encode`, whose strip list is `0x00 0x08 0x05 0x04 0x1B 0x7F 0x03 0x1C 0x15 0x1A 0x11 0x13 0x17 0x16 0x12 0x0F`, each replaced with a space before the wrap. Upstream applies that regardless of bracketed-paste mode, and its own comment says why: "a security measure to prevent pastes from containing bytes that could be used to inject commands", for "any text insertion method e.g. a paste, drag and drop, etc." The AX write is the one insertion path in the app that skips it.

measured against a bare shell, not inferred:

```
0x0F written to `zsh -f -i` with `expr 6 \* 7` at the prompt  ->  42 on stdout
```

zle binds `^O` to `accept-line-and-down-history`, so the line runs. Bash maps `0x0F` to the tty's VDISCARD instead and blackholes output until it is re-sent. `0x03` interrupts the running foreground command, `0x1B` feeds escape sequences to whatever is reading stdin.

this is not the mode-2004 caveat. That one governs the paste path, and here the paste path is never entered, so the submit does not depend on 2004 at all. The docstring at :126-128 names 2004 as the whole of the no-submit caveat and the README paragraph rests on the same claim, so both are currently stricter than the code.

Fix: `text.contains(where: \.isNewline) || text.unicodeScalars.contains { $0.value < 0x20 || $0.value == 0x7F }`. Keep the `isNewline` term, it also covers NEL/LS/PS above C0. `0x09` is not in the strip list, so tab-carrying dictated text still survives the paste branch intact. Add `"x\u{0F}"`, `"x\u{03}"` and `"x\u{1B}"` to `AccessibilityInsertTests`, and widen the caveat sentence.

**minor**

`AccessibilityDictationUITests.swift:73-75` - the comment says the open-dashboard assertion pins the `deckVisible` exclusion. It passes with either term deleted: all three sessions go to `openDashboard`, so every surface is a cell at `viewOnly: true, deckVisible: false` and the count is 0 either way. The post-close assertion at :84-85 is the leg that actually discriminates. 4431d06 replaced accurate wording with this one. Fix: correct the sentence. `openDashboard(members: Array(ids.prefix(2)))` would make the open leg discriminate too, worth doing but not required since the post-close leg already carries the coverage.

`GhosttySurfaceView.swift:207-224` - `deckVisible`'s docstring still lists drag registration and pointer tracking as its consumers, and `axExposed` is a third one now. You added exactly that note to `liveFocus` in the same push, so the convention is yours. Fix: one line naming `+Accessibility.swift`.

`AccessibilityDictationUITests.swift:13-14` - "there is no public API to call `setAccessibilityValue` on another process's element" is wrong. `AXUIElementSetAttributeValue` is public in ApplicationServices (`AXUIElement.h:223`), and it is how MacWhisper performs the write this bridge exists to accept. The conclusion holds, the reason does not, and the reason is what someone reads before deciding the write path cannot be covered. Fix: the blocker is that the runner needs AX trust, not a missing API. The `hasFocus` claim two lines down is correct, that one stands.

**rebase needed**

the PR conflicts with master in `README.md`, `GhosttySurfaceView.swift` and `WindowContentView.swift`, and it sits 85 commits behind. That is also why there is no CI on `1e63335`: the build workflow runs on `pull_request`, a conflicted PR has no merge ref to build, so nothing has run since 2026-07-27. Rebase and CI will pick it up.

I ran the rest here in the meantime. `swift test` 1575 pass, `swiftlint --strict` clean, Debug build clean, UI test target compiles. So the lint you could not run locally is fine.

**discardMarkedText**

bounded, and I would not hold the PR for it. If some input method commits instead of discarding, the re-entry is `insertText` with `currentKeyEvent == nil`, but a marked-text composition cannot hold a line break, so no Return is typed and no keybind fires. What is left is leftover composed characters arriving ahead of the dictated text, which ordinary IME typing already does.

umputun (commented):
nothing blocking left. The round-2/3/4 items hold up in the tree, the routing predicate covers what the five doc surfaces now claim, and `swift test` (1576), `swiftlint --strict` and a Debug build are all clean here.

I will take the rebase, no need to do it on your side. The branch is 145 commits behind with conflicts in three files, which is why nothing has built since 2026-07-27.

PR #248, opened 2026-07-18
feat: add opt-in hover close button on sidebar session rows

0717376:
Implements the hover-revealed close button proposed in discussion #247 — **opt-in and off by default**, per the feedback there.

demo

- Hovering a sidebar session row draws a quiet gray wash over the row and reveals a small "×" in the indentation gutter at the selection pill's left edge; the terminal icon and the trailing status/badge chrome stay exactly where they are.
- Hovering the "×" itself on the selected row shows a circular highlight (the Ghostty-tab affordance).
- Clicking closes only that session, through the same path as the right-click "Close Session" — the close-confirmation setting and the undo grace are honored, and a background window's "×" targets its own session.
- **Opt-in**: a "Close button on hover" toggle in Settings → Interface (Sidebar section), default OFF. `AppSettings.sessionCloseButtonEnabled` follows the `attentionButtonEnabled` pattern (chrome mirror on `GhosttyApp`), but is read live at hover time, so a flip needs no re-render pass.
- No control-API changes — pure click-routing to the existing close path (`session.close` already covers it), the same exemption as the workspace "+".

**Tests**: host-free default/round-trip in `AppSettingsTests`; XCUITest `testInlineCloseSessionButtonClosesSession` (AppKit's table accessibility does not expose the row-gutter button as an AX child, so the test drives it by coordinate anchored to the row label, and asserts on the visible row count because the persisted snapshot lags a close by the undo grace). `swift test` (1571), `swiftlint --strict`, the app build, and the neighboring `testInlineAddSessionButtonCreatesSession` are green. README and `site/docs.html` updated.

On the two-hover-buttons-on-opposite-edges concern from the discussion: trailing placement was tried first and collides with the badge/status cluster (revealing it shifts the name), so the gutter slot is the one spot where nothing moves. Happy to revisit the placement — and with the toggle default-off, nothing changes for anyone until they opt in.


umputun:
ran it for a while and it's the call I was worried about in #247. Closing a session is already well covered (⌘W, right-click, the palette), and on a row that already carries the icon, name, split marker, unseen badge and status dot, the hover × plus the wash reads as noise more than help for me.

the placement is the tell. Trailing collides with the badge/status cluster, the gutter puts it opposite the workspace +, and swapping it in over the icon avoids both but doesn't change whether the affordance earns its place. I don't think it does, so I'd rather not carry it even as an opt-in, that's still one more flag and one more code path for something I'd keep off.

the implementation itself is good, opt-in default off, reusing session.close, tested, and it addressed every concern that code could address. This is purely the feel call I flagged going in, and it went the other way for me. Closing it, but thx for building it out so I could actually feel it instead of guessing from the gif.

0717376:
Fair call — and thanks for actually running it before deciding, that's all I could ask for. The process was worth it either way.

PR #258, opened 2026-07-19
docs(agent-skill): correct the foreground check wording (follow-up to #257)

artfaal:
Follow-up to #257, addressing the review nit.

For the wrapped example — `--command "zsh -lc 'claude …'"` — `tree`'s `foreground` reports the **running** process (`claude …` once it's up), not the literal `zsh -lc '…'` argv passed in. It equals the argv only for a bare `--command "ssh host"` with no wrapper.

Rewords the check from *"foreground is your argv"* to *"foreground shows your program running, not a bare shell prompt"*, which is accurate for both the wrapped and bare cases.

Docs only. Thanks for the catch.


umputun (commented):
wording is correct now - `foreground shows your program running, not a bare shell prompt` holds for both the wrapped and the bare `--command` case. Good to merge.

umputun (approved):

PR #259, opened 2026-07-19
feat: highlight the active workspace and session in the sidebar

wievtsal:
## Why

The active session is already highlighted in the sidebar — its row carries the selection pill — but the workspace itself is not: every workspace name renders at the same full-strength theme foreground, so nothing tells you which workspace you're working in at a glance. This is especially needed when all workspaces are collapsed: the selection pill lives on the session row, which is hidden inside its collapsed parent — so there is simply no way to see which workspace is active.

The second important reason is that using the same color palette across all workspaces makes them feel less visually distinct, which can be conceptually confusing and makes it harder to differentiate between them at a glance. Dimming everything but the active context gives the sidebar a visual hierarchy: the active workspace and session at full strength, everything else receding.

## What

- **Non-active rows dim to a secondary color** — the exact tint the leading row icon already uses (theme foreground at 0.6 alpha), so no new color is introduced and it tracks the terminal theme. A workspace row dims unless it owns the active session; a session row dims unless it is the selected one (which also keeps its selection pill, as before).
- **Hovering a dimmed row instantly restores the full color** (no fade/animation), so a dimmed name is still comfortably readable under the pointer. This rides the tracking area that already reveals the workspace-row "+", now installed on session cells too.
- **The highlight anchors on `AppStore.activeWorkspaceID`**, a new host-free derivation: strictly the selected session's owning workspace, nil with no selection. Deliberately *not* `currentWorkspaceID` — its last-workspace fallback is a new-session placement anchor, and highlighting it would light up an arbitrary workspace when nothing is selected.
- The coordinator re-tints workspace rows from `reconcile()` when the active workspace moves, covering the same-row selection case where the selection delegate never re-fires.

## Testing

- New unit test `activeWorkspaceFollowsSelectionWithNoFallback` pins the no-fallback distinction vs `currentWorkspaceID`; full `agtermCore` suite green (1594 tests, 66 suites).
- `make build` and `make lint --strict` clean.
- Verified manually in an isolated dev instance: dimming follows selection across workspaces; hover un-dims with no animation.

Control-API note: pure visual chrome with nothing to drive — exempt from the control-coverage convention.

### DIFF
--- a/agterm/Views/SidebarRowViews.swift
+++ b/agterm/Views/SidebarRowViews.swift
@@ -27,6 +27,26 @@ final class SidebarCellView: NSTableCellView {
     private static let addButtonWidth: CGFloat = 16
     private var hoverTrackingArea: NSTrackingArea?
 
+    /// True for a row whose name dims to the same tint as the leading icon so the active row reads
+    /// at full strength: a workspace cell whose workspace is NOT the active one (the active session's
+    /// owner), or any session cell (a non-selected session dims; the selected one is exempted by
+    /// `setColors`' selected check). Set by the cell builder; the coordinator refreshes workspace
+    /// cells when the active workspace changes.
+    var isSecondary = false
+
+    /// Whether the pointer is inside the cell (the same tracking area that reveals the workspace
+    /// "+"). While hovered, a secondary row's name shows the full default color — an instant swap,
+    /// no animation. Reset via `resetHover` when a reused cell is reconfigured.
+    private var isHovered = false
+
+    /// The selection state last applied by `setColors`, so the hover re-tint can re-run it without
+    /// re-deriving the row's selection.
+    private var lastSelected = false
+
+    /// Clears transient hover state on cell reuse (the old row's `mouseExited` may never fire for
+    /// the recycled cell).
+    func resetHover() { isHovered = false }
+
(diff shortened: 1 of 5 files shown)


umputun:
this is a design call, not a quality one. The code is clean, the activeWorkspaceID hoist is host-free and tested and it's all green.

the reason I'm passing: dimming every other row to make the active one pop imposes a foreground/background hierarchy the model doesn't have. Workspaces are peers, the active one is just where the current session sits, not a more important context. And it's always-on for everyone with no toggle, which is a harder version of the call I made on #248 a couple of days ago. I closed an opt-in, default-off affordance there for adding clutter, so a permanent re-tint of every name isn't something I want to carry either. The session-row dimming also doubles up on the selection pill that already marks the active session, and it lowers the contrast on exactly the sibling names you scan to switch to.

the one case with real merit is all workspaces collapsed, where nothing shows which one owns the active session. But that state is easy to avoid and easy to leave: selecting a session auto-expands its workspace, and Collapse Workspaces keeps the active one open, so you only land in the blind spot by deliberately collapsing the active workspace, and a click on its disclosure triangle brings it back. If that ever feels worth solving it's a small positive marker on that one active row, behind an Interface toggle, not a dim of the whole sidebar.

closing it. Thx for the clean implementation and the clear writeup, it's just not a direction I want for the default sidebar.

PR #265, opened 2026-07-21
feat: per-workspace root directory

jokius:
Give a workspace an optional root directory. When set, every new session created in that workspace opens there — a hard override of the global `newSessionDirectory` setting; unset falls through to that setting, and a non-existent root falls back at spawn time. A workspace usually maps to a project, so "new session here" lands in the project directory without a global setting change (today it opens `$HOME` once the workspace has no session to inherit a cwd from).

**Model.** `Workspace.root` and `WorkspaceSnapshot.root` are Optional with no snapshot version bump — a legacy `workspaces.json` decodes with `root == nil`.

**Resolution.** `AppSettings.resolveNewSessionCwd(workspaceRoot:…)` applies the override; both new-session entry points honor it — `AppActions.resolvedNewSessionCwd(inWorkspace:)` (with an on-disk existence check so a stale/deleted root falls through instead of failing the spawn) and the control `makeSessionResponse`.

**Control + UI.** A `workspace.root` command — `ControlActions.setWorkspaceRoot`, `AppStore.setWorkspaceRoot`, the `ControlWorkspaceNode.root` read-back on `tree`, and `agtermctl workspace root <dir|--clear>` — plus workspace-row **Set Root Directory…** / **Clear Root Directory** context items in the sidebar.

**Tests.** `agtermCore` stays green (+7): resolve precedence, `Workspace.root`, snapshot codable + legacy decode, command round-trip, `ControlWorkspaceNode` omit-when-nil, dispatcher routing, and the `controlTree` read-back. The app target and `agtermctl` build, `make lint` is clean, and the bundled agent skill (`SKILL.md`/`reference.md`/`examples.md`) plus the `control-api` rule are updated (command count 61 → 62).

Developed downstream in [rook](https://github.com/jokius/rook) and ported back here (names/paths translated to `agterm`) as a goodwill contribution — no pressure to take it. One thing left for you: the website (`site/commands.html`, `site/docs.html`) still reads "61 commands" and has no `workspace.root` entry — I didn't want to guess at your hand-authored per-command HTML.

### DIFF
--- a/.claude/rules/control-api.md
+++ b/.claude/rules/control-api.md
@@ -237,9 +237,9 @@ paths:
   rules, then remaining targets resolve inside that same store so one command never mutates multiple windows.
   The top-level `target` also carries the first explicit batch target so a new CLI talking to a still-running
   pre-batch server degrades to a named session instead of accidentally acting on `active`.
-- **Command catalog (61 commands):**
+- **Command catalog (62 commands):**
   - `tree`
-  - `workspace.new`/`workspace.rename`/`workspace.delete`/`workspace.select`/`workspace.move`/`workspace.focus`
+  - `workspace.new`/`workspace.rename`/`workspace.delete`/`workspace.select`/`workspace.move`/`workspace.focus`/`workspace.root`
   - `session.new`/`session.duplicate`/`session.close`/`session.select`/`session.rename`/`session.reveal`/`session.move`/`session.type`/`session.split`/`session.scratch`/`session.focus`/`session.resize`/`session.go`/`session.copy`/`session.paste`/`session.selectall`/`session.text`/`session.search`/`session.status`/`session.flag`/`session.seen`/`session.background`/`session.overlay.open`/`session.overlay.close`/`session.overlay.resize`/`session.overlay.result`
   - `surface.zoom`
   - `dashboard`
--- a/agterm/AppActions.swift
+++ b/agterm/AppActions.swift
@@ -117,7 +117,7 @@ final class AppActions {
     func newSession() {
         guard uiActionsEnabled else { return }
         guard let store, let workspaceID = store.currentWorkspaceID,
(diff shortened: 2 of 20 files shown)


umputun:
going to decline this one.

a per-workspace root directory assumes a workspace maps to a directory, and agterm's model deliberately doesn't. A workspace holds the sessions for one project *or context* - plenty of people (me included) don't organize workspaces to mirror a directory tree, so a persisted `root` on every workspace plus the always-on "Set Root Directory…" row items pushes a convention that isn't universal.

the bigger reason is the layer. "open a new session in this directory" is something agterm already expresses several ways, none needing new persisted state:

- `agtermctl session new --cwd <dir> --workspace <w>` - whatever would set a workspace root already knows the directory, so it can pass it at creation. That also makes the new `workspace.root` command redundant with `session.new --cwd` for anything driving the socket.
- Duplicate Session (sidebar row, ⌃P palette, menu, `duplicate_session` bind, `session.duplicate` over the control channel) - a fresh shell in the source session's directory, same workspace, right after it. That's "another session in this project's dir" in one key.
- the `currentSession` new-session-directory mode - a new session inherits the active session's cwd.
- the global `newSessionDirectory` for a fixed default.

the only thing a persisted per-workspace root uniquely reaches is the first session in an empty workspace, which doesn't justify a 62nd command, per-workspace state, and permanent sidebar chrome for something the tool already does four other ways. Same line as the README: a sensible minimum out of the box, anything past it built over the control API instead of shipped as a built-in.

one more thing, separate from the decision: as written it doesn't persist. `AppStore.snapshot()` and `restore(from:)` rebuild each workspace without `root` (they skip the root-aware `workspaceSnapshot(_:)`/`workspace(from:)` helpers), so a set root is dropped on both save and restore and won't survive a restart. The three added tests pass because none exercises the real `snapshot()` to `restore()` round trip. Worth knowing for the rook version even though it doesn't change the call here.

there's no issue or discussion behind it - for a change this size the scope is worth settling before a PR.

PR #284, opened 2026-07-23
feat: add application Dock menu

melonamin:
## Summary

- add an AppKit Dock menu with New Session, Quick Terminal, and Dashboard actions
- include recent sessions and sessions needing attention
- retain dynamic menu targets and keep session actions scoped to the captured window

## Why

Expose common actions and session navigation from the Dock without requiring the app window to be frontmost.

Mostly remove the need in menu bar extra #140

## Screenshot
Screenshot 2026-07-23 at 13 57 24


umputun (commented):
two blocking issues before this is ready:

1. `agterm/AppDelegate+DockMenu.swift:181` - a non-idle session can appear in both Recent Sessions and Sessions Needing Attention, but the Recent path calls only `focusActiveSession()`. It won't reveal a tagged right or scratch pane, so the same session behaves differently depending on which submenu selected it. Call `revealActiveBlockedPane()` for every session selection and remove the `needsAttention` branch. That helper already falls back to ordinary focus for idle sessions.

2. `agterm/AppDelegate+DockMenu.swift:44-185` - this new AppKit UI has no test coverage. Pls add focused Dock-menu UI tests for menu contents and enabled states, MRU/current-session filtering, attention ordering, nil-sender target invocation, and the captured-window case where window B becomes frontmost before a window A session is selected. The dashboard/zoom stale-action checks should be covered too.

minor:

- `agterm/AppDelegate+DockMenu.swift:95` hardcodes the MRU limit as `10`, while the existing recent-session surfaces use `SessionSwitcher.maxCandidates`. Use the shared value here too.
- update `README.md` and `site/docs.html`. The attention docs still say there are "two more ways," and neither public surface mentions the Dock menu or its last-active-window scope.
- update `.claude/rules/menu-actions.md` and `.claude/rules/windows.md` for the new action surface. Add `AppDelegate+DockMenu.swift` to the applicable paths and document retained targets, captured store/window scope, the per-window modal check, and synchronous frontmost-window publication.

umputun (commented):
two blocking issues and one question remain:

1. `project.yml:182-215` - the new `agtermTests` target is application-hosted, but the Test action sets neither `AGTERM_STATE_DIR` nor `AGTERM_CONTROL_SOCKET`. Running the focused test launched agterm against the real config and app-support paths and started login shells before `DockMenuTests.setUp()` created its temporary library. It also killed my live control socket. Pls isolate the host process before `agtermApp.init()` runs. The current CI test job runs only `agtermCore`, so add a CI invocation for this target after it is safe to run.

2. `agterm/AppDelegate+DockMenu.swift:50-77` - should New Session, Quick Terminal, and Dashboard use the window captured when the menu opened, or whichever window is frontmost when the item is selected? The current code uses the latter, while README/site promise that all lists and actions stay tied to the captured window. If invocation-time frontmost is intended, pls narrow the docs to session lists/actions. If captured scope is intended, capture A's window/store and add the B-frontmost case for all three actions.

3. `agterm/AppDelegate+DockMenu.swift:178-180` - pane-aware selection is still wrong for an explicit `left` status or an omitted pane. `revealActiveBlockedPane()` falls through to `focusActiveSession()`, which follows the existing `splitFocused`; if the right pane was focused, the Dock jump stays on the right. Set the primary-pane target explicitly for `.left`/`.none` and add cases with the right pane initially focused.

umputun:
FYI: I have run out of weekly credits/limits for my daily-driver claude code and adopted skill, which is helping me review things, for codex. This is the reason why those reviews sound so categorical and demanding, as if codex writes to other models and not to humans. ;) nothing personal.

melonamin:
> nothing personal

No worries, I'm used to Codex.

Quick question: have you considered including a review skill or a `review-instructions.md` file? There's a lot of churn in reviews because our agents are looking for different things.

umputun (commented):
the three from last round check out. Test-action isolation holds on every route (I ran the target and my state dir, config and socket came back byte-identical), captured window/store scope is right, and `.left`/omitted targets the primary explicitly now. The round-1 items are in too, except the test coverage below.

**`agtermTests/DockMenuTests.swift` - coverage gaps.** the 9 tests cover menu contents, enabled states, MRU order and cap, attention ordering, nil-sender dispatch, and the window-B-frontmost captured scope for all three top-level items plus session selection. Not covered:

- stale dashboard/zoom inertness for the three TOP-LEVEL items. `makeCapturedSessionAction()` only builds a session row, so tests 8 and 9 exercise session selection only. Test 6 opens the dashboard on window B, never on captured window A after the menu is built. Nothing proves New Session / Quick Terminal / Dashboard go inert when A enters a modal post-build, which is the case `windows.md` now documents.
- the `.right` reveal branch. Only `.scratch` and `.left`/nil appear anywhere in the file, so `AppActions+Focus.swift:83-85` never runs. The Dock is the fifth caller of that branch.
- captured window CLOSED (not just modal) between build and invoke. Nothing calls `library.closeWindow`, so the `library.store(for:) === store` identity guard at `AppDelegate+DockMenu.swift:169` and the `weak store` capture are both unexercised.
- `applicationDockMenu` with `library == nil`, i.e. a Dock right-click before the scene `.task` wires it. The `store = library?.activeStore` nil path and the `store == nil` submenu branch are never hit.
- target invalidation across a menu rebuild. No test builds a second menu then invokes an item from the first, which is the one path where a stale item's target is already gone.

separately, `testSessionSelectionForcesPrimaryForExplicitAndOmittedLeftPane` doesn't prove what the name says. `DockMenuTestSurface` isn't a `GhosttySurfaceView`, so `focusSplitPane`'s `as? GhosttySurfaceView` cast never matches and first responder never moves. Both assertions check only that `splitFocused` was cleared, which happens at `AppActions+Focus.swift:90`, before `focusSplitPane` is called at all. Reasonable limit for a windowless hosted test, but rename it or add a comment so it doesn't read as focus coverage.

**question on `revealActiveBlockedPane`.** the fix went into the shared helper, so `.left`/omitted now clears `splitFocused` and forces the primary for every caller: sidebar click, ⌥⌘↑↓, first/last, both palettes, auto-follow. The agent-status hook sends `--pane left` from the primary pane and emits `active` on ordinary tool use, not only on a block. So a session whose agent is merely working, split hidden with the right pane zoomed, now un-zooms and jumps to the left pane on a plain sidebar click. Should the forced primary be gated on `needsAttention` (blocked/completed) rather than any non-idle status, or is pulling to the tagged pane on `active` what you want as well? No strong view here, it's a semantics call.

fyi the branch is 5 commits behind master and conflicting right now.

umputun:
> Quick question: have you considered including a review skill or a `review-instructions.md` file? There's a lot of churn in reviews because our agents are looking for different things.

I'm planning to publish it someday. As of now, it is way too universal and aslo too specific to my needs at the same time. It can do a review of almost anything on GitHub, GitLab, Gitea, for issues, pull requests, discussions, as well as perform local "review-fix-review" loops (multiple levels of both looping and depth), watch pull requests and merge requests progress, and a dozen of other things. It is like 2,000 lines of a crazy py script plus 1000 lines of md. Conceptually, review part is similar to Ralphex (multiple cc agents plus codex, in parallel), but with a lot of tweaks for external contributor's specifics.

umputun (commented):
the round-3 coverage gaps are closed except one. `.right` reveal, captured-window-closed, `library == nil` and rebuild invalidation all have tests that discriminate now, and the pane-selection test names are honest after the rename. `active` keeping the existing pane selection is what I wanted, so the reveal semantics are settled.

one thing I'd fix before merge, then a pile of minors.

**`.claude/rules/ci.md:32-34`** is now false. "CI does NOT run the XCUITests, it builds the app but never test-runs the app target; only the host-free `swift test` runs in CI" - the new `xcodebuild test -scheme agtermTests` step does test-run the app target (`TEST_HOST` is the app binary, `project.yml:193`). The job inventory at :10-16 is stale the same way: it lists `test` as `swift test` then `llvm-cov export` then artifact upload, with none of the four new steps in between, and puts the libghostty cache in `build` only. That file is path-scoped to `.github/workflows/**`, so it's the first thing read before any CI edit.

minor:

- `agtermTests/DockMenuTests.swift:288-301` - `testCapturedTopLevelActionsBecomeInertWhenTerminalZoomBegins` doesn't discriminate. `makeCapturedTopLevelActions()` never opens a second window, so the captured window is `library.activeWindowID` and all four assertions hold with the Dock closures' rechecks deleted: `newSession()` and `toggleQuickTerminal()` stop at their own `guard uiActionsEnabled`, `toggleDashboard()` at `guard !terminalZoomActive`, and :300 asserts a value only the test wrote. The dashboard twin is 1-for-3, :284 is the only assertion that discriminates. Build it the way `testTopLevelActionsStayWithCapturedWindowWhenAnotherBecomesFrontmost` does: open B, build the menu on A, put A into zoom, set frontmost to B, assert `frontmostWindowID` is still B. That fails the moment a recheck goes.
- `agtermTests/DockMenuTests.swift:539-554` - the stale target survives the rebuild only because the weak read at :543 autoreleases it into the test's pool. Delete that `XCTAssertNotNil` and `XCTUnwrap(item.target)` at :636 gets nil. Bind it first, `let staleTarget = try XCTUnwrap(staleItem.target)` before `dockMenu()`, and invoke through that. Keep `isValid`/`invalidate()`, it covers the window where AppKit has already loaded the target for dispatch, which dropping the array doesn't.
- `agterm/AppDelegate+DockMenu.swift:196-209` - the pre-selection capture is Dock-only, so a `completed --auto-reset --pane right` reveals the split from the Dock and from nowhere else. The other six callers read the indicator after `selectSession` already cleared it (`AppStore.swift:332`); that part predates this PR, the split doesn't, and the bundled hooks send exactly this shape on every finished turn. Either hoist the pre-clear capture into `selectSession` so all seven agree, or drop the `captured:` overload.
- `.claude/rules/menu-actions.md:317,319` - "falls through to `focusActiveSession`" isn't true anymore, the branch is `splitFocused = false` plus `focusSplitPane(wantSplit: false)`. Not a rename either: `focusActiveSession` bails on `renamePending` / `palette?.mode != nil` and targets `topmostSurface`, `focusSplitPane` does neither and bumps `focusGeneration`. The godoc on the function was fixed in this push, the rule wasn't.
- `.claude/rules/menu-actions.md:33-35` and `windows.md:29-30` - "stays inert even if AppKit still retains its target" is wrong, `NSMenuItem.target` is `weak`. That's why `dockMenuActionTargets` exists at all, and `AppDelegate.swift:26-29` states it correctly. Say the array is the only strong ref and `invalidate()` covers the already-dispatching case.
- `.claude/rules/menu-actions.md:322-329`, `notifications.md:269-273`, `agent-skill/reference.md:365-370` - the Dock is missing from the reveal-caller lists, and the menu-actions one is explicitly closed ("fires from ALL of") with "All paths route through the one helper so they can't drift" right after it. README, docs.html and SKILL.md all got "a Dock-menu session".
- `site/commands.html:1103` - `session.status --pane` still says the tag makes "any user-initiated GUI selection reveal and focus the tagged pane". Every other surface got the `blocked`/`completed` qualifier.
- `agent-skill/examples.md:516-518` - same miss, the prose above the `active --pane right` example still describes the ungated behavior. The example line itself is fine.
- `.claude/rules/menu-actions.md:27-51` and `windows.md:27-40` - hard-wrapped mid-sentence, CLAUDE.md asks for one sentence per line in these files. 9 of 23 added lines in menu-actions, 5 of 14 in windows, against 1 of 525 across the rest of both files.
- same two blocks - no keep-in-sync exemption callout. Every Dock item drives something the socket already has (`session.new`, `quick`, `dashboard`, `session.select`) so nothing is owed, but CLAUDE.md wants that said out loud. The `open -a agterm /path` bullet in `windows.md` is the shape to copy.
- `agterm/AppDelegate+DockMenu.swift:115-121` - `recentDockSessions` re-implements `WindowContentView.recentSessions` (`WindowContentView+RecentSessions.swift:16-20`) exactly, and its comment claims they match with nothing enforcing it. `AppStore+Recency.swift` already holds `recentSessions(limit:)`, so a `navigableRecentSessions(limit:)` there with both surfaces calling it is about five lines.
- `project.yml:182-197` - `SWIFT_STRICT_CONCURRENCY: minimal` and `ENABLE_HARDENED_RUNTIME: NO` with no comment, while `agtermUITests` fifteen lines up explains both of its identical overrides. One line each is enough.
- `.github/workflows/ci.yml:45-65` - pls fold the hosted run into `build` instead of `test`. `build` already restores that cache and installs xcodegen, and as it stands both jobs use the same key with no `restore-keys`, so any `setup.sh` change builds libghostty twice in parallel and one of the two cache saves collides. It also puts the app build plus the `agtermctl` release build in front of `coverage`.
- `Makefile:35` / `scripts/test.sh` - `make test` is still agtermCore only, so the hosted suite sits outside the documented pre-commit loop. A `make test-app` running the same `xcodebuild test`, plus a line in CLAUDE.md's build/test list.
- `agtermTests/.swiftlint.yml` - missing, so the file is linted at the root 1000/800 rather than the 2000 test budget `agtermUITests/` and `agtermCore/Tests/` get. Not urgent, the type body is 554, and this applies the existing convention rather than changing any limit. The path needs adding to the `swift` paths-filter too.

umputun (commented):
all 16 items from last round are addressed, and pushing the capture down into the store was the better of the two options. `selectSession`/`navigateSession` returning the pre-reset indicator, auto-follow carrying it in the notification, and deleting the no-arg overload so nothing can silently take the old path is a cleaner fix than the one I would have accepted. `navigableRecentSessions(limit:)` behaves identically to both originals it replaces, both test fixes now discriminate (I checked the counterfactuals: drop either recheck, or `invalidate()`, and the assertions fail), and `scripts/test-app.sh` runs clean end to end.

one thing the refactor introduced.

**`agterm/AppActions.swift:456-465`** - attention-nav (⌃⌥↑/⌃⌥↓) no longer reveals the tagged pane when the blocked session is already selected. `attentionTarget` excludes the current session (`AppStore.swift:592`), so `navigateSession(.nextAttention)` returns nil, and `revealActiveBlockedPane(captured: nil)` takes the `guard let indicator else { focusActiveSession(); return }` exit at `AppActions+Focus.swift:75` before `needsAttention` or `statusPane` are ever read. The deleted no-arg overload re-read the live session, so the no-op nav still revealed. True at 7df70fe, still true at fe0cefb, gone at HEAD.

it's reachable in the ordinary flow: an agent sets `blocked --pane right`, I keep typing in the left pane, `clearedBy` doesn't clear a right-tagged block, so the session stays selected and blocked. ⌃⌥↓ then does nothing instead of flipping focus to the waiting pane. Same shape for a scratch tag, and `examples.md:515-529` still documents attention-nav as the thing that takes you to the tagged pane. Nothing catches it either, every case in `PaneAwareStatusUITests.swift` calls `parkOnNewSession()` first, so nav always moves and never returns nil.

fix is `let indicator = store?.navigateSession(.nextAttention) ?? store?.activeSession?.agentIndicator` in both attention actions. Nav returns nil only when no selection happened, so nothing was auto-reset and the live read gives exactly what `selectSession` would have returned. The four plain-nav actions have the same shape on their nil path (empty `navigableSessions`, e.g. flagged mode with nothing flagged) and can take the same `??`. Pls add a case with the sole attention session selected and its tag on a pane that isn't focused.

minor:

- `.swiftlint.yml:4,34` - both header comments still say the 2000-line test budget comes from nested configs in `agtermUITests/` and `agtermCore/Tests/`. CLAUDE.md got the third entry in this same commit, so the two now disagree about a fact one of them exists to record.

umputun (approved):
round-5 fix checks out, and nothing new came in with it. `?? store?.activeSession?.agentIndicator` is exact on both attention actions and the four plain-nav ones (the `??` RHS is an autoclosure, so it can never read a post-`autoReset` value), and `testAttentionNavRevealsTaggedPaneWhenSoleAttentionSessionIsAlreadySelected` discriminates: drop the `??` and both `pollOnScreen` assertions fail.

merging this. Nothing below is blocking, it is all minor and most of it is stuff my earlier rounds walked past rather than anything this push introduced. Writing it down so it is on the record, I will go through the list myself and fix whatever actually needs it.

**code**

- `agterm/AppActions+Focus.swift:93-95` - the `.left`/nil branch writes `splitFocused = false` + `focusSplitPane(wantSplit: false)` unconditionally where master ended in `focusActiveSession()`. That is the documented contract for attention-nav and the Dock row, but the four PLAIN-nav actions get it too: with exactly one navigable session (flagged mode with one flag, or a focus filter on a one-session workspace) `navigateSession` self-wraps, `selectSession` returns the live indicator, and ⌥⌘↓ flips you off the split you were typing in. With the split hidden and `splitFocused` true, `WindowContentView.swift:437` renders the right pane maximized, so it swaps the visible terminal, not just first responder. `clearedBy` is pane-scoped, so a left-tagged block survives typing in the right pane and the state persists.
  Fix is to keep the reveal for attention-nav and skip it for plain nav when `navigateSession` resolved to the session already selected.
- `agterm/AppActions+Focus.swift:130-170` - `focusSplitPane` checks zoom, dashboard and the quick terminal, but not `renamePending` or `palette?.mode != nil`, which `focusActiveSession` re-checks on every one of its 12 retries five lines up. Pre-PR that gap only covered the rare `.right` tag; now a plain `session status blocked` with no `--pane` routes through it, so a row click followed within ~360ms by ⌘R or a palette open keeps stealing first responder. Two early-returns in `focusSplitPane` close it.
- `agterm/AppActions.swift:784-786` - widening `toggleQuickTerminal` from `guard !terminalZoomActive` to `uiActionsEnabled` is a drive-by. The Dock closure already re-checks `uiActionsEnabled(for: windowID)` itself, and the View-menu item carried `.disabled(modalActive)` before this PR, so nothing regresses, but it changes a shared method the Dock did not need. Either it stays with a line in `menu-actions.md` alongside the other `uiActionsEnabled` callers, or it goes.

**comments**

- `agterm/AppActions.swift:47-48` - "`frontmostDashboard?.isOpen` mirrors `terminalZoomActive`, resolved on the frontmost window like the zoom target" describes an expression the body no longer has; it is `uiActionsEnabled(for: library.activeWindowID)` now. It also does not say that `uiActionsEnabled(for: nil)` returns false where the old expression returned true.
- `agterm/AppActions+Focus.swift:45-46` - the caller list omits the Dock, while the paragraph appended at `:71-73` in the same comment does list it.
- `agterm/AppActions.swift:450-455` - the attention godoc reads as a plain "navigateSession then reveal the moved-to session", so the `??` on `:459`/`:465` looks redundant. That is the one place it is doing real work, since `attentionTarget` breaks on `idx == curIndex`. The sole-selected-attention case needs naming there so it does not get tidied away later.

**rules**

- `.claude/rules/sidebar.md:85` - still `AppActions.revealActiveBlockedPane()`, the overload this PR deleted. `menu-actions.md`, `notifications.md`, `control-api.md` and `windows.md` all got `(captured:)`.
- `.claude/rules/sidebar.md:91` - "attention-nav, plain session nav, the command palettes, and idle auto-follow" is missing the Dock row; the mirrored sentence at `menu-actions.md:311` has it.
- `.claude/rules/menu-actions.md:307` - "an empty filtered list preserves ordinary focus behavior" is not true. With an empty filtered list and a still-selected `needsAttention` session the fallback fires and the reveal runs, which is what `AppActions.swift:423-424` actually says.
- `.claude/rules/ci.md` - the rewrite dropped `🚨 Nothing to report`, and it now appears nowhere in the repo. The surviving sentence still says to verify the Coveralls build after changing that path without saying what a silent failure looks like. The cache key and the `$GITHUB_WORKSPACE/` strip also went, though both are readable off `ci.yml`.
- `.claude/rules/ci.md` - worth recording that the `build` job compiles the app twice on purpose: Release for the WMO/link check, Debug because `DockMenuTests` is `@testable import agterm` and `ENABLE_TESTABILITY` is Debug-only. Both are needed, so it is a cost to write down rather than a thing to unify.

**tests**

- `agtermTests/DockMenuTests.swift:65-82` - `testHostProcessStartsWithIsolatedStateSocketAndShellFreeScene` does not cover the shell-free half: delete the `if Self.isHostedUnitTest { Color.clear }` branch and it still passes. The isolation assertions on `:67-81` are solid, so the honest fix is the rename, not a new assertion (the obvious candidate, `controlServer == nil`, rides an async `.task` and would pass in the broken case too).
- `agtermTests/DockMenuTests.swift` - nothing invokes the Dashboard item with `dashboardWasOpen == true`. `testContentsEmptyStatesAndEnabledStates:121-126` gets one line away and stops at the enabled assertion, so the close-toggle both new rules call out is unexercised. Wants the dashboard opened, the menu built, the item invoked, then `isOpen == false`.

melonamin:
> I'm planning to publish it someday. As of now, it is way too universal and aslo too specific to my needs at the same time. It can do a review of almost anything on GitHub, GitLab, Gitea, for issues, pull requests, discussions, as well as perform local "review-fix-review" loops (multiple levels of both looping and depth), watch pull requests and merge requests progress, and a dozen of other things. It is like 2,000 lines of a crazy py script plus 1000 lines of md. Conceptually, review part is similar to Ralphex (multiple cc agents plus codex, in parallel), but with a lot of tweaks for external contributor's specifics.

I'm not talking about the whole machinery, but understand that it may be hard to separate.

For now I'm trying to distill your reviews into a skill:

Screenshot 2026-07-26 at 06 55 36

Otherwise, this becomes frustrating: we're putting "human in the loop" in the wrong place. Not sure what's so special about this specific PR that it took so many iterations.

umputun:
Yep, the loop was a pain, agree. And not just that - it wasn't even completed, but rather terminated, and I took the last report and fixed it in post-pr activity. The question is "what changed" and this is multiple factors answer:

1. I started to use a new skill for review I made, and this skill was supposed to be deeper but at the same time I expected it to converge faster
2. Run out of Cladue code limits and used Codex in the middle of this process
3. Last 2 reviews used Opus5

I don't think any single of these factors is the reason, probably a combination. All this ping-pong made me change the review skill significantly, introducing several mechanisms like materiality detection, the level of obnoxiousness, verifier step to clean crap after detection, and so on. Not sure if any of this even helped, time will tell

Another thing, directly related to your suggestion is this: I have started to work on a small app "revmux". This thing is what skill can call to do review of various kinds and this thing will have a configuration (set of prompts, agents, models and efforts) which can be stored either on user level or project level. The skill to drive it will also be part of this project. The loop we had not the only reason to have this thing instead of just skill, but for sure one of reasons.

PR #289, opened 2026-07-24
feat(agent-status): add OpenCode lifecycle plugin

culler127:
Summary

- Bundle an OpenCode lifecycle plugin (agterm/Resources/agent-status/opencode/) mirroring the Pi extension pattern: session.status busy/retry → active --blink, idle → completed --auto-reset, permission/question prompts and session.error → blocked. The deprecated session.idle event is ignored so it does not double-fire with session.status(type=idle).

- Plugin exports only the plugin function; pure helpers live in a separate agterm-status-logic.mjs so OpenCode's legacy loader (which rejects non-function exports) accepts it. A node --test fixture covers event→args mapping, child-session whitelist, and the sequential report queue.

- Install the plugin from Help ▸ Install Agent Status Hooks… when ~/.config/opencode exists, with the same ownership/degrade-to-warning policy as Pi (OpenCodeResult + isWarning, folded into install() via an InstallOutcome aggregate to stay under the large-tuple lint).

- Drop opencode from the default AGTERM_AGENT_RE (both integration.sh and integration.fish) so the coarse shell active/idle no longer fights the plugin; add a defaultShellAgentRegex host-free constant + tests that pin the shell files to it.

- Keep-in-sync surfaces updated: README.md, site/docs.html, site/llms.txt, site/index.html, docs/troubleshooting.md, agterm/Resources/agent-skill/troubleshooting.md, .claude/rules/control-api.md.
- 

Test plan
 

- [ x ] cd agtermCore && swift test — 1741 tests in 73 suites green (incl. new OpenCode path/ownership, default regex, shell-integration sync tests)
- [ x ]  swiftlint lint --strict — 0 violations in 259 files
- [ x ]  make build (Debug) — BUILD SUCCEEDED
- [ x ]  node --test tests/opencode-plugin/agterm-status.test.mjs — 12/12 (incl. OpenCode loader-compatibility assertion)
- [ x ]  Manual: isolated dev instance (AGTERM_STATE_DIR=/tmp/...), installed the bundled agent-status package + OpenCode plugin, restarted OpenCode inside an agterm session — sidebar glyph cycled active→blocked→completed across a turn


umputun (commented):
two structural asks first, then the specifics.

**collapse the plugin back to a single file.** the `.mjs` split exists so node can import the helpers, and it costs a fair amount: a two-file install with an ordering hazard, a `package.json` that ships to users with no runtime purpose, two extra path constants in `AgentHooksInstall`, and a failure mode where a missing `.mjs` breaks OpenCode at startup instead of no-op'ing. `pi/agterm-status.ts` is one self-contained file and I'd like this to match it. For coverage, `agtermCore/Tests/agtermCoreTests/CodexStatusHookTests.swift` is the pattern here, it reaches `agterm-codex-status.sh` via `#filePath` and runs it against a stub that records argv. The same works for the plugin through the `AGTERM_STATUS_WRAPPER` seam you already support, and it runs under `swift test` and in CI for free.

**narrow the event surface.** the docs describe `session.status` plus the blocked events, but the code also reports on `tool.execute.before`/`after`, `session.compacted`, `chat.message`, plus a child-session set and a five-event child whitelist. Every one of those extra arms maps to the same `active --blink` that `session.status(busy)` already gives you. Only `permission.replied`/`question.replied`/`question.rejected` do real work, they clear `blocked`. Cutting to `session.status` + the three blocked events + the three reply events drops the child set entirely, which removes two of the items below.

blocking:

1. `agterm/AgentHooksInstaller.swift:339-349`, the entry point is written before its dependency. If the `.mjs` write fails you return `.writeFailed` and the alert says the plugin "was skipped", but `agterm-status.js` is already on disk importing a module that isn't there. OpenCode then throws on module resolution at every start, which is worse than not installing at all. Write the dependency first, entry last.

2. `agterm/AgentHooksInstaller.swift:336-346`, only the `.js` goes through `mayOverwriteOpenCodePlugin`. `existingLogic` is read to decide `.alreadyConfigured` and then the write is unconditional, through a symlink if one is planted there. The comment above it says "Same ownership / degrade-to-warning policy as Pi", and `README.md:466` (mirrored in `site/docs.html`) tells users a same-named plugin without the marker is preserved. Neither holds for the companion, which carries no marker at all. Moot if you collapse to one file.

3. `agterm-status-logic.mjs:55-58`, the insert guard fires on any event carrying `properties.info` with `id` and `parentID`, not only session events. OpenCode's `message.updated` carries `info`, and `Assistant` has a required `parentID` (`packages/schema/src/v1/session.ts:453-461`), so every assistant message id ends up in a set named `childSessions`. Nothing misbehaves today, ids are prefixed `msg_` vs `ses_` so the lookup can never match, but the invariant is wrong and the set never shrinks. `session.deleted` also adds the id instead of removing it, since the guard runs before the switch.

4. the node suite isn't wired to anything. `.github/workflows/ci.yml` has no node step and its paths-filter doesn't cover `tests/**`, so a PR touching only that file runs zero jobs and reports green. Nothing in the `Makefile` or `scripts/test.sh` runs it either, and `node --test tests/opencode-plugin/` fails on node 26, only the explicit file path works. Separately the part most likely to break in a user's home, `spawnReport` and `defaultWrapperPath`, has no coverage at all, the loader tests stop at `typeof hooks.event === "function"`. Mostly resolves itself if the collapse above moves coverage to Swift.

5. `agtermCore/Sources/agtermCore/AgentHooksInstall.swift:49`, `defaultShellAgentRegex` has no production caller, only the two new tests reference it. Every other constant in that file is used from `AgentHooksInstaller.swift`. Keep the expected string local to the test.

nits:

- `agterm-status.js:32-39`, the 1s timer is never cleared, so every report leaves an armed timer that later kills a reaped pid. `child.kill()` also reaches only the wrapper shell, the `agtermctl` descendant survives and still delivers its side effect, so the timeout abandons the report rather than cancelling it and the queue advances past a call that is still running.
- `agterm-status-logic.mjs:112`, `enqueue` returns the unprotected `pending`, only `chain` gets the `.catch`. Nothing reaches it today, `spawnReport` never rejects. Still worth `return pending.catch(() => {})`, since `chat.message` goes through `Effect.promise` on OpenCode's side where a rejection is a defect that kills the prompt fiber. `pi/agterm-status.ts:16-20` states that guarantee explicitly.
- `AgentHooksInstallTests.swift:345-375`, three things. The `!re.contains(name)` loop at :348 can't fail, :347 already pins `re` to the exact literal. `"pi"` is a substring match, so adding `copilot` to the default set later trips an assertion that reads "must not include pi". And the line filter at :365 also selects `integration.fish:7`, the commented example, so a doc comment gets asserted as if it were the shipped default (`integration.sh:7` escapes it only because it uses `export ...RE='` rather than `:=`).
- `.gitignore`, `Readme_herdr.md` and `TODO.md` are local scratch files, neither is tracked nor produced here, and `TODO.md` would hide a real repo-level TODO from everyone later. Pls drop the hunk, `.git/info/exclude` is the place for those.
- the herdr references in the comments (`agterm-status-logic.mjs:5` and `:102`, and the "child whitelist matches herdr" test name). These files get installed into `~/.config/opencode/plugins/`, and a reader there has no way to tell what herdr is or why agterm's mapping is defined by parity with it. Pls state the reason directly instead, smth like "child-session events that still mean the parent pane needs the user" and "serialize reports so a slow spawn can't reorder status".
- semantic line breaks in `.claude/rules/control-api.md`, two of the added lines are 167 and 121 cols. `agterm/Resources/agent-skill/troubleshooting.md:133` is 149. The rest of the added prose is fine.

culler127:
Thank you so much for such a detailed and thorough review! 
’ve done my best to address all of your feedback. 
All the fixes are already committed and ready for another look.

umputun (commented):
round-1 items check out. Both structural asks are in, blocking 1-3 are moot with the collapse, 4 moved into `swift test`, 5 is gone, and all six nits are addressed.

two blockers, both about state the mapper doesn't keep. OpenCode's own status plugin (`packages/tui/src/feature-plugins/system/notifications.ts` in the opencode repo) is a good reference for both, it solves exactly these with two `Set`s.

**blocking:**

1. `agterm/Resources/agent-status/opencode/agterm-status.js:44` with `:28-29`, `session.error` -> `blocked` gets overwritten milliseconds later. `SessionProcessor.halt` publishes `Session.Event.Error` and then calls `status.set(sessionID, {type:"idle"})` on the next line (`packages/opencode/src/session/processor.ts:619-624`, same pair at :611-612). `idle` maps to `completed --auto-reset`, and the report queue is strictly serial, so the order is guaranteed rather than racy. A failed turn ends up showing the completed glyph, while `README.md:465`, `site/docs.html` and `.claude/rules/control-api.md` all say `session.error` sets `blocked`. Their plugin latches it: `errored.add(sessionID)` on `session.error`, then on `session.status(idle)` `if (errored.has(sessionID)) { errored.delete(sessionID); return }`, cleared again on busy/retry.

2. `agtermCore/Tests/agtermCoreTests/OpenCodeStatusHookTests.swift:94`, `:130`, `:269`, `--experimental-default-type=module` no longer exists in node, so node exits 9 before importing anything and every test in the suite fails. Red on node 26 locally and red in CI on 59ad71da (unit tests: 9 issues across 6 tests; build and swiftlint pass). Drop the flag from all three sites, the harness is `.mjs` and the plugin loads as ESM through node's own module-syntax detection without it. Worth fixing while you're there: `standardError` goes into a `Pipe()` nothing reads, which is why the failure prints a bare `terminationStatus == 0` and never names the cause.

**nits:**

- `agterm-status.js:36-57`, the mapper ignores `properties.sessionID`, so a subagent's `session.status(idle)` reports `completed --auto-reset` while the parent session is still busy, and `needsAttention` (`agtermCore/Sources/agtermCore/AgentStatus.swift:11`) pulls that still-working session into attention nav and auto-follow. It corrects on the parent's next transition, so it's a flicker rather than a stuck state. Same file as above handles it with an `active` set: busy/retry adds the id, idle returns early unless the id is in the set. That also covers an `idle` with no preceding `busy`, and it's narrower than the child machinery I asked you to drop.
- `OpenCodeStatusHookTests.swift:41`, `:124`, `:207`, a missing node hits `Issue.record`, so `swift test` fails instead of skipping on a machine without node. `CodexStatusHookTests` drives `/bin/bash`, which is always present, and README's requirements don't mention node. Pls make it a skip via `.enabled(if:)` and add a clause to the build-from-source requirements.
- `agterm/AgentHooksInstaller.swift:452`, the `.noOpenCode` arm says "Start OpenCode once, then run this again", but `copyBundledFolder()` refreshes the shell integration on every run, so that user has already lost coarse `opencode` detection and gets nothing back until `~/.config/opencode` exists. Keep the regex removal, it matches how claude/codex/pi are handled, but say in that arm that coarse shell detection for opencode is off by default now.

culler127:
Thanks for the round-2 review
 — the pointer to `packages/tui/src/feature-plugins/system/notifications.ts` made both blockers
straightforward; I followed it directly, including its `!sessionID` / `!active.has(sessionID)` guard on
`session.error`. All five items are addressed, docs updated in lockstep.

All the fixes are already committed and ready for another look.

Verified: 1749 tests green on node 22.23 and 26.4.0, skips cleanly without node, lint and build pass.

umputun (commented):
round-2 blocking 2 and nit 3 check out. Blocking 1's latch is in but brings a new problem, nit 1 isn't fixed, and nit 2's gate has a hole.

**blocking:**

1. `agterm/Resources/agent-status/opencode/agterm-status.js:91-100`, the `active` set doesn't do what nit 1 asked. A `task` subagent gets its own session and its own busy/idle pair, so by the time its idle arrives it is already in `active`: parent busy, child busy, child idle returns `COMPLETED`. `packages/opencode/src/tool/task.ts:159` creates the child with `parentID` and prompts it by the child's id, `session/run-state.ts:59-68` publishes busy/idle under that id, and `plugin/index.ts:251-258` fans both to every hook filtered only by directory. The parent doesn't re-enter `SessionProcessor.process` while it waits on a foreground `task` (`processor.ts:638`), so nothing corrects the glyph until the tool call returns and the row sits in attention nav and auto-follow for that whole span. `README.md:466`, `site/docs.html`, `.claude/rules/control-api.md:163` and the comment at `agterm-status.js:54-56` all state the guarantee now.

   `OpenCodeStatusHookTests.swift:222-233` passes because it never sends `busy` for `ses_child`, which isn't what opencode does. Simplest fix is to make the gate set-wide rather than per-id:

   ```js
   case "idle":
     if (!sessionID || !active.has(sessionID)) return null;
     active.delete(sessionID);
     if (errored.has(sessionID)) { errored.delete(sessionID); return null; }
     if (active.size > 0) return null;
     return COMPLETED;
   ```

   that also covers two interleaved top-level sessions. Their plugin does it the other way, `api.state.session.get(sessionID)?.parentID` at `packages/tui/src/feature-plugins/system/notifications.ts:76-77`, and you have `client` on the plugin input if you'd rather resolve it.

2. `agterm-status.js:108-112`, the `errored` latch makes Esc stick on `blocked`. `Effect.onInterrupt` calls `halt(new DOMException("Aborted","AbortError"))` (`processor.ts:648-654`) and halt publishes `session.error` then `status.set(idle)` on the next line, so the latch swallows the idle and `blocked` is the terminal report. `blocked` has no auto-reset and ranks first in attention nav (`agtermCore/Sources/agtermCore/AgentStatus.swift:11`, `:29-38`), and the keystroke clear doesn't save it: `clearedByKeystroke` fires on key-down, before the byte reaches opencode, so the `blocked` report arrives after it and overwrites. Before this round the same sequence ended on `completed --auto-reset`, which self-clears on visit.

   same arm, smaller case: `ContextOverflowError` with auto-compaction on publishes the error and returns with no following idle (`processor.ts:615-617`), so a routine compaction paints `blocked` (and fires the Blocked sound if it's set) until the loop resumes with busy.

   `event.properties.error` is on the payload, so you can skip `error?.name === "MessageAbortedError"` and the overflow error and latch only for the errors that actually end the turn. Their plugin treats the abort as a transient toast, `sessionErrorMessage` at `notifications.ts:21`.

**nits:**

- the docs say `session.error` sets `blocked` unconditionally (`README.md:466`, `site/docs.html:1993-1999`, `.claude/rules/control-api.md:164`) but the code gates it on `active.has(sessionID)`. The same README sentence spells out the analogous precondition for `idle`, so it reads as an oversight rather than a choice. `sessionID` is optional on `EventSessionError` and at least two publishers omit it (`skill/index.ts:114`, `plugin/index.ts:136`). Pls keep the gate, it matches their plugin and is what makes the latch safe, and add "for a session already reported busy" to the three surfaces.

- `README.md:108` says "without Node those tests skip", but the `.enabled(if:)` gate only checks that `node` resolves. The plugin is a `.js` with no `package.json` above it, so it loads only through module-syntax detection, unflagged in 20.19 and 22.7. Below that the suite is enabled and all nine tests fail on the import instead of skipping. Pls name the floor in that line.

- `spawnReport` has no upper bound now and `createReportQueue` serializes every later report behind it. I take the point from `94b75f3` that the 1s timer was killing a reaped pid and reordering under load, so this isn't a request to put it back as it was. But `agtermctl`'s socket read has no timeout and `ControlServer.handleConnection` ends in a `DispatchSemaphore.wait()` with no timeout either, so a wedge anywhere on that path parks the chain for the life of the opencode process. If you want a bound, make it a real cancellation instead of an abandonment: `spawn(..., { detached: true })` plus `process.kill(-child.pid)` on a generous timeout, 5-10s rather than 1. Not blocking.

branch is 16 commits behind master, pls rebase with the next push.

culler127:
Thanks — the pointers into `processor.ts` and `task.ts` made both blockers easy to pin down.

- **Set-wide gate.** `idle` now completes only when the active Set is empty after removing the id, so a
  task subagent's own busy/idle pair can't complete a busy parent. The old test never sent `busy` for
  the child, as you spotted — it does now, plus a case for two interleaved top-level sessions.
- **Transient errors.** `MessageAbortedError`, `AbortError` and `ContextOverflowError` no longer latch,
  so Esc ends on `completed --auto-reset` and compaction doesn't flash blocked. Both covered by tests.
- **Docs.** "for a session already reported busy" added to all three surfaces.
- **Node floor.** The gate now checks the version, not just that `node` resolves; README names
  ≥ 20.19 / ≥ 22.7. Verified with a stub reporting v21.7.3 — the suite skips instead of failing.
- **spawnReport.** Took your suggestion: 10s bound with `detached` + `process.kill(-pid)`. Confirmed it's
  real cancellation — with a wrapper that hangs on a `sleep 60` child, the descendant is gone afterwards.

Rebased onto master (now on top of a1930e5, no merge commits) and force-pushed. 1805 tests, lint and
build all green on the new base.

umputun (commented):
round-3 items check out. The set-wide gate, the transient-error skip, the node floor, the doc precondition, the bounded `spawnReport` and the rebase are all in.

one blocker, same class as the last two rounds: this fix moved the gate but not the latch.

**blocking:**

1. `agterm/Resources/agent-status/opencode/agterm-status.js:121-130`, the `errored` latch is per-id but `COMPLETED` is now decided set-wide, so when one opencode instance has two `ses_*` sessions in flight, whichever idles second erases the other's `blocked`. Both feed the same agterm pane (one plugin instance, one `AGTERM_SESSION_ID`), so that pane ends a failed turn showing `completed --auto-reset`. Two in flight is the same setup as round-3 blocking 1, a `task` subagent gets its own sessionID and its own busy/idle pair, and `interleavedTopLevelSessionsCompleteOnlyWhenAllIdle` covers the other way to get there. Call them A and B, both busy: `session.error(A)` reports `blocked` and sets `errored={A}`; A's own idle is swallowed at :124-127 **and clears the latch**; B's idle then finds `active` empty and nothing latched, so it reports `completed --auto-reset` over the `blocked`. `halt` publishes the error and then `status.set(idle)` on the next line (`processor.ts:620-624`), so error-then-its-own-idle is the ordinary order rather than a race. Pls keep the id in `errored` when its own idle is swallowed, and consult the latch where `COMPLETED` would be returned, smth like `if (errored.size > 0) { errored.clear(); return null; }`, still clearing per-id on busy/retry. Test: `busy(A)`, `busy(B)`, `error(A)`, `idle(A)`, `idle(B)` → `["active --blink","active --blink","blocked"]`.

**nits:**

- `agterm-status.js:138-144` with `:22-23`, the `ContextOverflowError` skip is by name, and that name also comes back on the branch where the turn really ends. With `compaction.auto === false` (or `OPENCODE_DISABLE_AUTOCOMPACT`) `halt` publishes the error and then sets idle (`processor.ts:608-613`), so the failed turn falls through to the idle arm and reports `completed --auto-reset`; same result when compaction gives up and `runLoop` breaks. Both painted `blocked` before this round. Rather than deciding at error time, record the id and return null, then let a following busy/retry clear it, which is what keeps compaction from flashing blocked, and a following idle report `BLOCKED`. `MessageAbortedError`/`AbortError` are fine on the unconditional skip.
- `OpenCodeStatusHookTests.swift:286-314`, neither test can observe the regression it guards. Both are the same four-event fixture with different ids, and the oracle is three argv lines with no ordering in it. Swap `active.delete(sessionID)` for `active.clear()` at `agterm-status.js:123` and the recorded output is byte-identical while `completed` fires on the child's idle with the parent still busy. Inserting `event("permission.asked")` between the two idles and asserting `["active --blink","active --blink","blocked","completed --auto-reset"]` makes the ordering observable; the second test can then be a different interleaving rather than the same one renamed.
- `README.md:108` and the `.enabled(if:)` message at `OpenCodeStatusHookTests.swift:68`, "≥ 20.19 or ≥ 22.7" reads as two lower bounds unioned, which admits 21.x and 22.0-22.6, and the gate rejects both. The doc comment at :34-35 already states it correctly. Smth like "22.7+ (or 20.19+ on the 20.x line)".

culler127:
Thanks . 
Latch now stays set-wide through the swallowed idle (same A/B case you described), 
overflow is decided by the next event, and the gate tests / Node wording are updated.

All the fixes are already committed and ready for another look.

umputun (commented):
round-4 items check out. The set-wide latch through the swallowed idle, the deferred overflow decision, the strengthened gate tests and the node floor wording are all in.

merging as is. One leftover below, not worth another round, pls pick it up in a follow-up if you want it.

**nit:**

- `agterm/Resources/agent-status/opencode/agterm-status.js:144-154`, a clean turn can end with the pane still on `active --blink`. The comment at `:149` ("leave its blocked standing") holds only while nothing has repainted the pane since the error, and a busy for another session does exactly that. `busy(A)`, `error(A)` → `blocked` with `errored={A}`; `idle(A)` swallowed at `:146` with the id kept; `busy(B)` → `:129` deletes only B, so `errored` is still `{A}` and `active --blink` is painted over the blocked; `idle(B)` → `:150` sees `errored.size > 0`, clears the set and returns null. Neither terminal state survives. `README.md:472` says `idle` sets `completed --auto-reset` "only when no session remains busy", and here none does and nothing is reported.

  it's the sequential form of the two-session case rather than an interleaving, so it needs no concurrency: A's turn ends in an error, the next prompt goes to B. `active` isn't in `needsAttention` and doesn't pull auto-follow, so it's a stale blinking badge and not a wrong-pane jump, and the next turn clears it. Before `d4eba48` that sequence completed normally, which is why I'd still rather it went in.

  `turnAfterAnErrorCompletesNormally:358-376` can't catch it, its second turn reuses `ses_root` whose busy clears its own latch at `:129`. The ordering that wants a test is `busy(A)`, `error(A)`, `idle(A)`, `busy(B)`, `idle(B)`, and nothing in the suite currently has a busy between the swallowed idle and the clearing idle.

  the invariant is "once every session is idle, the pane is never left showing `active`", and two shapes hold it. `errored.clear()` on busy/retry at `:129` is a one-word change, needs no edit to any existing expectation, and generalizes what that line already encodes ("work resumed, so the error wasn't the turn's end") from per-id to set-wide, which matches `COMPLETED` already being decided set-wide; the sequence above then ends on `completed --auto-reset`. Or return `BLOCKED` at `:150-153` instead of null, which costs a trailing duplicate `blocked` in `siblingIdleDoesNotCompleteOverAnotherSessionsError:341-356` and, where a subagent fails but the parent recovers and finishes, ends the turn on `blocked`. I'd take the first.

PR #311, opened 2026-07-28
add the new-session-in-workspace recipe

skripalschikov:
## What is the problem?

⌘N creates a session in the workspace you are already in. Putting one in a *different* workspace means going there first — find it in the sidebar, select it, then ⌘N, then navigate back to what you were doing. The cost of starting a session somewhere else scales with how far away that somewhere is, and with a dozen workspaces it stops being worth doing.

`session new --workspace-name` already covers this over the socket, but typing an exact workspace name into a shell is not a mid-flow action.

## How does this solve it?

One chord opens a floating overlay running three fzf prompts: pick the workspace (fuzzy-matched, so a few letters is enough), name a new one if you picked the create entry, name the session or leave it blank. Then a single `session new --workspace-name [--create-workspace] [--name]` does the work.

The destination becomes a search box, so it costs the same whether the workspace is the current one or the twelfth one down, and you never leave the session you pressed the key in. Creating the workspace is folded into that same call rather than being a step of its own, so cancelling at any prompt leaves nothing behind.

## What's in it

`cookbook/new-session-in-workspace/` — one POSIX script plus the README, and the index row.

The script is a single file that re-invokes itself: a custom command runs detached with no tty, so it cannot host fzf. It opens an overlay running `sh <itself> --pick`, and that second invocation draws the prompts. Same shape as the bundled `agent-skill/scripts/show-image.sh`, including its `sq()` single-quote escaper for the path embedded in the overlay command.

The version pin is 0.8.0, for `overlay open --follow` — what pulls you to the overlay when the key is pressed from a different session. The other commands it uses are older.

## On the missing Discussion

CONTRIBUTING asks for a Discussion before a first PR and I have not opened one — deliberately, and I will move this to one if you would rather.

That step exists so an idea that does not fit the project's direction is caught before the work. For a recipe, the direction question is already answered in writing: `cookbook/CONTRIBUTING.md` specifies what a recipe has to be, and this meets it — a workflow I actually run, one kebab-case directory, the six headings, a version pin, an index row. It adds no Swift and touches no shared surface, and `cookbook/` is documented as pinned snapshots rather than a maintained API. What is left is whether the recipe is useful and whether it is correct, which is a review question rather than a design one.

If that reads as skipping the step rather than a reason for it, say so and I will open the Discussion.

## Notes

- It has been on ⌘T in my own keymap for a couple of weeks. The committed version differs from mine only where the recipe rules require it: the hardcoded home path replaced by self-location, `AGTERMCTL` / `AGT_BIN_PATH` / `AGT_OVERLAY_SIZE` overrides, and `/usr/local/bin` alongside `/opt/homebrew/bin` for an Intel Homebrew.
- One deliberate behavioural change from my own version: the launcher passes `--socket` explicitly. My original exported `AGTERM_SOCKET`, which does nothing — `agtermctl` resolves `--socket`, then `AGTERM_STATE_DIR`, then app-support, and never reads `AGTERM_SOCKET`. Passing `$AGT_SOCKET` (custom command) or `$AGTERM_SOCKET` (overlay pty) through means the recipe also works under an `AGTERM_CONTROL_SOCKET` override. Written up in *How it works*, since it looks load-bearing and is not.
- *Limits* names the two non-obvious ones: picking `＋ New workspace…` creates a real workspace, and the key is a silent no-op in a window whose sessions have all exited, because there is nothing for `--target active` to resolve to.
- shellcheck clean, and the whole `cookbook` CI job run locally: layout, index in both directions, the six headings, shebang, and `zsh -n` over the existing `.zsh` recipes.


umputun (commented):
two things before merge.

on the missing Discussion: not needed for a recipe. `cookbook/CONTRIBUTING.md` is the rule for these and it says to open a pull request with the directory, its README and the index row in one change; the root CONTRIBUTING covers everything else.

`cookbook/new-session-in-workspace/agt-new-session.sh:17` - `AGT_BIN_PATH` and `AGTERMCTL` never reach the picker. The overlay's shell is spawned from the app's own environment plus the `AGTERM_*` set, so nothing exported for the launcher propagates, and a prefix on the keymap line reaches the launcher only. The half that runs `fzf`, `jq` and the tree call still gets the defaults. Fix: prefix the assignments onto the overlay command, which is `eval`ed, so `AGTERMCTL=$(sq "$AGTERMCTL") AGT_BIN_PATH=$(sq "${AGT_BIN_PATH:-/opt/homebrew/bin:/usr/local/bin}") /bin/sh $(sq "$self") --pick` works. Or say in *Setup* that both have to be set in the app's environment.

`cookbook/new-session-in-workspace/README.md:84` - the all-sessions-exited case is not silent. `agtermctl` exits non-zero when the target does not resolve, and a custom command exiting non-zero posts a failure banner naming the status. The error text is lost, the failure is not. Fix: keep the "output goes nowhere" half, drop "silently", and name the banner.

skripalschikov:
Both fixed in 843de6c, and thanks — the first one was a real bug, not a doc gap.

**`AGTERMCTL` / `AGT_BIN_PATH` not reaching the picker.** You're right, and it defeated the point of the indirection: the overrides were honored in the launcher, which needs neither, and dropped in the picker, which is where `fzf`, `jq` and both CLI calls actually run. Took the prefix option, since telling the reader to set them in the app's environment asks for a `launchctl setenv` to fix a script variable.

`AGT_BIN_PATH` now resolves to a variable at the top instead of inline, so the default is written once and the same value is what gets carried across:

```sh
agt session overlay open \
    "AGTERMCTL=$(sq "$AGTERMCTL") AGT_BIN_PATH=$(sq "$AGT_BIN_PATH") /bin/sh $(sq "$self") --pick" \
    --target "${AGT_SESSION_ID:-active}" --follow --size-percent "${AGT_OVERLAY_SIZE:-60}"
```

Both `sq`-escaped like the path. Verified by evaluating the built command the way `OverlayCapture.shellLine` does, under `env -i` with a launchd-like `PATH` and only `AGTERM_SOCKET` set: the picker resolves the overridden binary and finds `fzf` through the carried `AGT_BIN_PATH`. The same eval without the prefix reaches neither, which was the bug. *How it works* now explains the boundary, since "exported, therefore inherited" is the natural wrong assumption.

**The empty-window case is not silent.** Correct, and my wording was wrong twice over — the exit is non-zero and `notifyCommandFailure` posts a *Command failed* banner carrying the name and `exit <status>`. Reworded to keep the lost-error-text half, drop "silently", and name the banner; it also notes the banner is gated on Settings ▸ Notifications, since that is the one case where nothing appears.

**On the Discussion** — understood, and thanks for the clarification. I read the root CONTRIBUTING's first-PR rule as covering everything in the repo; that `cookbook/CONTRIBUTING.md` is self-contained for recipes is the part I missed.

Behaviour is unchanged — the eight prompt paths (each workspace kind × named/blank session name × cancel at each of the three prompts) still produce the same `session new` invocations. `shellcheck` and the rest of the `cookbook` job are clean locally.

PR #328, opened 2026-07-30
docs(cookbook): add the opencode-session-resume recipe

cherkale:
Third session-resume recipe, next to `claude-session-resume` and `codex-session-resume`: each agterm tab reopens its own opencode conversation after a restart. From [discussion #71](https://github.com/umputun/agterm/discussions/71#discussioncomment-17845561).

opencode will not take a session id at launch either, so the function maps tab id to session id the way the codex one does. It cannot glob for the conversation, though: sessions live in sqlite, so the id comes from `opencode session list --format json`, which lists the current project's top-level sessions while `--session` is not scoped at all — so everything the list offers is something a resume will accept, and a subagent's child session, which the list leaves out, can never be mapped by mistake. The session row also appears only once the first message is sent, not when the TUI starts, so the id is adopted after the run instead of captured at launch.

What the shape follows: one `.zsh` holding both functions, committed without the executable bit; README in the six-heading template with the byline; index row in the same change; Requirements naming agterm 0.3.1 or later, `zsh`, and opencode 1.18.10 as the version this was checked against.

Verified against the real CLI at 1.18.10:

- an unknown `--session` id makes opencode print `Session not found` and exit instead of starting, which is why the function never passes an id it has not just seen listed
- a mapped id reopens exactly that conversation, with `ps` showing `opencode --session ses_…`
- the parse returns precisely the ids `session list` prints, and a subagent's child session is in the database but absent from the list
- a TUI left open without a message creates no session row at all

Verified against a stub `opencode`, for the branches that need many runs: passthrough of subcommands and of an explicit `--session`, adopting the conversation a run creates, resuming a mapped one, the replayed `--session` line being re-decided rather than doubled, a deleted conversation falling back to a fresh run and re-adopting, two tabs keeping separate mappings, and the exit code passing through.

The cookbook gates were run locally: index in both directions, the six headings, kebab-case, shebang, `zsh -n`. The function is installed in my own `~/.zshrc` and drives my opencode tabs.

Limits keeps the two you called out — the `session list --format json` shape and adopting another tab's id — plus the mapping-written-only-after-exit one the codex recipe shares. Two entries are new since the thread, both from re-checking scoping afterwards: outside a git repository the project is `global`, which spans every non-repository directory on the machine, so the neighbour whose id can be adopted may be a tab working somewhere else entirely; and the existence check being project-scoped means a tab used in another repository starts a fresh conversation there and the mapping moves with it. That same re-check corrected what the thread post claimed — that the list and `--session` are scoped alike. They are not: the list is per project, `--session` takes any session that exists. The recipe only needs the list to be a subset of what a resume accepts, which it is.

### DIFF
--- a/cookbook/README.md
+++ b/cookbook/README.md
@@ -18,9 +18,10 @@ Recipes come from other people as well as the maintainer. Every one is reviewed 
 | [status-announcer](status-announcer/) | demo: speak agent status changes from a dedicated session | 0.16.0, jq |
 | [claude-session-resume](claude-session-resume/) | each tab reopens its own Claude Code conversation after a restart | 0.3.1, zsh, Claude Code |
 | [codex-session-resume](codex-session-resume/) | each tab reopens its own codex conversation after a restart | 0.3.1, zsh, Codex CLI |
+| [opencode-session-resume](opencode-session-resume/) | each tab reopens its own opencode conversation after a restart | 0.3.1, zsh, opencode |
 | [new-session-in-workspace](new-session-in-workspace/) | pick a workspace with fzf and start a session in it | 0.8.0, jq, fzf |
 
-Most recipes need `agtermctl` on your PATH; **Help ▸ Install Command Line Tool…** puts it there. The two session-resume recipes are shell functions that read the environment agterm gives a session and never call the CLI. Some recipes also need `jq`, `fzf`, or a particular shell, and each recipe's *Requirements* section says which, along with the minimum agterm version it needs. Recipes are snapshots rather than a maintained surface: the control API grows by addition, so they rarely break, and one that does gets fixed when it is reported.
+Most recipes need `agtermctl` on your PATH; **Help ▸ Install Command Line Tool…** puts it there. The three session-resume recipes are shell functions that read the environment agterm gives a session and never call the CLI. Some recipes also need `jq`, `fzf`, or a particular shell, and each recipe's *Requirements* section says which, along with the minimum agterm version it needs. Recipes are snapshots rather than a maintained surface: the control API grows by addition, so they rarely break, and one that does gets fixed when it is reported.
 
 ## Contributing a recipe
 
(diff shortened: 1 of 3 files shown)


umputun (commented):
good with this, one fix in the Usage steps first.

`cookbook/opencode-session-resume/README.md:48` - the sequence skips the quit, but the mapping is only written after opencode exits (`opencode-resume.zsh:57`). Following it on a fresh install gives a new conversation and a bare `opencode` in `ps`, not `opencode --session ses_...`. Your Limits bullet at :72 says exactly that, and the version in discussion #71 had the quit step.
Fix: "start it, send at least one message, quit it, then start it again in the same tab".

umputun (approved):
quit step is back, that's what I wanted. Merging this. Two nits below, fix them or leave them, either is fine.

`README.md:19` says `$XDG_STATE_HOME` is honored, and the script does honor it (`opencode-resume.zsh:29`), but the removal step at `:44` and the Usage paragraph at `:52` both name `~/.local/state/opencode/agterm/` flat. A reader with a custom state home follows the removal step and leaves his bindings behind. `${XDG_STATE_HOME:-~/.local/state}/opencode/agterm/` in both spots would match what the code does.

`README.md:13` and `:52` say an explicit `--session` passes through untouched. It does, except when the id you type is the one already mapped: the replay branch at `opencode-resume.zsh:32` matches on id equality alone, so `opencode --session <mapped-id> --fork` drops the flag and re-decides. Fresh conversation instead of the fork. Narrow enough that I'd soften the wording rather than restructure the check.

cherkale:
Both fixed in #330 

One refinement on the second one, after re-running it: the flag is taken back only when the id is the tab's own, and what that costs depends on whether the conversation is still in the project's list. Listed, `opencode --session <mapped-id> --fork` reaches opencode as `--session <mapped-id> --fork` and forks fine; not listed, it arrives as a bare `--fork` and opencode rejects it. The fresh conversation is the case without `--fork` — `opencode --session <mapped-id>` for an id the list does not hold, which opencode itself would have opened. The wording now says that, and the check is unchanged. Table with all four cases is in the follow-up.

PR #336, opened 2026-08-01
docs(cookbook): add the kimi-agent-status recipe

x9x9x9x9x9x91:
The recipe promised in #332, in the cookbook shape you suggested there: Kimi Code needs none of the installer's app-side machinery — the stock `agterm-agent-status.sh` works unmodified — so the whole delta is four `[[hooks]]` entries in `~/.kimi-code/config.toml` plus the explanation. README-only, like overlay-and-split.

Following your notes on the Discussion:

- **Minimum version** is worked out from what the recipe uses, not what I run: the installed status script, `--blink`, and `--auto-reset` all predate the earliest tag (their commits are ancestors of v0.3.1), so it names 0.3.1 with the same wording the claude-session-resume recipe uses for the same situation.
- **PermissionRequest exercised for real** (kimi-code 0.31.0, manual mode, driven to an actual approval dialog): it fired only while the dialog was on screen and never for an auto-approved tool — no codex-style false-flag, so no pane watcher in the recipe. The no-dialog modes (auto, yolo, `-p`) never fire it; that's in *Limits*, together with the sharper consequence that a question from the agent arrives as plain chat, so a completed row can really be a question waiting.
- **Exit-code-2-blocks and hook-stdout-reaches-the-model** are in *How it works*; both verified live (a PreToolUse probe exiting 2 stopped the write, and its stdout visibly steered the model's reply).

Also verified while writing: hooks load at kimi startup only (a hook added while a TUI ran did not fire), hence the restart sentence in *Setup*.

Everything stated was tested on kimi-code 0.31.0 + agterm 0.19.1, wired the same way this config has been running since the #332 post.

### DIFF
--- a/cookbook/README.md
+++ b/cookbook/README.md
@@ -19,6 +19,7 @@ Recipes come from other people as well as the maintainer. Every one is reviewed 
 | [claude-session-resume](claude-session-resume/) | each tab reopens its own Claude Code conversation after a restart | 0.3.1, zsh, Claude Code |
 | [codex-session-resume](codex-session-resume/) | each tab reopens its own codex conversation after a restart | 0.3.1, zsh, Codex CLI |
 | [opencode-session-resume](opencode-session-resume/) | each tab reopens its own opencode conversation after a restart | 0.3.1, zsh, opencode |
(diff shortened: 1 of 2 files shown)


umputun (approved):
lgtm. The four hook entries match what the installer writes for Claude Code, the script path and flags are right, and 0.3.1 is the correct floor.

PR #340, opened 2026-08-01
fix(agtermctl): survive oversized requests instead of dying on SIGPIPE

### LINKED ISSUE #338: agtermctl: any request over 1 MiB kills the CLI with SIGPIPE (exit 141, no output) — client socket never sets SO_NOSIGPIPE
**What is the problem?**

The control-socket client never sets `SO_NOSIGPIPE` on its fd, while the server caps a request line at 1 MiB and closes the connection when that's exceeded. So any request over 1 MiB kills `agtermctl` with SIGPIPE before it can read the server's error: exit 141, zero output.

The sequence:

1. `SocketClient.connect()` opens the socket without `SO_NOSIGPIPE` (`agtermCore/Sources/agtermctlKit/SocketClient.swift:47-81`), and `writeAll` loops raw `write()`s (`:84-94`).
2. The server reads the request through a 1 MiB cap (`agterm/Control/ControlServer.swift:78`, enforced at `:283`), replies `request too large or read failed` (`:242`), and closes the connection.
3. A multi-MiB request doesn't fit the socket buffer in one write, so the client is still mid-`writeAll` when the peer closes — the next `write()` raises SIGPIPE, which is default-fatal. The process dies on signal 13 with no output, and the server's error message is never read.

The server protects itself against exactly this case (`ControlServer.swift:231-233` — "a write to a client that already hung up would raise the default-fatal SIGPIPE and take the whole app down"); the client is missing the same option, and nothing else in the CLI handles SIGPIPE (`rg -n 'SIGPIPE|sigaction' agtermCore/Sources/agtermctl*` comes back empty).

This is reachable from ordinary use because `session type --stdin` reads stdin unbounded into a single request (`SessionCommands.swift:205`) — piping a large file crosses 1 MiB easily, and a big `pick` list or long `session restore` command can too.

Repro on 0.19.1 (macOS 26.5.2):

```
$ agtermctl session new --name sigpipe-repro --json
{"result":{"id":"2996D9ED-EFCB-4274-A7B8-ADEF382EF6FB"},"ok":true}

$ printf 'x%.0s' {1..3000000} | agtermctl session type --stdin --target 2996D9ED-EFCB-4274-A7B8-ADEF382EF6FB; echo "exit=$?"
exit=141

$ printf 'hello' | agtermctl session type --stdin --target 2996D9ED-EFCB-4274-A7B8-ADEF382EF6FB; echo "exit=$?"
ok
exit=0
```

The 3 MB payload produces no output at all — nothing on stderr, nothing typed into the session — so a script can't tell it from a crash, and `--json` consumers get nothing to parse. The small control payload works fine.

Related mismatch that makes this reachable: the client's line cap is 64 MiB (`SocketClient.swift:25`) and its doc comment assumes "requests stay small", but nothing actually bounds the request side, while the server rejects at 1 MiB. So even with SIGPIPE fixed, an oversized request would die with `write failed: Broken pipe` instead of the server's clearer "request too large".

**Suggested fix**

Two small changes in `SocketClient`:

1. Set `SO_NOSIGPIPE` on the fd right after `socket()` in `connect()`, mirroring the server, so the write fails as a normal `EPIPE` through the existing `SocketClientError` path instead of killing the process.
2. Check the encoded request length against the server's cap before the first write, so the failure is a clean "request too large" message — ideally with the 1 MiB constant hoisted into `agtermCore` so the two sides can't drift.

Happy to send a PR for both.

x9x9x9x9x9x91:
Fixes #338.

A request over the server's 1 MiB line cap made the server close the connection mid-write; the client fd had no `SO_NOSIGPIPE` (the server side sets it, with a comment naming this exact hazard), so agtermctl died on signal 13 — exit 141, zero output.

Two changes, both needed — the signal fix alone still loses the server's "request too large" message, because `writeAll` throws on EPIPE before `readLine` ever runs:

- `SO_NOSIGPIPE` on the client fd right after `socket()`, mirroring the server. Darwin-only guard so the Glibc test build (7de50a4) keeps compiling — Glibc has no `SO_NOSIGPIPE`, and that build only serves the test suite.
- The encoded request is checked against the cap before connecting, so an oversized request fails with a readable error naming the size and the cap. The newline is appended after the check, matching the server's newline-excluded count in `readLine`.
- The cap moves into agtermCore as `ControlWire.maxRequestLineBytes` and `ControlServer` reads it from there, so the two sides can't drift.

Tests: an oversized request throws before connecting (error prefix pinned, so a connect failure can't satisfy it); a hang-up stub server proves a mid-write peer close surfaces as a thrown `write failed` error rather than killing the process — removing the `setsockopt` makes the test runner die on signal 13, so the test has teeth.

Verified: `swift test` (2114 tests green), `make build`, `make test-app`, `swiftlint --strict` clean; live repro before/after — the 3 MB `--stdin` payload from the issue went from exit 141 with no output to exit 1 with `request too large (3000078 bytes, cap 1048576)`, and a just-under-cap payload still goes through.

### DIFF
--- a/agterm/Control/ControlServer.swift
+++ b/agterm/Control/ControlServer.swift
@@ -73,9 +73,9 @@ final class ControlServer {
         }
     }
 
-    /// 1 MiB cap on a request line, far above any realistic `session.type` payload; over it the line is
-    /// rejected and the connection closed, so a bad client can't grow the buffer unbounded.
-    nonisolated private static let maxLineBytes = 1 << 20
+    /// Cap on a request line, shared with the client via `ControlWire` so the two sides can't drift; over
+    /// it the line is rejected and the connection closed, so a bad client can't grow the buffer unbounded.
+    nonisolated private static let maxLineBytes = ControlWire.maxRequestLineBytes
 
     /// Seconds a blocking client read may stall before timing out (EAGAIN → connection closed), so a stalled
     /// client can't park the serial accept loop forever.
--- a/agtermCore/Sources/agtermCore/ControlProtocol.swift
+++ b/agtermCore/Sources/agtermCore/ControlProtocol.swift
@@ -336,12 +336,11 @@ public struct ControlArgs: Codable, Sendable, Equatable {
     }
 }
 
-/// Wire-level limits both ends of the control socket share. The server rejects a request line over
-/// `maxRequestLineBytes` and closes the connection; the client checks the same cap before writing, so an
-/// oversized request fails with a readable error instead of a write to a closing peer.
+/// 1 MiB cap on a request line (newline excluded), far above any realistic `session.type` payload. Over it
+/// the server rejects the line and closes the connection, so a bad client can't grow the buffer unbounded;
+/// the client checks the same cap before writing, so an oversized request fails with a readable error
+/// instead of a write to a closing peer. Shared so the two sides cannot drift.
 public enum ControlWire {
-    /// 1 MiB cap on a request line (newline excluded); over it the server rejects the line and closes the
-    /// connection, so a bad client can't grow the buffer unbounded.
     public static let maxRequestLineBytes = 1 << 20
 }
 
--- a/agtermCore/Sources/agtermCore/ControlProtocol.swift
+++ b/agtermCore/Sources/agtermCore/ControlProtocol.swift
@@ -336,6 +336,15 @@ public struct ControlArgs: Codable, Sendable, Equatable {
     }
 }
 
+/// Wire-level limits both ends of the control socket share. The server rejects a request line over
+/// `maxRequestLineBytes` and closes the connection; the client checks the same cap before writing, so an
+/// oversized request fails with a readable error instead of a write to a closing peer.
+public enum ControlWire {
+    /// 1 MiB cap on a request line (newline excluded); over it the server rejects the line and closes the
+    /// connection, so a bad client can't grow the buffer unbounded.
+    public static let maxRequestLineBytes = 1 << 20
+}
+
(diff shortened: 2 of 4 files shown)


umputun (commented):
code is fine. Both halves of the fix do what #338 needs, the client cap matches the server's enforcement byte-for-byte, and both new tests fail if their production fix is removed. Only two comment fixes below, minor but worth doing before merge.

`agtermCore/Sources/agtermctlKit/SocketClient.swift:71-72` - the Darwin-only guard is justified with "the Glibc build only serves the test suite, which never writes to a dead peer", but `writeToHungUpPeerThrowsInsteadOfDying` in this same commit is unguarded and does exactly that. Fix: cut that clause, "Darwin-only, Glibc has no `SO_NOSIGPIPE`" is the whole constraint.

`agtermCore/Sources/agtermCore/ControlProtocol.swift:339-346` - the enum doc and the member doc three lines below it restate the same contract, and the old `ControlServer.maxLineBytes` doc's "far above any realistic `session.type` payload" went with it, so the one fact about how the cap was sized is no longer anywhere in the tree. Fix: single doc on `ControlWire` carrying that sizing rationale plus the newline exclusion, drop the member doc, matching `OverlayResultError` and `ControlNotify` in the same file.

x9x9x9x9x9x91:
Both applied in 810790e — the Glibc clause was contradicted by the new test in the same commit, you're right; and the sizing rationale now lives on the single ControlWire doc with the newline exclusion, member doc dropped. Build/tests/lint re-run green.

umputun (approved):
lgtm. Last round's two items check out, nothing new.

Issue #324, opened 2026-07-30
Resize cursor does not hold over the sidebar and split dividers


andryushkin:
## What is the problem?

Hovering either divider only *flashes* the ↔ resize cursor: it appears for a moment, then reverts to the terminal's I-beam, and keeps alternating on every mouse move. Aiming at a divider takes several tries. Dragging works — the pointer feedback is what's broken.

Two places, one cause:

- the sidebar / terminal divider;
- the divider between two split panes.

### Steps to reproduce

1. Open a window with the sidebar visible.
2. Move the pointer onto the divider between the sidebar and the terminal, then move it a few points up and down without leaving the divider.
3. Split a session and repeat on the divider between the two panes.

**Expected:** ↔ stays for as long as the pointer is over the grab area.
**Actual:** ↔ and the I-beam alternate on every move.

### Environment

agterm 0.19.0, macOS 26.5.2, Apple silicon. Also present on current `master` (6700661).

## Cause

`NSCursor` is process-global, and over a divider two writers set it for the same event.

The chrome is drawn *over* a pane, but `GhosttySurfaceView`'s tracking area deliberately ignores sibling overlap — `GhosttySurfaceView+Tracking.swift` documents this, and it is what issue #225 relies on. So a pane keeps receiving `mouseMoved` while the pointer is over the divider drawn on top of it, and re-asserts the libghostty shape there. Meanwhile the sidebar handle sets ↔ from `.onHover`, which fires once on entry, and AppKit owns the split divider's cursor. One set on entry cannot win against a writer that runs on every move.

The sidebar handle is 12pt centred on its 1px line, so roughly half of it lies inside the terminal surface; the split divider's grab area likewise reaches past the drawn line into both panes.

## Measurements

Instrumenting every cursor writer and logging 520 events across several crossings:

- `mouseMoved` fires continuously while the pointer is over the divider, as predicted;
- `applyMouseShape` and `GHOSTTY_ACTION_MOUSE_VISIBILITY` never fired in those runs — the terminal does not hide the cursor, it repaints it;
- after the pane stops writing, a cursor replacement still occurs **after** a synchronous `NSCursor.set()` from the SwiftUI handler returns. Re-asserting once more on the next runloop turn is what makes the handle stable; with either half alone it flickers.

I have not identified which framework path performs that last replacement — that would need a symbolic breakpoint on `-[NSCursor set]`. The ordering requirement is measured, the mechanism behind it is not.

## How it can be fixed

[...]

Issue #50, opened 2026-06-30
feat: keyboard/CLI command to resize the split divider


lenis2000:
**What I'd like:** a way to resize a session's split panes from the keyboard / `agtermctl`, not only by dragging the divider with the mouse.

**Current state:** the divider is mouse-draggable, which works fine. But there's no programmatic path:
- `agtermctl session split` only does `on|off|toggle`, and `session focus` only does `left|right|other` — neither changes the ratio.
- There's no `resize_split` (or similar) built-in action, so there's nothing to `map` in `keymap.conf`.
- `agtermctl window resize` exists but resizes the whole macOS window, not the panes inside a session.

**Proposal (either/both):**
1. An `agtermctl session resize` command — e.g. `--split-ratio 0.7`, or relative `--grow-left 5%` / `--grow-right 5%`, with a `--target` like the other session commands. This makes it scriptable and lets me bind it via a `command "…"` in `keymap.conf`.
2. A bindable built-in action (e.g. `resize_split_left` / `resize_split_right`, or a `grow_pane`/`shrink_pane` pair) so it can go straight in `map …` without a helper script.

**Why:** coming from Ghostty/kitty I had keyboard split-resize bound (e.g. "shrink the right pane to minimum"), and it's the one split operation that currently has no keyboard or CLI equivalent in agterm.

**Environment:** agterm 0.3.1, macOS 26.5.1.

umputun:
shipped in #59: `agtermctl session resize`.

- `--split-ratio R` sets the absolute left-pane fraction (0..1)
- `--grow-left D` / `--grow-right D` nudge it relatively

it clamps to 0.05..0.95, persists across relaunch, and prints the applied fraction.

I kept it CLI/control-only instead of a dedicated built-in action: it mirrors `window resize`, and a keymap binding is one custom command away. Drop this in `keymap.conf`, then `agtermctl keymap reload`:

```conf
command "grow left pane"  cmd+ctrl+l agtermctl session resize --grow-left 0.05
command "grow right pane" cmd+ctrl+h agtermctl session resize --grow-right 0.05
command "even split"      cmd+ctrl+0 agtermctl session resize --split-ratio 0.5
```

your "shrink the right pane to minimum" case is `--grow-left` held down (or a single `--split-ratio 0.95`), since 0.95 is the left-pane cap.

[...]

PR #341, opened 2026-08-01
feat(split): restore an even split on a divider double-click

skripalschikov:
Double-clicking a split's divider snaps it back to 50/50. A drag can never land exactly on an even split, and `NSSplitView`'s own double-click only collapses a pane through a delegate SwiftUI owns, so the gesture is recognized from one app-wide mouse monitor that sees the second click after the first one's drag-tracking loop ends — dragging is untouched.

The claimants are the probes whose panes are on screen and uncovered (the deck's `visible` stays true under a floating overlay, whose terminal would otherwise pass the chrome hit test), and a press only counts as a double-click if the divider has not moved since the previous one, so fine-tuning with a nudge-drag and re-grabbing does not throw the adjustment away. This is a GUI shortcut for an existing capability — `session resize --split-ratio 0.5` already did it over the control channel — so no new protocol command.

Also here: the issue #324 ownership rule now lives once as `terminalOwnsHit(_:host:)` with `ownsPointer` calling it, and the band arithmetic moved to host-free `SplitDividerBand` in `agtermCore`.

## Test plan

- `swift test` — 2104 tests pass, including 6 new `SplitDividerBandTests` covering hairline gaps, thick dividers, zero slop and the masked titlebar strip.
- `SplitUITests.testDoubleClickDividerRestoresEvenSplit` drives a real drag then a real double-click and reads back the persisted fraction. Verified against a negative control: with the gesture disabled the test fails at the reset assertion (ratio stays 0.675).
- `SplitUITests.testSplitFocusKeyboardNavAndCollapse` still passes; Debug and Release build clean; `swiftlint --strict` reports zero findings.
- Not automated: the floating-overlay and compact-titlebar paths, and the cursor behaviour around the extracted `terminalOwnsHit` (manual per the libghostty rules).

### DIFF
--- a/.claude/rules/menu-actions.md
+++ b/.claude/rules/menu-actions.md
@@ -94,6 +94,14 @@ paths:
 - Persist each pane cwd and the 0...1 left-pane `splitRatio`. `SplitRatioAccessor` is an unconditional
   background representable on primary, introspects `NSSplitView`, retries until width exists, observes
   `didResizeSubviews`, and debounces save by about 0.4 seconds. Regular saves and quit flush also persist it.
+- Double-clicking the divider restores `splitRatioDefault`. AppKit offers no hook: `NSSplitView`'s own
+  double-click collapses a pane through the delegate SwiftUI owns. One app-wide `SplitProbeView` monitor,
+  never one per probe, sees the second click after the first one's drag-tracking loop ends, so dragging is
+  unaffected; consume that press and its release so neither starts a drag nor reports a phantom button-up.
+  `clickCount == 2` also matches a re-grab after a nudge-drag, so require the divider not to have moved
+  since the previous press. Claimants are the probes whose panes are on screen AND uncovered: the deck's
+  `visible` stays true under a floating overlay, whose terminal passes `terminalOwnsHit`. Host-free
+  `SplitDividerBand` owns the band arithmetic and already excludes the masked titlebar strip.
 - `SplitRatioAccessor` masks only the compact-titlebar divider overrun. At 30pt compact height, SwiftUI
   padding lies inside the safe-area band and AppKit expands `NSSplitView` full height; normal 48px mode is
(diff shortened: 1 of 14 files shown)


umputun:
going to pass on the gesture.

the even split is one custom action away, written out in the agterm skill's examples: `command "even split" cmd+ctrl+0 agtermctl session resize --split-ratio 0.5` in `keymap.conf`. I kept `session resize` CLI/control-only on #50 on purpose, and a mouse alias for the one fraction isn't worth a new permanent code path.

the tradeoff underneath it doesn't tune away either. The gesture needs a grab band beside the divider, and that band is terminal text: at 3pt slop a double-click within 3pt of the divider snaps the split instead of selecting a word, and at zero slop the target is a 1pt line nobody can hit. Word selection is the far more frequent gesture and it loses either way. On top of that this would be the app's first mouse-event monitor, installed for the process lifetime and able to swallow a left click app-wide. The five existing monitors are all `.keyDown`/`.flagsChanged` on transient UI.

`terminalOwnsHit` and the host-free `SplitDividerBand` are a different matter, they stand on their own. Send those as their own PR and I'll take them.

umputun:
the even-split gesture is in, as #357. Not this patch: a different implementation of the same feature.

what changed my mind: #344 went in two hours after I passed on this, and made the divider band resolve from `split.hitTest` rather than a guessed width, for both the resize cursor and the pane's own `ownsPointer`. That killed the tradeoff I objected to. The gesture sits on that band now, so it only claims pixels where the split already swallows the press for its own drag, and word selection loses nothing. No slop constant, no `SplitDividerBand`, and no band arithmetic at all.

the monitor is still a monitor, but it goes in with the first split and comes out once the last probe leaves its window, so an app that never splits installs none.

what did carry over is the shape of the gesture: one app-wide monitor seeing the second press after the first one's tracking loop ends, the paired-release swallow, and requiring the divider not to have moved since the previous press so a nudge-drag re-grab keeps the adjustment. That last one earns its place.

PR #354, opened 2026-08-02
fix: sanitize control chars in names and --cwd on the control/GUI paths (#347)

### LINKED ISSUE #347: `session rename` (control path) doesn't strip control chars, so a newline in a name survives into the `{AGT_SESSION_NAME}` `/bin/sh -c` sink
`TerminalText`'s doc-comment already spells out the invariant this misses: a session name/cwd "flow unquoted into a `/bin/sh -c` line via the `{AGT_SESSION_NAME}`/`{AGT_SESSION_PWD}` custom-command tokens, so a control character — above all a newline, which `sh -c` reads as a command separator — must never survive into the stored value." The OSC path enforces that (`TerminalText.sanitized`, applied where a terminal program sets the title/cwd — `GhosttySurfaceView.swift:366,382`). The **control path doesn't** — `agtermctl session rename` runs the name through `trimmedOrNil` only, which trims surrounding whitespace but leaves an interior newline intact.

### Chain (all on master @aa4a96d)

- The control-socket rename arms hand the name straight to the store with no sanitizing — `session rename` (`agterm/Control/ControlServer+SessionActions.swift:214`), `workspace rename` (`ControlServer+WorkspaceCommands.swift:33`), `window rename` (`ControlServer+WindowCommands.swift:206`), and `session new --name` (via `AppStore.addSession`, `agtermCore/Sources/agtermCore/AppStore.swift:355`).
- The store keeps an interior newline. Every arm normalizes with `trimmedOrNil` (`Session.swift:604`), which is `trimmingCharacters(in: .whitespacesAndNewlines)` — it strips only *surrounding* whitespace/newlines, so a newline in the middle survives: `renameSession` (`AppStore+Naming.swift:9,11`), `renameWorkspace` (`AppStore+Naming.swift:18,21`), `renameWindow` (`WindowLibrary.swift:399-401`).
- `session.customName` → `session.displayName` (`Session.swift:322-323`) → `CommandContext.sessionName` (`agterm/Commands/CustomCommandRunner.swift:265`).
- A fired custom command expands `{AGT_SESSION_NAME}` **unquoted** into the command line — the token value is concatenated straight in (`CommandContext.expand`, `CustomCommand.swift:126`, `result += value`), and that line is run as `/bin/sh -c line` (`CustomCommandRunner.swift:290,293`). A newline in the name is a statement separator, so the text after it runs as its own command.

The `$AGT_SESSION_NAME` environment form is safe (it's a shell variable value, where a newline is data); the `{AGT_SESSION_NAME}` brace form is the unquoted one the doc-comment flags as the vector.

### Repro

Needs a keymap custom command that references the brace token (~1 min):

1. Bind any custom command whose command string contains `{AGT_SESSION_NAME}`, e.g. `echo {AGT_SESSION_NAME} >> /tmp/agt-name.log`.
2. `agtermctl session rename <id> $'demo\ntouch /tmp/agt-pwned'` — `session tree` shows the name back with the embedded newline, i.e. it survived the store.
3. Fire the bound command on that session → `/bin/sh -c` runs `echo demo`, then `touch /tmp/agt-pwned` as a second statement. `/tmp/agt-pwned` appears.

This matters because a session name is often set from untrusted-ish data — a branch name, a PR title, a filename — by exactly the kind of automation the control socket exists for.

### Fix

Run the control-path name/cwd arms through `TerminalText.sanitized` (the function and its rationale already exist), same as the OSC path. Scope is the invisible-control-char vector only — visible shell metacharacters in a name stay the caller's concern via the `$AGT_X` quoted env form, consistent with what the doc-comment already says. Happy to send the PR if you want it.

x9x9x9x9x9x91:
Fixes #347.

An interior newline in a session/workspace/window **name** or a session **`--cwd`** survives into the stored value and, when a custom command fires, expands unquoted into the `/bin/sh -c` line via the `{AGT_SESSION_NAME}` / `{AGT_SESSION_PWD}` / `{AGT_WORKSPACE_NAME}` / `{AGT_WINDOW_NAME}` tokens — where a newline is a statement separator. The OSC path already closes this with `TerminalText.sanitized`; the control-socket and GUI rename paths ran the value through `trimmedOrNil` only, which strips *surrounding* whitespace but leaves an interior newline intact.

### Placement

Sanitizing lives in the four `agtermCore` store methods — `renameSession` / `renameWorkspace` (`AppStore+Naming.swift`), `renameWindow` (`WindowLibrary.swift`), `addSession` (`AppStore.swift`). This is the option you leaned to: it covers the control arms, the GUI rename (`SidebarRenameController` → `renameSession`/`renameWorkspace`, `AppActions` → `renameWindow`) and whatever calls these next, and the tests stay host-free. I didn't argue an alternative — the boundary-reader cost (the control arm shows nothing) is paid down with a one-line comment at each store method naming the token and the sink.

`--cwd` is folded in as you asked: it seeds `initialCwd`, which feeds `effectiveCwd` → `{AGT_SESSION_PWD}` until OSC 7 reports, so `addSession` sanitizes both the name and the cwd.

Scope is the invisible control-character vector only — visible shell metacharacters in a name/path stay the caller's concern via the quoted `$AGT_X` env form, consistent with `TerminalText`'s doc-comment.

### Tests

One per arm pinning that an interior newline does not survive the store (`addSession` covers both the name and the cwd arm), with the existing blank/trim cases kept.

### Verification

- `./scripts/test.sh` — 2204 tests in 86 suites pass.
- `swiftlint lint --strict` — clean.

### DIFF
--- a/agtermCore/Sources/agtermCore/AppStore+Naming.swift
+++ b/agtermCore/Sources/agtermCore/AppStore+Naming.swift
@@ -6,7 +6,9 @@ extension AppStore {
     /// Sets a session's custom name, with a blank value restoring its automatic display name.
     public func renameSession(_ sessionID: UUID, to name: String) {
         guard let session = session(withID: sessionID) else { return }
-        let renamed = name.trimmedOrNil
+        // customName reaches {AGT_SESSION_NAME}, which expands unquoted into /bin/sh -c, so strip control
+        // characters (a newline is a statement separator there) as the OSC path does. See TerminalText.
+        let renamed = TerminalText.sanitized(name).trimmedOrNil
         guard session.customName != renamed else { return }
         session.customName = renamed
         scheduleTreeChanged()
@@ -15,7 +17,8 @@ extension AppStore {
 
     /// Renames a workspace. Blank and same-value names are structural no-ops.
     public func renameWorkspace(_ workspaceID: UUID, to name: String) {
-        guard let trimmed = name.trimmedOrNil,
+        // {AGT_WORKSPACE_NAME} expands unquoted into /bin/sh -c; strip control chars as the OSC path does (TerminalText).
+        guard let trimmed = TerminalText.sanitized(name).trimmedOrNil,
               let index = workspaces.firstIndex(where: { $0.id == workspaceID }),
               workspaces[index].name != trimmed else { return }
         workspaces[index].name = trimmed
--- a/agtermCore/Sources/agtermCore/AppStore.swift
+++ b/agtermCore/Sources/agtermCore/AppStore.swift
@@ -352,7 +352,11 @@ public final class AppStore {
     public func addSession(toWorkspace workspaceID: UUID, cwd: String, command: String? = nil,
                            name: String? = nil, wait: Bool = false, at index: Int? = nil, select: Bool = true) -> Session? {
         guard let wsIndex = workspaces.firstIndex(where: { $0.id == workspaceID }) else { return nil }
-        let session = Session(initialCwd: cwd, customName: name?.trimmedOrNil)
+        // Both seed values reach a custom-command token that expands unquoted into /bin/sh -c: cwd via
+        // initialCwd → effectiveCwd → {AGT_SESSION_PWD} until OSC 7 reports, and name via customName →
+        // {AGT_SESSION_NAME}. Strip control characters the same way the OSC path does. See TerminalText.
+        let session = Session(initialCwd: TerminalText.sanitized(cwd),
+                              customName: name.map(TerminalText.sanitized)?.trimmedOrNil)
         session.initialCommand = command
         session.commandWait = wait
         if let index {
--- a/agtermCore/Sources/agtermCore/AppStore.swift
+++ b/agtermCore/Sources/agtermCore/AppStore.swift
@@ -317,7 +317,8 @@ public final class AppStore {
     /// false leaves the target where it is. `ensureWorkspace(named:revealNewWorkspace:)` forwards both.
     @discardableResult
     public func addWorkspace(name: String, collapsed: Bool = false, revealNewWorkspace: Bool = true) -> Workspace {
-        let workspace = Workspace(name: name, isExpanded: !collapsed)
+        // {AGT_WORKSPACE_NAME} expands unquoted into /bin/sh -c; strip control chars as the OSC path does (TerminalText).
+        let workspace = Workspace(name: TerminalText.sanitized(name), isExpanded: !collapsed)
         workspaces.append(workspace)
         if revealNewWorkspace {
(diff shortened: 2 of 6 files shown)


umputun (changes requested):
the four arms from #347 are right, but the create-side twins of two of them still reach the same sink, so `{AGT_WORKSPACE_NAME}` and `{AGT_WINDOW_NAME}` stay injectable.

`agtermCore/Sources/agtermCore/AppStore.swift:320` - `addWorkspace` stores the name verbatim, so `workspace new --name` and `session new --workspace-name --create-workspace` still carry an interior newline into the unquoted `/bin/sh -c` line. `.workspaceNew` applies no validation, unlike `.workspaceRename` two cases below it.

```swift
let workspace = Workspace(name: TerminalText.sanitized(name), isExpanded: !collapsed)
```

that one needs the lookup needle sanitized as well, or `ensureWorkspace` stops being idempotent and appends a second workspace on every call. It belongs in `workspace(named:)` at :334 rather than in `ensureWorkspace` at :342, because `ControlServer+SessionActions.swift:120` calls `workspace(named:)` directly on the path without `--create-workspace`, so sanitizing only at :342 leaves a raw needle against a sanitized stored name and reads back as a silent "workspace not found".

```swift
guard let needle = TerminalText.sanitized(name).trimmedOrNil else { return nil }
```

`agtermCore/Sources/agtermCore/WindowLibrary.swift:320` - `newWindow` keeps the `name?.trimmedOrNil` shape that `renameWindow` was just changed away from, so `window new --name` reaches `{AGT_WINDOW_NAME}` the same way.

```swift
let info = WindowInfo(name: name.map(TerminalText.sanitized)?.trimmedOrNil ?? defaultWindowName)
```

`newWindowBlankNameFallsBackToDefault` still holds with that: `sanitized` leaves whitespace alone, so the trim-to-nil path is unchanged.

both arms want a store test each, plus one pinning that `ensureWorkspace` called twice with an interior newline yields a single workspace.

x9x9x9x9x9x91:
applied all three as suggested, da6d8c5. One extra edge on the same path: ensureWorkspace now sanitizes before its blank check too, so a control-char-only name reads as blank and returns nil instead of appending an unmatchable empty-named workspace on every call, pinned in the idempotency test.

The new tests pushed AppStoreTests.swift over the 2000 line lint budget, so the sanitize family moved to AppStoreNameSanitizeTests.swift. swift test (2208), make test-app and make lint all green.

umputun (approved):
all three are in. The `ensureWorkspace` blank-gate edge is a good catch and the review missed it: sanitizing only in `workspace(named:)` would have let a control-char-only name past `trimmedOrNil`, then stored it blank and appended an unmatchable workspace on every call.

test split is right, and the file was already at the 2000 ceiling.

PR #363, opened 2026-08-04
fix: decode sidebarMode lossily so an unknown value can't wipe the persisted tree (#360)

### LINKED ISSUE #360: Unknown sidebarMode value in workspaces.json wipes the whole persisted tree (decode throws, load() recovers to an empty Snapshot)
`SessionSnapshot`'s custom `init(from:)` already spells out the failure this one misses: an unknown enum value in a persisted field, on decode, "would fail the whole `SessionSnapshot` and make `PersistenceStore.load` start fresh, **wiping everything**" (`agtermCore/Sources/agtermCore/Snapshot.swift:174-177`). That comment guards `backgroundWatermark` with a lossy `(try? …) ?? nil` (`Snapshot.swift:192`), and the same guard protects `sessionRecency` (`Snapshot.swift:83`). `Snapshot.sidebarMode` decodes strictly and is the field the guard was never applied to.

### Chain (all on master @4d263a7)

- `Snapshot.init(from:)` decodes it with a plain `try c.decodeIfPresent(SidebarMode.self, forKey: .sidebarMode)` (`Snapshot.swift:72`).
- `SidebarMode` is a `String`-raw `Codable` enum, cases `tree`/`flagged` (`SidebarMode.swift:4`). `decodeIfPresent` tolerates a **missing** key (→ nil), but a present key with an unknown raw value throws `DecodingError.dataCorrupted` — it isn't the missing-key case the `Optional` covers.
- That throw fails the whole `Snapshot` decode, which `load()` runs under `try?` and recovers to a default empty `Snapshot()` (`PersistenceStore.swift:31`). Every workspace and session in that store is dropped.
- The loss then persists: the next mutation schedules a save (`AppStore.scheduleSave`, `AppStore.swift:403,640`) that writes the empty tree back over the file, so the original is gone even if the bad value was a one-character slip.

The value is always on disk to be corrupted — `snapshot()` passes the live `sidebarMode` (default `.tree`) unconditionally (`AppStore.swift:793`), so every saved `workspaces.json` (and each per-window `windows/<id>.json`) carries a `"sidebarMode"` key. The `version` gate doesn't catch it either: it's stayed `1` across the optional fields added since (`#32`, `#111`, `#255`, `#271`, `#297`), so a build reading a value a newer release introduced passes the version check and then throws on decode.

### Repro

Quit agterm with at least one workspace, then (~1 min):

1. In `~/Library/Application Support/agterm/workspaces.json`, change `"sidebarMode": "tree"` to `"sidebarMode": "grid"` (any value that isn't `tree`/`flagged` — a downgrade past a future mode, or a hand-edit typo).
2. Relaunch. The sidebar comes up empty — every workspace and session is gone, not just the sidebar mode.
3. Interact once (open a session, resize) and the empty tree is saved back over the file.

### Fix

Give `sidebarMode` the same lossy decode as its two neighbours — `(try? c.decodeIfPresent(SidebarMode.self, forKey: .sidebarMode)) ?? nil` — so an unknown value drops to nil (→ `.tree` via the existing `?? .tree` at the call site) instead of failing the load. There's a matching test gap: `PersistenceTests.swift:219` covers the *missing* `sidebarMode` key and `:340-351` covers a bad *value* but on `sessionRecency`; nothing covers a bad `sidebarMode` value. Happy to send the PR if you want it.

x9x9x9x9x9x91:
Fixes #360.

`sidebarMode` was the one enum in `Snapshot` still decoded strictly, so an unrecognized raw value in `workspaces.json` (written by a newer build, or a hand edit) failed the whole `Snapshot` decode — and `load()` recovers from a throw by starting fresh, wiping every workspace and session over a non-essential display field.

This applies the same `try?` guard `sessionRecency` and `backgroundWatermark` already have, scoped to `sidebarMode` only: the remaining strict optionals are primitives and can't gain a value a newer build writes, so strict stays correct for them.

Also adds the missing test case: `PersistenceTests` covered a bad `sessionRecency` value and a *missing* `sidebarMode` key, but not a bad `sidebarMode` value. New `malformedSidebarModeDropsToNilKeepingTree` mirrors `malformedRecencyDropsToNilKeepingTree` (unknown raw value + wrong JSON type, tree survives, field drops to nil).

Full suite (2341 tests) + `make lint` strict green.

### DIFF
--- a/agtermCore/Sources/agtermCore/Snapshot.swift
+++ b/agtermCore/Sources/agtermCore/Snapshot.swift
@@ -55,10 +55,12 @@ public struct Snapshot: Codable, Equatable, Sendable {
         case focusedWorkspaceID
     }
 
-    /// Custom decode so `sessionRecency` is LOSSY: a present-but-invalid list (a malformed UUID, a wrong
-    /// JSON type after a hand edit) drops to nil. `Optional` alone tolerates only a MISSING key, so one bad
-    /// MRU entry would fail the whole `Snapshot` and `PersistenceStore.load` would start fresh, wiping every
-    /// workspace and session over a non-essential field. Mirrors `backgroundWatermark` below; others strict.
+    /// Custom decode so `sessionRecency` and `sidebarMode` are LOSSY: a present-but-invalid value (a
+    /// malformed UUID, an unknown enum raw value from a newer build or a hand edit) drops to nil.
+    /// `Optional` alone tolerates only a MISSING key, so one bad value would fail the whole `Snapshot` and
+    /// `PersistenceStore.load` would start fresh, wiping every workspace and session over a non-essential
+    /// field. Mirrors `backgroundWatermark` below; the remaining optionals stay strict — primitives and
+    /// UUIDs can't gain a value a newer build writes.
     ///
     /// Also migrates the legacy `focusedWorkspaceID`: its presence implied the filter was on, so it becomes
     /// a one-member ENABLED set. Neither key present decodes to nil/nil — an empty, disabled filter.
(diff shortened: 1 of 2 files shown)


umputun (approved):
lgtm

Issue #199, opened 2026-07-12
agent-status mis-attributed to the wrong pane after promote + re-split (baked AGTERM_PANE)


umputun:
follow-up from #121 and the pane-identity discussion in #176.

`AGTERM_PANE` is baked into a split shell's env at spawn (`left`/`right`/`scratch`), and the agent-status hook forwards it as `--pane` on every status call. That baked value is a static pane identity, so it can't follow a pane whose role changes at runtime. #121 (promote the split survivor into the main pane on primary exit) makes that reachable in a new way.

**repro**

1. split a session: left = plain shell, right = an agent (claude/codex). The right shell is baked `AGTERM_PANE=right`.
2. exit the left shell. The survivor (the agent) promotes into the main slot. `closePrimaryPane` re-tags its existing `.right` status block to `.left`, and `setAgentIndicator`'s `!hasSplit` coercion keeps later statuses correct while there is no split.
3. re-split, e.g. to get a helper shell. `hasSplit` is true again and a fresh right pane opens.
4. the agent, now in the MAIN pane but with a shell still baked `right`, emits its next status. The hook calls `session.status --pane right`.

**result**

the `.right` to `.left` coercion in `setAgentIndicator` is gated on `!hasSplit`, so with a live split it does not fire and the block is tagged `.right`, attributed to the fresh helper pane rather than the main pane where the agent actually is. `tree` reports a self-consistent `split:true` + `statusPane:"right"` (no invariant catches it), attention-reveal focuses the wrong pane, and the block only clears if you type in the helper pane instead of the agent's.

**why it is not a quick fix**

the baked env is immutable after spawn, and the `!hasSplit` gate is deliberately unable to fire during a live split (it protects the split-realization window). So the coercion structurally cannot tell "agent in the promoted main pane emitting right" from "agent in the real fresh split emitting right", both arrive as `--pane right` with `hasSplit == true`. A proper fix needs a pane identity resolved live rather than baked, which is what #176 (`AGT_FOCUS`) is about. Note that #176's proposal is an app-injected variable for chord-fired custom commands, and the agent-status hook runs from inside the shell, so this path needs its own handling: reconcile the incoming `--pane` against live pane state on the `session.status` side rather than trusting the baked value.

low urgency and same class as the other baked-env edges. Filing so it is tracked instead of lost in the PR thread.

PR #364, opened 2026-08-04
cookbook: annotate Claude's replies in revdiff and send the notes back

p4elkin:
Adds `cookbook/annotate-claude-replies/`: a chord that opens everything Claude Code has said since your last prompt in revdiff, on the session you pressed it in, and puts your notes back at that session's prompt.

The gap it fills is composition rather than a new command. Several recipes run a TUI in an overlay, and `examples.md` shows `overlay open --block` reading a program's output file, but nothing yet takes that output and feeds it back into the session it came from. This does that end to end.

### What it looks like

One exchange usually produces several replies, because Claude writes, runs a tool, then writes again. They go into one markdown file under per-reply headings, so revdiff's sidebar becomes a table of contents that can name them in words. The prompt that started the exchange opens the file and the replies follow in order:

```
10:52 · what you asked
10:58 · reply 1
11:04 · reply 2, the one you just read
```

Notes come back with the lines they point at already quoted, and are sent as soon as you quit:

```
## Note 1 — 11:04 · reply 2, the one you just read, line 12
> the launcher already opens revdiff in an agterm overlay
which launcher? I never set this up
```

### Commands used

`session overlay open --block`, `session paste`, `session type --pane`, `notify`, `tree --json`, plus the `{AGT_SESSION_ID}` and `{AGT_PANE}` keymap tokens.

### Minimum version

**0.13.0**, from `{AGT_PANE}` reporting the pane a custom command fired from (#210) — that is how a note taken in a split gets routed back to the right half. Everything else is older: `session overlay open --block` and `notify` shipped in 0.3.1, `session type --pane` in 0.7.0.

### Behaviour worth reviewing

Three things a reader should agree to before running it, all stated in *Limits*:

- It injects text at the prompt of the session the chord fired in **and sends it**, so the agent acts on the notes without you seeing them first. `AGTERM_ANNOTATE_SUBMIT=0` stops at the prompt. Sending is the default because annotating and quitting being the whole interaction is the point of binding it to a chord, but it is the setting most worth a second opinion.
- It uses the system clipboard, because `session paste` is the only way to land multi-line text at a prompt without submitting it. The prior clipboard is saved and restored, but only if it was plain text. `AGTERM_ANNOTATE_INLINE=0` avoids the clipboard entirely.
- It reads Claude Code's session transcripts and copies replies into a file under `TMPDIR`. The recipe never deletes them; macOS `dirhelper` purges `/var/folders/…/T` by access time after 3 days, so they persist until then.

It closes no sessions, deletes no workspaces, and kills no shells.

### Three API details the README records

`session overlay open --block` exits with the program's own status, so a non-zero code can equally mean agterm refused before the program ever ran. Reading only the code turns `overlay already open` into a meaningless "exited with 1".

An overlay's command runs under the app's GUI `PATH`, which has neither `/opt/homebrew/bin` nor `~/.local/bin`, so a bare binary name there exits 127 and the overlay flashes shut. The script resolves an absolute path first. The same trap applies to the script itself, which agterm runs detached in a small-`PATH` `/bin/sh` with stdio on `/dev/null` — hence the log under `/tmp` rather than printed output.

`session paste` takes no `--pane` and always targets the main pane, so a chord fired from a split or scratch pane falls back to typing a one-line pointer at the notes file.

### Two things the transcript format forces

A hook only runs when a turn ends, so any copy of the message it makes is stale the moment it misses one. The hook therefore records `transcript_path`, and the reply is read live when the chord fires. A project directory holds one transcript per session, so picking the newest by modification time would open whichever session was busiest rather than the one you are in — that pairing cannot be worked out afterwards, which is why the hook exists at all.

Peer-session messages, background task notifications and slash-command output all arrive as *user* records. Treating one as a typed prompt puts the exchange boundary in the wrong place and hides replies. `userType` is `external` on genuine prompts too, and `isMeta` was false on a real `<teammate-message>`, so the tags are matched as well.

### Checks

`shellcheck` clean across all cookbook scripts, `sh -n` clean, six headings present, index row added, shebangs and executable bits set. Exercised under `env -i` with the stripped `PATH` agterm gives a detached custom command, covering the overlay-busy, no-transcript, repeat-press, split-pane, clipboard-restore, injected-boundary and optional-flag paths, and against real transcripts rather than only fixtures.

Optional revdiff flags (`--no-tree`, `--preview`) are added only when the binary advertises them, so a released revdiff gets `--wrap --wrap-indent=2` and is never handed an unknown flag.

### Open questions for you

Two things I would rather you decide than me: whether sending on quit should be the default, and whether a recipe that needs a Claude Code hook belongs in the cookbook at all, given the rest are pure `agtermctl`.

Happy to split, rename, or drop any part of it.

### DIFF
--- a/cookbook/README.md
+++ b/cookbook/README.md
@@ -22,6 +22,7 @@ Recipes come from other people as well as the maintainer. Every one is reviewed 
 | [kimi-agent-status](kimi-agent-status/) | Kimi Code sessions report agent status onto their sidebar row | 0.3.1, Kimi Code |
 | [claude-recap](claude-recap/) | one key lists what the Claude Code run in a session was working on | 0.10.0, zsh, jq, Claude Code |
 | [new-session-in-workspace](new-session-in-workspace/) | pick a workspace with fzf and start a session in it | 0.8.0, jq, fzf |
+| [annotate-claude-replies](annotate-claude-replies/) | mark up Claude's answers in revdiff and send the notes back | 0.13.0, revdiff, python3, Claude Code |
 | [project-launcher](project-launcher/) | pick a project anywhere — or type "project + prompt" — and get a session in its workspace | 0.19.0, jq |
 | [flagged-dashboard](flagged-dashboard/) | grid the flagged panes that are running something, one chord to show and dismiss | 0.20.0, jq |
 
--- a/cookbook/annotate-claude-replies/README.md
+++ b/cookbook/annotate-claude-replies/README.md
@@ -1,0 +1,126 @@
+# Annotate Claude's replies
+
+Mark up a long Claude Code answer the way you would mark up a diff, and send the notes back to the prompt you are sitting at.
+
+## What it does
+
+One chord opens everything Claude has said since your last prompt in [revdiff](https://github.com/umputun/revdiff), inside an overlay on the session you pressed it in. You put a note on any line you do not follow or disagree with. On quit the notes are put at that session's prompt and sent, so annotating and quitting is the whole interaction.
+
+A long answer is the case this is for. Quoting the parts you want to argue with is slow, and "the third paragraph" is worse.
+
+One exchange usually produces several replies, because Claude writes, runs a tool, then writes again. They all go into one markdown file, each under its own heading, and revdiff builds its sidebar from those headings. The prompt that started the exchange opens the file, and the replies follow in the order they happened:
+
+```
+10:52 · what you asked
+10:58 · reply 1
+11:04 · reply 2, the one you just read
+```
+
+The notes that come back are not only your comments. Each one quotes the lines it points at and names the reply it landed on, so the answer you get is about the sentences you marked:
+
+```
+## Note 1 — 11:04 · the reply you just read, line 12
+> the launcher already opens revdiff in an agterm overlay
+which launcher? I never set this up
+```
+
+## Requirements
+
+- agterm 0.13.0 or later. That is where `{AGT_PANE}` began reporting the pane a custom command fired from, which is how the notes get routed back into the right half of a split. Everything else it uses is older: `session overlay open --block` and `notify` shipped in 0.3.1, `session type --pane` in 0.7.0.
+- [revdiff](https://github.com/umputun/revdiff), on your `PATH`
+- Claude Code. The recipe reads Claude Code's own session transcripts, so no other agent will work without being ported.
+- `python3`, which macOS ships
+
+`AGTERMCTL`, `REVDIFF` and `PYTHON` at the top of the script take an absolute path if any of them is somewhere unusual. `REVDIFF` is worth knowing about if you keep more than one build: the script resolves whatever `revdiff` the widened `PATH` finds first, which is not necessarily the one your interactive shell picks.
+
+## Setup
+
+Copy the four files somewhere on your machine, say `~/.local/bin/agterm-annotate/`, and make them executable:
+
+```sh
+chmod +x annotate-replies.sh annotate-extract.py annotate-render.py save-transcript-path.py
+```
+
+The three helpers must sit **beside** `annotate-replies.sh`; it looks for them in its own directory.
+
+Register the Stop hook by adding an entry to the `Stop` array in `~/.claude/settings.json`, alongside anything already there:
+
+```json
+{
+  "hooks": {
+    "Stop": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 ~/.local/bin/agterm-annotate/save-transcript-path.py",
+            "timeout": 5,
+            "async": true
+          }
+        ]
+      }
+    ]
+  }
+}
+```
+
+Then add the keybinding to `~/.config/agterm/keymap.conf` and apply it with File ▸ Reload Keymap or `agtermctl keymap reload`:
+
+```
+command "Annotate replies"  cmd+ctrl+e  ~/.local/bin/agterm-annotate/annotate-replies.sh "{AGT_SESSION_ID}" "{AGT_PANE}"
+```
+
+Pick a chord that is free in your own keymap. A custom command cannot shadow a built-in, so one that collides is quietly demoted to palette-only and the chord appears to do nothing. On 0.19.0 and later `agtermctl keymap list` shows what every chord resolved to; before that, read `keymap.conf` and the commented list of built-in defaults at the top of it. Either way the command is always reachable by name from the action palette, which is the quickest way to tell a bad chord from a broken script.
+
+Two settings, read from the environment. Put them in front of the script in the keymap line, or change the defaults at the top of the script — a keymap change needs a reload before it takes effect, an edit to the script does not:
+
+- `AGTERM_ANNOTATE_SUBMIT=0` leaves the notes at the prompt for you to read over and send yourself. The default sends them, so quitting revdiff is the whole interaction.
+- `AGTERM_ANNOTATE_PROMPTS=2` reaches back a further exchange, and so on. The default of 1 is everything since your last prompt.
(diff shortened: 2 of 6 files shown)


umputun (commented):
two things to fix before merge, plus a handful of small ones. Answering your two questions first.

sending on quit stays the default. The design is right and three of the four prose sites already document it, only the opening summary contradicts them. And a recipe that needs a Claude Code hook is fine here, `claude-recap` already installs a `statusLine` command into the same `~/.claude/settings.json`.

one process note before the findings. You asked for review and then kept committing and pushing, so this review is against whatever the head happened to be when I started, and I re-anchored the line numbers onto `c13ab9d` afterwards. Pls hold the branch still once you ask for a review, otherwise the line numbers and sometimes the findings themselves go stale between reading and writing.

**blocking**

`annotate-replies.sh:174-199` - the only gate before pasting and sending Return is that the pointer file exists at `:74`, and that file outlives the Claude process that wrote it. After `/exit` in the pane, or an app relaunch that restores the pane to a login shell, the pointer and the transcript both still resolve, so the notes get pasted at a bare shell prompt and submitted. Every quoted line carries the `> ` prefix from `annotate-render.py:71`, which the shell reads as a redirect: it creates or truncates a file named after the next word, in the pane's cwd, before the command lookup fails. `README.md:91` says the opposite. Fix: gate the paste on a live agent, `foreground`/`splitForeground` are already in the `tree --json` you fetch at `:93` and are omitted at a bare prompt.

`annotate-replies.sh:73` + `save-transcript-path.py:31` - the pointer is keyed on the shell's baked `AGTERM_PANE` but looked up with the live `{AGT_PANE}`, and those diverge after a promote (#199). With an agent in each half of a split, closing the left pane promotes the right one, and the chord in the survivor opens the dead pane's replies and submits notes on them into the live agent. There is no `{AGT_PANE_ID}` token to reconcile the two from a script, so a *Limits* sentence is the proportionate fix.

**small**

`README.md:7` - says the notes land "unsent, for you to read over and send", but `:27` defaults `SUBMIT=1` and `:197` sends a Return. Setup, Usage and Limits all say it sends. Reword `:7` to match them.

`annotate-replies.sh:126` - "The sidebar lists them, newest first" is wrong, `annotate-extract.py:219-223` writes the prompt first then iterates in transcript order, and `shape()` labels the last section "the one you just read". This one is the text you hand revdiff as `--description-file`, so it is on screen while the reader works. Same claim at `:118`, `annotate-extract.py:14` and `README.md:85`, while `README.md:11` and `:101` already say oldest first.

`annotate-extract.py:124` - `picked = order[start:] or order[-1:]` reaches into the previous exchange when the newest prompt has no assistant text block yet, while `opening` is the new prompt, so the file pairs a new question with an old answer. Dropping the fallback alone does not fix it, `main:190-194` then substitutes `last_assistant_message` which is the same stale reply. Return an empty opening too:

```suggestion
    picked = order[start:]
    if not picked:
        return ("", ""), []
```

`README.md:56` - `/Users/you/.local/bin/...` is an absolute path under a home directory, which `cookbook/CONTRIBUTING.md` bars. It is the only `/Users/` in the whole cookbook, the other 16 recipes use `$HOME` or `~`. `claude-recap/README.md:63` has the closest case, a tilde path inside a `"command"` string in the same settings file.

`save-transcript-path.py:47-60` and `annotate-replies.sh:204` - the hook writes the full `last_assistant_message` plus `cwd` to `~/.cache/agterm-annotate/` on every finished turn in every agterm pane, whether or not the chord is ever pressed, and `:204` puts `last-notes.md` there too. *Limits* names only `TMPDIR`. Worth a bullet, and `claude_session_id`/`cwd` are never read back by anything in the recipe.

p4elkin:
All seven addressed on `55be21e`, and the branch is now still — I will not push again without saying so.

Taking the process note first: fair, and it was worse than stale line numbers. You reviewed one head while I was pushing another, so some of what you read no longer existed. That was my fault for asking and then continuing.

**Blocking — sending into a dead pane.** Fixed as you suggested, and the failure was worse than I had thought: the paste is what carries `> ` prefixes, so a bare shell would act on them as redirects before the command lookup failed. The pane's foreground now gates the whole send, read from the `tree --json` that was already being made, so it costs no extra call. `left` reads `foreground`, `right` reads `splitForeground`, and `scratch` reports unknown and is treated as live — no read-back exposes a scratch foreground, and that path never pastes anyway. Tested across all five combinations.

**Blocking — the promoted pane.** Confirmed, and I agree a *Limits* sentence is proportionate; there is no `{AGT_PANE_ID}` token to reconcile the baked value with the live one. Added, naming the promote and what it opens.

**`annotate-extract.py:124`.** Your suggestion alone did leave it broken, exactly as you said — `main` then substituted `last_assistant_message`, the same stale reply. The stored message is now only used when the transcript could not be read at all; when it reads fine and the exchange has no reply yet, nothing is shown. Three cases tested: unanswered prompt, missing transcript, normal exchange.

**The ordering claims.** All four corrected, including the `--description-file` text, which was the one actually on screen while someone works.

**`README.md:56`.** Corrected to a tilde, and thank you for the `claude-recap` pointer — I had convinced myself a hook command was not shell-expanded inside quotes, which your example shows is wrong.

**The cache directory.** Added to *Limits*: written on every finished turn in every pane whether or not the chord is pressed, holding each pane's last reply in full plus `last-notes.md`, and nothing prunes it. `claude_session_id` and `cwd` are gone from the record — you were right that nothing read them.

**`README.md:7`.** Reworded to match Setup, Usage and Limits.

umputun (commented):
all seven from last round check out, both blockers included.

one thing I did to your branch, against the stillness you promised: CI's `ruff` step had never actually run on this PR. The green run you saw was `c13ab9d`, before the rebase pulled that step into your base, and it fails on `55be21e` with two errors. I pushed `38888a2` rather than send you a third round for two lint lines, `startswith` collapsed into one tuple call and the hook's blind `except` marked `# noqa: BLE001` with its reason, since swallowing everything there is deliberate. It needs another CI approval from me before it goes green.

merging once it does. The rest is notes, none of it blocking.

**`save-transcript-path.py:31`** - overlay surfaces get `AGTERM_SESSION_ID` but no `AGTERM_PANE` (`SurfaceEnvironment.swift:26`, "overlay surfaces pass nil"), so `or "left"` keys an overlay's claude onto the main pane's pointer, and the main-pane chord then opens its replies. Same shape as the promote case you documented. Main, split and scratch always carry the variable, so accepting only those three is safe.

**`annotate-replies.sh:140,146,215`** - the gate answers "is anything running", not "is an agent running": ssh, `less` and tmux all read as `yes`, an empty probe becomes `unknown`, and only a literal `no` refuses. It is also read at `:125` and acted on at `:215`, with the whole revdiff session in between.

**`README.md:29`** - 0.13.0 predates the field the gate now depends on. Per CHANGELOG v0.20.2 a `--command` pane "always read as idle in `tree --json`", so on 0.13.0 through 0.20.1 the chord refuses to send while claude is plainly running. Fails closed, so it is a dead chord rather than a misdirected send.

**`README.md:122`** - "Nothing under `${TMPDIR}/agterm-annotate/` is ever deleted" is not what `annotate-extract.py:206-208` does; it unlinks every prior `*.md` in the pane's directory on each run.

**`README.md:81`** - `--preview` is not a revdiff flag and there is no rendered-markdown mode (master `98a25b5`), so the read-rendered-then-toggle-to-source flow is not available. `--no-tree` is right, including your note that it is master-only.

**`README.md:33`** - `cookbook/CONTRIBUTING.md` asks for the Python version in *Requirements*; the floor reads as 3.9.

one smaller: the Limits cleanup advice does not cover a clipboard manager, which takes the `pbcopy` into its own history and survives deleting both directories.

PR #365, opened 2026-08-04
feat(cookbook): add fish port of claude-session-resume

Arelav:
## Summary

- Adds `claude-resume.fish`, a fish port of the existing `claude-resume.zsh` per-tab Claude Code session resume recipe, next to it in `cookbook/claude-session-resume/`.
- fish's own autoload mechanism makes setup simpler than zsh's `source ~/.zshrc` step: drop the file in `~/.config/fish/functions/claude.fish` and it's picked up automatically, no restart or explicit sourcing needed.
- Updates the recipe README's Requirements, Setup, and Limits sections to cover both shells, and the cookbook index row.

Per CONTRIBUTING.md, `.fish` isn't one of the linted extensions (`.sh`/`.zsh`/`.py`), so nothing gates it in CI — flagging that here for review.

## Test plan

- [x] `fish --no-execute cookbook/claude-session-resume/claude-resume.fish` — parses clean
- [x] Live agterm tab: pin with `--session-id`, resume after restart with `--resume` — confirmed working

### DIFF
--- a/cookbook/README.md
+++ b/cookbook/README.md
@@ -16,7 +16,7 @@ Recipes come from other people as well as the maintainer. Every one is reviewed 
 | [native-dir-picker](native-dir-picker/) | pick a directory in the native picker and type it into the shell | 0.19.0, fd, jq |
 | [overlay-and-split](overlay-and-split/) | keymap lines: a stateful split toggle and TUI overlays | 0.10.0, jq |
 | [status-announcer](status-announcer/) | demo: speak agent status changes from a dedicated session | 0.16.0, jq |
-| [claude-session-resume](claude-session-resume/) | each tab reopens its own Claude Code conversation after a restart | 0.3.1, zsh, Claude Code |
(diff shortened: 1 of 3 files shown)


umputun (approved):
lgtm. One note, not blocking.

`cookbook/claude-session-resume/README.md` - worth a clause in *Limits*: the fish copy is autoloaded by non-interactive fish too, so a fish script calling `claude` from inside a tab is intercepted where the zsh copy would pass through. Setup already says "interactive or not", so the scope is stated; it's the consequence that isn't. `-p`/`--print` in the passthrough list covers most scripted calls, so the residual is narrow.

on the `.fish` gating you flagged: confirmed, nothing in CI reads it. The shellcheck, `zsh -n` and ruff steps match only `.sh`/`.zsh`/`.py`, and the shebang sweep uses the same three. The structural checks (headings, index, kebab-case) are extension-independent and do still cover the recipe.

PR #366, opened 2026-08-04
fix(cookbook): passthrough non-interactive fish claude calls

Arelav:
## Summary

Follow-up to #365. Fish autoloads `claude.fish` for scripts too, unlike zsh's `.zshrc`-only sourcing (the zsh version never sees a non-interactive shell). Concretely: Claude Code's own Bash tool calling `claude` from inside an already-running agterm tab hit the resume logic and could collide on the tab's session id (`Session ID ... is already in use`) instead of passing through untouched.

- `claude-resume.fish`: guard the top of the function with `status is-interactive`, passthrough otherwise.
- `README.md`: note the zsh/fish difference and the guard in the Setup section.

## Test plan

- [x] `fish --no-execute cookbook/claude-session-resume/claude-resume.fish` — parses clean
- [x] Verified manually: non-interactive `claude` call from a script inside a tab now passes through untouched instead of hitting the resume logic

### DIFF
--- a/cookbook/claude-session-resume/README.md
+++ b/cookbook/claude-session-resume/README.md
@@ -56,7 +56,7 @@ mkdir -p ~/.config/fish/functions
 cp claude-resume.fish ~/.config/fish/functions/claude.fish
 ```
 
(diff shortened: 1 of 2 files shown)


umputun (approved):
lgtm

PR #370, opened 2026-08-05
fix: capture running commands before window-close surface teardown

### LINKED ISSUE #369: Exiting by closing the window loses every captured running command (restore-running-command)
**What is the problem?**

With **Restore running commands on restart** enabled, quitting via ⌘Q captures each pane's
running command and re-runs it on the next launch — but exiting by **closing the window**
(red button / the last window's close, which auto-quits the app) silently loses every captured
command: all panes come back as plain shells. Closing a NON-last window has the same defect —
its sessions' commands are never captured at all, so reopening that window also restores plain
shells.

Repro (released 0.20.2):

1. Settings ▸ General → Restore running commands: on.
2. In a session, run something long-lived, e.g. `sleep 500`.
3. Exit by closing the window (red button), not ⌘Q. The app auto-quits after the last window.
4. Relaunch → the session is a plain shell; the state file has `foregroundCommand: null`.
5. Same steps but quitting with ⌘Q → `sleep 500` is re-run. The two exits behave differently.

**Why it happens**

Closing a window tears its surfaces down in `WindowAccessor`'s `willClose` (and saves the store)
BEFORE the auto-quit reaches `applicationWillTerminate` — whose capture walks the sessions but
reads only panes that still HAVE a surface. By then there are none, so nothing is captured and
the save persists nulls. ⌘Q works because the capture runs while every surface is still alive.

**How the fix solves it**

Capture the closing window's sessions in `willClose`, right before `store.save()` and the surface
teardown — the last moment the foreground pids are readable. The capture is skipped when
`library.isTerminating` (the ⌘Q path has already captured; a second read could overwrite a good
value with nil for a foreground that exited in between). This also fixes the non-last-window
close, which previously never captured. Covered by a UI test exercising the
close-the-last-window exit (`RestoreCommandUITests.testWindowCloseExitCapturesRunningCommand`).

PR with the fix follows.

i-kozlov:
Fixes #369

**What is the problem?**

With Restore running commands on restart enabled, exiting by closing the window (red button /
last window's close → auto-quit) silently loses every captured running command: closing tears
each surface down in `WindowAccessor`'s `willClose` (and saves the store) before the auto-quit
reaches `applicationWillTerminate`, whose capture reads only panes that still have a surface —
by then there are none, so the save persists nulls and every pane restores as a plain shell.
⌘Q works because its capture runs while the surfaces are still alive. Closing a NON-last window
had the same defect: its sessions' commands were never captured at all. Details and manual repro
in #369.

**How does this solve it?**

Capture the closing window's sessions in `willClose`, right before `store.save()` and the
surface teardown — the last moment the foreground pids are readable. The capture is skipped
under `library.isTerminating`: the ⌘Q path has already captured every open session, and a second
read could overwrite a good value with nil for a foreground that exited in between. The capture
core is hoisted into a static `AppDelegate.captureForegroundCommands(sessions:)` so the quit
path and the close path share one implementation. This also makes closing a non-last window
capture its commands.

**Testing**

- New UI test `RestoreCommandUITests.testWindowCloseExitCapturesRunningCommand`: seeds a running
  `tee`, closes the last window over the control socket, asserts the auto-quit completed, and
  verifies the command re-runs on relaunch.
- Full `RestoreCommandUITests` suite green (17 tests), so the ⌘Q capture, the `session restore`
  overrides, and the split/command-session paths are unaffected.
- `make build`, `make test` (2345 tests), `make test-app`, `make lint` all green.
- Manual before/after on a stand: running `sleep 500`, window-close exit — state carries
  `foregroundCommand: null` before the fix, `["sleep","500"]` after; ⌘Q path unchanged.


umputun (changes requested):
the last-window exit fix is right, but the capture isn't scoped to that exit.

`agterm/Views/WindowAccessor.swift:155` - the guard is only `!library.isTerminating`, which is false for an ordinary mid-session close too, so every close now persists the live foreground argv, and `AppStore.session(from:launchRestore:)` copies `foregroundCommand` back unconditionally at `AppStore.swift:976` (`launchRestore` gates only the sticky override at :984). Closing a window running `npm run dev` or `ssh prod` and reopening it in the same run re-executes it, against `.claude/rules/settings.md:104` and README.md:559, both of which say clean-quit only.
Fix: gate `foregroundCommand`/`splitForegroundCommand` on `launchRestore` in `session(from:)`, next to the existing override gate. Two callers: `restore(from:launchRestore:)` propagates, and `workspace(from:)` (:991) is reached only from Reopen Closed Item, where suppressing the replay is also right. Keeps the non-last-window capture you added.

`testWindowReopenDoesNotArmTheOverride` can't catch this - it pins an override on an idle pane, so nothing is captured and it passes either way. Worth a regression test that closes a command-running window while another stays open, reopens it, and asserts the marker is not recreated.

umputun (changes requested):
the `launchRestore` gate is right and every caller checks out, but the capture is still not scoped to the exit, so the round-1 problem moved rather than closed.

`agterm/Views/WindowAccessor.swift:155` - the guard is still only `!library.isTerminating`, so an ordinary mid-session close captures the live argv and `store.save()` writes it into `windows/<id>.json`. The gate at `AppStore.swift:984` suppresses the copy in memory only, nothing clears the persisted value, a closed window has no store left to overwrite it, and `restore.clear` walks `allOpenSessions()` so it can't reach the file either. Repro: run `npm run dev` in window 1, close it, keep working in window 2, then exit by closing window 2. `closeWindow` nils `stores[id]` before reading `activeWindowID`, so the index saves `frontmost: nil` with every entry closed, and the next launch's `reopen` falls back to `windows.first` with `launchRestore: true` and re-runs window 1's `npm run dev`. Window 2's capture, the one #369 is about, is discarded.
Fix: capture only when this close is the app exit. `library.openIDs() == [windowID]` identifies it there, since `closeWindow` hasn't run yet. That drops the non-last-window capture, but it has no correct consumer: a mid-run reopen is gated now, and the only other reader is the launch restore, which can't tell that window from the one open at exit. ⌘Q still skips on `isTerminating`, and `testWindowCloseExitCapturesRunningCommand` still passes.

`agtermUITests/RestoreCommandUITests.swift:347` - `app.staticTexts` is app-scoped and the `window.new` window already has a seeded session row, so this assertion resolves against window 2 and its "window reopened" message is false. `window.select`'s own `pollUntil` is what waits here.
Fix: wait for `app.staticTexts["session-row"].count >= 2` before the drain.

i-kozlov:
While auditing the gate we hit three smaller things adjacent to this PR and deliberately left them out — flagging for your call (fold into this PR / follow-up issue / ignore):

1. The two capture sites read the toggle from different sources: `GhosttyApp.shared.restoreRunningCommand` in `willClose` (WindowAccessor.swift:155) vs `settingsModel?.settings.restoreRunningCommand` in `applicationWillTerminate` (AppDelegate.swift:319). They are in sync today via `SettingsModel.persistAndApply()`, but a future write that bypasses it would silently desync window-close capture from quit capture. Unify on one accessor, or leave as is?

2. `clearRestoreCommands` (ControlServer.swift:394) sweeps only `allOpenSessions()`, so a window closed mid-run keeps its captured argv in `windows/<id>.json`. After 4f4e528 that residue cannot replay anywhere (a mid-run reopen rewrites the snapshot, orphan recovery strips captures), so it is inert data on disk — but the command's "every captured foreground command" wording now slightly overpromises.

3. `Session.swift:186` doc comment still says foregroundCommand is "captured at the last clean quit" — predates the willClose capture point; the file is untouched by this PR, so we left it alone.

umputun (changes requested):
still open - the scrub substitution doesn't close round 2's hole, and the test point wasn't applied.

`agterm/Views/WindowAccessor.swift:155` - the capture still fires on every non-terminating close, and
neither scrub reaches a window that is closed and never reopened: `loadStore`'s write-back needs a mid-run
reopen, and `recoverOrphanedWindows` only runs when `windows.json` is missing or corrupt. Round 2's repro
runs end to end unchanged, and the residue survives arbitrarily long, a window closed weeks ago keeps its
argv until something reopens it.
Fix: scope the capture to the exit at that site, `library.openIDs() == [windowID]`. The closing window's
store is still loaded there, so it reads exactly "this close is the exit". Any alternative has to hold:
after a close that is not the exit, `windows/<id>.json` carries no
`foregroundCommand`/`splitForegroundCommand` by the time the store is dropped. An in-memory gate, or a
scrub that needs a later reopen, can't do that: the closed window has no store left and its file isn't
read again until a launch restore. Once the capture is scoped, both scrubs have nothing left to
compensate for.

`agtermUITests/RestoreCommandUITests.swift:347` - unchanged from round 2, and my prescription there was
wrong as written: `app.staticTexts["session-row"]` is a single element and has no `count`. The suite's own
form is `app.staticTexts.matching(identifier: "session-row").count`, which is 1 while only the surviving
window is open, so it can't be pre-satisfied.
Fix: poll that count for `>= 2` before the 3-second drain instead of `firstMatch.waitForExistence`.

the regression round 2 asked for is still missing, and it's the all-closed exit rather than the mid-run
reopen: window 1 runs a marker, close it and leave it closed, exit by closing window 2, relaunch, assert
window 1's marker did not fire and window 2's did.

on the three you flagged: 1, leave it, `SettingsModel.init` mirrors the flag into `GhosttyApp` before any
window exists, and it's moot once the capture is scoped. 2, not inert, that residue is the replay vector
above, and scoping the capture removes it at the source. 3, real, but it's on my list and I'll deal with
it, leave `Session.swift` alone.

i-kozlov:
Round 3 applied, plus one more thing our review sweep caught in the same area.

**Capture scoped to the exit** — the willClose capture now runs only under `!library.isTerminating && library.openIDs() == [windowID]`, exactly the site you named; a non-last close persists nothing. The mid-run-reopen UI test now also asserts a non-last close leaves no capture on disk (verified red without the guard). The two scrubs from the previous round stay as defense-in-depth for stale files written by earlier builds — with the capture scoped they have nothing new to compensate for, say the word and I'll drop them instead.

**UI assertion** — switched to polling `app.staticTexts.matching(identifier: "session-row").count >= 2` before the drain, per your correction.

**All-closed-exit regression test** — added as prescribed (`testAllClosedExitReplaysOnlyTheExitWindowsCapture`): window 1 runs a marker and is closed mid-session, exit via window 2, relaunch asserts window 1's marker did not fire and window 2's did.

**Found while making that test pass: the reopen fallback loses the exit window's replay.** Closing the last window persists `frontmost: nil` (`closeWindow` drops the store before `activeWindowID` is read — the mechanism you traced in round 2), so the next launch's never-windowless fallback opens `windows.first`, the oldest library entry, not the exit window. For any library where those differ, the close-the-last-window restore silently did nothing — your prescribed test's second assertion fails on that code. Fixed by pinning `frontmostWindowID` to the closing window when the open set empties (`WindowLibrary.closeWindow`), so the fallback reopens the exit window; covered by a unit test seeding a two-window index (red without the pin: the relaunch opened the oldest window and no replay happened).

On the three flagged items — settled per your calls: toggle reads left as-is, the residue point is moot with the capture scoped, `Session.swift` left alone.

Two pre-existing edges noted while auditing (both behind `windowClose`'s no-registered-window fallback, untouched by this PR — flagging, not fixing): that fallback calls `library.closeWindow` directly, so a last-window close through it skips the willClose capture entirely; and if it fires for a store whose NSWindow never attached (`window.new` racing `window.close`), the new frontmost pin would point at that phantom window for the next launch's fallback. Both need the same narrow race to matter — your call whether either deserves a follow-up.

umputun (commented):
no new findings, all three points from the last round check out.

pushed a merge of master to your branch and resolved the `WindowLibrary.swift` conflict. Master's 8fe39334 rewrote the same `closeWindow` frontmost lines while this was open; I kept your form, since it also covers a frontmost left nil or stale by `removeWindow` when the last close happens.

Issue #309, opened 2026-07-28
OSC 111 (reset background color) is ignored — TUIs leave the pane recolored after exit


skripalschikov:
agterm honors `OSC 11` (set background color) but not `OSC 111` (reset background to default), so any TUI that colors the background and restores it on exit leaves the session stuck with the app's color.

Hit it with [`sbx`](https://docs.docker.com/ai/sandboxes/) (Docker Sandboxes 0.37.0) — a Bubble Tea **v2** app, whose renderer emits `ansi.ResetBackgroundColor` (= `OSC 111`) in [`close()`](https://github.com/charmbracelet/bubbletea/blob/main/cursed_renderer.go). After quitting sbx the session keeps sbx's background instead of the theme's.

### Repro (no sbx needed)

In an agterm session — query the background, set it, reset it, query again:

```sh
bg() { old=$(stty -g); stty raw -echo; printf '\e]11;?\e\\' >/dev/tty; \
       dd bs=1 count=24 2>/dev/null | cat -v; stty "$old"; echo; }
bg; printf '\e]11;#123456\e\\'; bg; printf '\e]111\e\\'; bg
```

Expected: the third query reports the theme background again.
Actual: it still reports `rgb:1212/3434/5656`, and the pane stays visibly recolored.

Same result with the BEL terminator (`\e]111\a`, which is what `charmbracelet/x/ansi` actually sends).

### Scope — only the background reset is affected

Measured in one agterm session, and the same script in standalone Ghostty.app for comparison:

| sequence | agterm 0.15.3 | Ghostty.app |
|---|---|---|
| `OSC 10` set fg → `OSC 110` reset | restored | — |
| `OSC 11` set bg → `OSC 111` reset | **not restored** | restored |
| `OSC 12` set cursor → `OSC 112` reset | restored | — |

Ghostty's core does parse 110/111/112 ([`src/terminal/osc.zig`](https://github.com/ghostty-org/ghostty/blob/main/src/terminal/osc.zig)), so the guess is that agterm's interception of the OSC-11 color change (for the native pane background / the `session background color` feature) has no handler for the reset operation.

### Recovery

Does **not** help:

- `agtermctl session background color <#hex>` followed by `agtermctl session background clear`
- `agtermctl config reload`
- `agtermctl theme set <current theme>`

[...]

Issue #274, opened 2026-07-22
Session background watermark not rendered on the scratch terminal


skkap:
`agtermctl session background image …` sets a per-session watermark. It renders correctly on the **main pane and split panes**, but the **scratch terminal shows no watermark**. Since it's a per-session setting, I'd expect the scratch surface to show it too.

**Steps to reproduce**
1. Set a background image on a session: `agtermctl session background image ~/.config/agterm/backgrounds/mark.png --opacity 0.16`
2. Open the scratch terminal (`agtermctl scratch on`).
3. The scratch surface renders with no watermark; main and split panes show it fine.

**Expected:** the watermark also renders on the scratch terminal, consistent with the main and split panes.

**Environment:** agterm 0.15.3, macOS.

PR #372, opened 2026-08-05
feat(theme): pin any pane to its own theme

skripalschikov:
## The problem

agterm's theme is app-wide, so every surface renders the same palette. But panes are not interchangeable: the scratch is a throwaway shell, an overlay is a transient tool, and a split pane is usually a different task from its sibling. Both full-coverage surfaces — the scratch and a full overlay — replace the session's panes in place, so nothing on screen distinguishes "I am in my work shell" from "I am in the scratch" except remembering which one you toggled last.

`session background` is the closest existing tool and cannot do this. It sets a background image, text watermark, or solid color, leaving foreground, ANSI colors, cursor, and selection on the app-wide theme. It can tint a pane; it cannot make one read as light while the rest of the app is dark.

Nor is this reachable through a `keymap.conf` custom command. `theme set` is app-wide by construction — there is no existing per-surface path for a custom action to call, so unlike the even-split gesture in #341 this is not an alias for something already there.

## How this solves it

Per-surface config overlays already exist and already carry four things: the watermark, font zoom, `background-opacity`, and the OSC-11 opacity lift. This adds one `theme = …` line to the same `WatermarkConfig` builders and reuses `configWithOverlay` unchanged — no new mechanism, and libghostty resolves a conditional `light:X,dark:Y` per surface from the scheme each surface already records.

```bash
agtermctl session theme Dracula --pane right
agtermctl session theme "GitHub Light" --dark "GitHub Dark" --pane scratch
agtermctl session theme --clear --pane right
agtermctl session overlay open lazygit --theme Dracula --theme-dark Nord
```

The theme line is emitted **first** in the overlay, so a `.color` watermark's explicit `background` still wins over it and the existing precedence is unchanged.

The override belongs to the pane **slot**, not the surface, following `splitRestoreCommand`: it persists across restart, survives a scratch exiting and respawning, and a pane can be themed before it exists. Unlike `splitRestoreCommand` an override for an absent pane is kept rather than dropped, so a hidden pane's theme never becomes unclearable.

Pane vocabulary stays `left|right|scratch`, matching `session status --pane` and `session type --pane`. No `--pane all`.

## Convention checklist

- `Command` case `session.theme` in `ControlProtocol.swift`, with `ControlSessionThemeOptions`
- `ControlDispatcher.dispatchSessionTheme` owns parsing, validation and error text; the app does the side effect through `ControlActions.setSessionTheme`
- `agtermctl session theme` in `agtermctlKit`, plus `--theme`/`--theme-dark` on `overlay open`
- protocol round-trip tests and an end-to-end `agtermUITests/SessionThemeUITests`
- read-back on the `tree` node as `theme` / `splitTheme` / `scratchTheme`
- bundled skill (`SKILL.md`, `reference.md`) and `site/commands.html` + `site/docs.html` updated

No GUI and no keyboard shortcut. App-wide theming stays with Settings, View ▸ Select Theme…, the palette, and `theme set`; this is control-only on purpose, so it adds no Settings surface and no `InterfaceElement`.

## Two bugs found while building it

- **OSC-111 baseline.** A themed pane now resolves its default background from its own theme rather than the app's. Without it, a program that set and then reset its background in a pinned pane would strand it recolored — the #309 failure mode, reintroduced through a different door.
- **Inactive-pane mute wash.** `washColor(for:)` fell back to the app theme, so the wash over a pinned pane blended background→wrong-background and tinted it instead of fading its text. The same gap hit the floating overlay's panel backing, which matters more than it sounds: a theme-only overlay emits no `background-opacity` line, so under translucency the surface stays at the base config's zero and that SwiftUI backing is exactly what you see. Both overlay readers now share one `overlaySurfaceColor` helper.

## Verified

Left pane on the app theme reported `rgb:3030/3030/3030` while the Dracula-pinned right pane reported `rgb:2828/2a2a/3636`; a `--theme Dracula` overlay reported `rgb:2828/2a2a/3636` against `rgb:3030/3030/3030` for an untinted one.

Theme-picker live preview was checked specifically: `previewTheme` → `apply()` → `reloadConfig` broadcasts the shared config then calls `reapplySessionConfigIfNeeded()`, whose guard now includes `paneThemeOverride`, so a pinned pane re-asserts the way watermarks and zoom already do. Chrome colors still come from the app-wide config.

`swift test` 2374 pass · `make test-app` 155 pass · `make lint` clean · Debug build succeeds, all re-run after rebasing onto 6b74787.

## Notes for review

- **`agtermUITests/SessionThemeUITests` has not been run.** The XCUITest runner cannot bind its socket on my machine — an untouched `SessionTextUITests` fails the same way — so that suite is unverified here and needs your run.
- **Scope**: 34 files, but +447/-62 is source, +312 tests, +116 docs. The file count is the four mirror surfaces the convention requires, not breadth of change.
- **`AppStore.swift` is at 999/1000 lines.** It was at 996 before this branch. I moved `setThemeOverride` and the snapshot restore into `AppStore+Panes.swift` rather than raise the limit, but the file is one line from the ceiling and the next feature to touch it will have to split it.
- **Chrome cannot follow a pinned pane** — pin the scratch to light while the app is dark and the detail area goes light against dark sidebar and titlebar. Inherent to per-pane theming; flagging it as an accepted tradeoff rather than an oversight.
- **`WatermarkConfig` now builds per-surface config overlays generally**, watermarks being one input, so its name has drifted. I left the rename out to keep this diff focused.


umputun:
going to decline this one, but there's a smaller shape here worth talking about.

the premise doesn't hold as stated. It reads as "nothing separates the work shell from the scratch", and that was settled the other way already: #274 reported a session watermark not rendering on the scratch as a bug, and #275 fixed it by having scratch surfaces inherit the owning session's visual config. A scratch reading as its own thing is the behavior I removed on a user report.

to be clear about what the decline isn't: the `session background` argument in the body is accurate. It's session-scoped, has no `--pane`, and reaches only the background layer. That part is right, it just isn't the reason.

what I don't want is the mechanism. A 75th command, a new wire type, theme-name validation, a theme-file parse and hex cache, the light/dark variant, three persisted fields on `Session`, three on `SessionSnapshot`, three tree fields, and the four mirror surfaces to keep in sync. That's a lot of permanent API for pinning a palette, and `AppStore.swift` ends one line off the limit.

if the underlying want is real, the proportional version is `--pane` on the existing `session background`. No new command, no new type, and `BackgroundWatermark` already covers image, text and solid color, so all three modes get it at once. It has to be an explicit override with inherit as the default, or it re-breaks #274. Open an issue and we can settle the scope there first, same as I asked on #265. I'm not convinced yet that it's needed at all: left and right sit side by side with the inactive one already washed, so the scratch is the only case with anything behind it.

one thing separate from the decision, and it applies to any version of this. `closePrimaryPane` migrates cwd, title, foregroundCommand, restoreCommand and the pane overlays to the promoted survivor, but the branch adds no theme migration. Pin the right pane, let the left shell exit, and the survivor keeps rendering the pinned theme while `paneThemeOverride` reads the empty left slot. `session theme --clear --pane right` resolves to a nil surface and no-ops, and the stale `splitTheme` snapshots into the next split.

PR #383, opened 2026-08-07
docs(cookbook): add kiro-agent-status recipe

bitcldr:
Adds a `kiro-agent-status` recipe, per your call in #382 to send this against `cookbook/` rather than as a bundled adapter.

## Why a recipe and not a hook adapter

Kiro CLI is the one agent that can't reach the sidebar glyph through hooks, for two independent reasons I checked against kiro-cli 2.16.1 rather than its docs:

- Hooks are declared **per agent**, inside each `~/.kiro/agents/<name>.json`, with no global file. Merging into one covers one agent and gets overwritten by whatever manages the rest. A standalone global hook config is silently ignored in 2.x, and under `--v3` too.
- There is **no permission event**. `kiro-cli agent validate` accepts exactly five: `agentSpawn`, `userPromptSubmit`, `preToolUse`, `postToolUse`, `stop`. `preToolUse` fires with the same payload whether a tool was auto-approved or is sitting waiting on you, so nothing in the hook surface can tell `blocked` from ordinary work.

So this drives the glyph from the shell. A `kiro-cli`/`kiro` shell function starts a small poller, runs the real binary in the foreground, then kills the poller and clears the row when it returns — for any reason, including a crash or a Ctrl-C, since that's what job control already guarantees and it needed no code of its own. All three states come from the poller, which reads the pane's own recent text through `agtermctl session text`: `active` while kiro's working footer is up, `blocked` while the approval dialog is up, `completed` once a turn's markers are gone.

**Per turn, not per session.** I had the wrapper report `active` on entry at first, and following my own README caught it: `kiro-cli chat` is a REPL, so a turn ends when kiro returns to its prompt, long before the process exits. The row pulsed for the entire session, including a chat I opened and never prompted. The wrapper now only speaks for what it can actually see — its single status write is a plain `idle` after kiro-cli exits.

## Scope note

If you only want to know *whether* kiro is running, this recipe is unnecessary and the README says so up front: adding `kiro` to `AGTERM_AGENT_RE` in the shipped shell integration is one env var and no code. The delta here is everything that needs turn granularity, which that integration can't produce for a REPL — `blocked` while the dialog is on screen, `completed` at the end of each turn, and an `active` that stops pulsing in between. The two are alternatives rather than layers, and *Setup* says not to enable both, since the shipped integration pins one `active` from command start to the next shell prompt.

## Design points worth a reviewer's eye

- **`function kiro-cli { }`, not `kiro-cli() { }`.** The latter alias-expands in the name position while the file is being sourced, so a pre-existing `kiro-cli` or `kiro` alias makes sourcing a syntax error and breaks every new shell — and an alias pointing at kiro-cli is exactly the case this sets out to support.
- **A function, not a `preexec` regex.** `preexec` sees the literal text you typed, before alias or abbreviation expansion, so it misses `alis` even when `alis` expands to `kiro-cli …`. Function lookup happens after expansion and catches both.
- **`command "$real" "$@" && rc=0 || rc=$?`.** Under `set -e`/`ERR_EXIT` a bare call would exit the shell on a nonzero kiro-cli and skip the `idle`, stranding a blinking glyph. `if ! cmd` also survives errexit but loses the code — inside the branch `$?` is the negation's `0`.
- **Markers are matched as lines, not substrings.** Kiro talking *about* its own UI isn't kiro running. Ask it about this recipe and its answer quotes the footer mid-sentence; the real one occupies a line preceded by nothing but indentation or a spinner glyph. Substring matching pinned the row blinking on the first case, and raised a false `blocked` off an answer that described the dialog.
- **Turn-end is debounced over three frames.** Kiro's screen goes briefly clear of both markers *within* a turn — between two tool calls, and in the frame right after an approval is answered but before the footer returns. Calling one quiet frame the end fires the finish sound and drops the blink mid-turn. A failed pane read resets that streak too, since a frame never seen may well have held the footer.
- **A status write counts as delivered only if it succeeded.** Otherwise a `blocked` believed sent but never drawn sits wrong on the row until some unrelated later transition repairs it. Only a missing status script is detectable — the stock one ends in `|| true; exit 0` so a hook can't break a turn — and retrying next tick while the classification still holds costs one extra call and needs no read-back.
- **Two stop conditions, no check budget.** `kiro-cli chat` can sit at a prompt all day, so a fixed cap would stop reporting mid-session. The poller exits when its spawning shell dies (`kill -0 $PPID`), which also covers a hard-killed app since the pty takes the shell down with it, and after 10 consecutive failed pane reads, since no app means no socket. Neither normally runs — the wrapper kills it as soon as kiro-cli returns.
- **`--pane` on the read, none on the write.** `session text` with no `--pane` reads whatever is on screen (the focused pane, or a scratch terminal covering it), not necessarily kiro's pane. The write goes through the stock `agterm-agent-status.sh`, which already forwards `AGTERM_PANE` and `AGTERM_PANE_ID`, and that token is what corrects a stale role after a pane is promoted.
- **`agtermctl` is a real requirement, not just the hooks package.** The read side needs it by name, and the hooks installer bakes its CLI path into the script it writes as a *local* default without exporting it — so hooks-only, which is what Setup asks for, gives you working writes and a dead read side. The poller falls back to the bundled binary in `~/Applications` and `/Applications`; *Requirements* states the constraint and the `AGTERMCTL` override.
- **Leaving `blocked` is driven by the dialog disappearing**, not by the footer coming back. Requiring the footer stranded `blocked` on the No path, which returns kiro straight to its prompt with nothing in between — and a false `blocked` is the one wrong state that actively summons you to a session that doesn't need you.

## Version floor

`0.5.0`. `session text` and the `--pane left|right` this reads with both shipped together in #46; only the later `--pane scratch` value, which this doesn't use, needed 0.7.0. (I said 0.7.0 in the discussion — thanks for the correction.)

## Testing

Manually on kiro-cli 2.16.1 in zsh, bash, and fish, plus scripted harnesses driving the poller through synthetic screen sequences rather than only the happy path: multi-turn REPL sequences, mid-turn marker gaps and the answered-dialog frame (no spurious `completed`), quoted-marker screens (no pinned blink, no false `blocked`), a status script that fails its first writes (retried, not latched), interleaved read failures, alias routing, errexit exit-code preservation, the blocked→active and blocked→No transitions, poller-killed-mid-run (the wrapper still clears the row), zero orphaned pollers after teardown in all three shells, and a pty-topology check confirming the shell does die with its master holder, which is what makes the `$PPID` watch valid.

Local gates green across the cookbook tree: layout and index parity both directions, headings, shebangs, file modes, `shellcheck`, `zsh -n`.

## One thing to flag about CI

**`kiro-agent-status.fish` gets no automated check.** The cookbook job lints `*.sh` with shellcheck, `*.zsh` with `zsh -n`, and `*.py` with ruff; there's no `.fish` step, and fish isn't on the `ubuntu-latest` image. I ran `fish -n` by hand (fish 4.8.1) and hand-tested the file, but nothing in CI will catch a future regression in it. Happy to drop the fish variant if you'd rather not carry an unlinted file, or to add a `fish -n` step in a separate PR if you want the gate — I didn't want to touch CI inside a cookbook contribution.

Refs #382


umputun (commented):
one blocker, rest are notes.

**`cookbook/kiro-agent-status/kiro-status-detector.sh` is committed 100644.** Both wrappers exec it directly (`.zsh:37`, `.fish:35`) and *Setup*'s `cp` doesn't add the bit, so on a reader's machine the launch fails with EACCES into `>/dev/null 2>&1` and the only status that ever fires is the wrapper's closing `idle`. CONTRIBUTING names this case: "a perfectly good script committed non-executable passes every check and then fails on the reader's machine with permission denied". Fix: `git update-index --chmod=+x cookbook/kiro-agent-status/kiro-status-detector.sh`. Worth guarding the start too, `[ -x "$KIRO_STATUS_DETECTOR" ]` before launching and before the `kill`, because fish leaves `$last_pid` untouched when a launch fails, so the `kill` targets whatever background job ran before it in that shell.

the rest, none blocking:

`README.md:15` - the 0.5.0 floor doesn't hold up. `--pane scratch` is used: the detector defaults `AGTERM_PANE` only when unset and forwards it verbatim, so a kiro-cli started in a scratch terminal sends it, and 0.5.0 rejects the value. `AGTERM_PANE` itself isn't injected until 0.7.1 (#130), so on 0.5.0-0.7.0 the detector falls back to `left` and can't read a right-split kiro. The `AGTERM_PANE_ID` claims at `:102` and `:119` need 0.13.0 (#213). Fix: base floor 0.7.1, with the two token sentences qualified as 0.13.0.

`README.md:119` - the promoted-pane bullet is wrong both ways. `session text --pane right` with no split returns `ok:false` and agtermctl exits nonzero, so the reads don't "return nothing", they fail, they don't count as quiet frames, and the poller breaks out at `KIRO_STATUS_MAX_READ_FAILURES` about five seconds in. It isn't only `blocked` that stops, the poller is gone, so a pulsing `active` or a `blocked` sits there until kiro-cli exits. Fix: rewrite the bullet. A cleanup write on the way out would land, since the stock script forwards `AGTERM_PANE_ID`.

`README.md:11` and `:88` - job control does not cover Ctrl-C. Reproduced under a pty in bash 5.3 and zsh 5.9: a terminal SIGINT that kills the foreground child aborts the rest of the function body, so `kill "$pid"` and the `idle` never run and the disowned poller keeps going for the life of the shell. Narrow in practice, kiro-cli in raw mode swallows Ctrl-C itself, but the claim is unconditional. Fix: soften both sentences and add the orphan to *Limits*.

`kiro-status-detector.sh:100-105` - the anchor comment names "this recipe in a pager, a git diff of it" as cases it rejects, and it rejects neither: `^[^[:alpha:]]*` passes any quote, bullet, diff marker or line-number prefix. Lines 102 and 108 of the detector match its own working pattern, as do lines of its own diff. Only the mid-sentence prose case holds. Fix: drop the two examples rather than tighten the anchor, false negatives are the worse direction here.

`kiro-agent-status.fish:1` - no shebang. CONTRIBUTING says every script carries one, and `claude-session-resume/claude-resume.fish` has `#!/usr/bin/env fish` despite also being sourced and 644. CI's shebang check globs only `.sh/.zsh/.py`, so nothing catches it.

`kiro-agent-status.zsh:1` - `#!/usr/bin/env bash` on a `.zsh` file. The other five cookbook `.zsh` files use `zsh`. The extension is right, `${(%):-%x}` at `:23` would red shellcheck, so it's just the shebang.

on the fish variant: keep it. An unlinted file is worth more than no fish support, and adding a `fish -n` step is my problem, not yours.

bitcldr:
Thanks for the detailed review - every point is addressed in the latest push.

umputun (approved):
lgtm, all seven addressed. Checked each against the head rather than taking the summary for it: the detector is 100755 with the `-x` guard before the launch and the argv check before the kill, the floor is right (#130 is 0.7.1, #117 is 0.7.0, #213 is 0.13.0), the promoted-pane bullet now says the reads fail and the poller gives up, the SIGINT exception is in How it works, the zsh header and Limits, and both shebangs are fixed.

the anchor comment came back better than what I asked for. I said drop the two examples; naming which false positives survive, these files in a pager included, is the more useful version.

PR #385, opened 2026-08-07
feat(sidebar): copy a row's name from the context menu

skkap:
## What is the problem?

A workspace's name cannot be got out of the app, and a session's cannot be got out of the sidebar. The row's `NSTextField` is not selectable outside rename mode, so reading a name to reuse elsewhere means retyping it, or entering rename mode and copying out of the field — which risks committing an edit to a name you only wanted to read.

For **session** rows a `keymap.conf` line already does it, and an earlier version of this description was wrong to claim otherwise:

```
command "Copy Name" cmd+shift+c printf %s "$AGT_SESSION_NAME" | pbcopy
```

`{AGT_SESSION_NAME}` is `Session.displayName` (`CustomCommandRunner.swift:265`), the property the row's name comes from, and right-clicking a session row outside the selection selects it (`WorkspaceSidebar.swift:775`), so the clicked session is active by the time an action runs.

For **workspace** rows there is no equivalent. That same guard is `node.kind == .session`, so right-clicking a workspace row changes no selection, and `{AGT_WORKSPACE_NAME}` resolves the active session's workspace (`CustomCommandRunner.swift:268`, via `store.workspace(forSession:)`) — which can be a different one. A workspace name can only be read by eye.

## How does this solve it?

**Copy Name** under Rename, on session and workspace rows. Single-target, gated on `sessionCount == 1` like Duplicate Session and Reveal in Finder, so it is absent under a multi-selection.

The workspace half is the part not reachable today, so it carries the feature. The session half is there for reach: a `keymap.conf` line only helps people who read the keymap docs.

It copies `Session.displayName`, so an un-renamed session yields its title or cwd basename rather than the empty `customName`. Names resolve through the sidebar's window-local store, like Close/Flag/Duplicate, so a background window copies its own row. A blank workspace name and a row closed between the right-click and the choice both leave the pasteboard untouched rather than clearing it.

`workspaceName` sits in `agtermCore` because the blank-to-nil gate is a model policy; the session case is the existing `session(withID:)` plus the published `displayName`, with the app target doing only the pasteboard write.

## Control API

None, on your ground rather than mine — the socket never writes the user's clipboard. `session copy` returns the selection so it does not have to, `session paste` reads it as input, and a form that returned the name would duplicate what `tree` carries. Written into `.claude/rules/control-api.md` as a rule rather than an argument to re-make: **an action whose only effect is a system-pasteboard write gets no control command.** `sidebar.md` cross-references it.

## Docs

`README.md`, `site/docs.html`, `.claude/rules/sidebar.md`, `.claude/rules/control-api.md`. `CHANGELOG.md` left alone. `plugins/agterm/skills/agterm/` and `site/commands.html` unchanged — both document control commands only, and there is no command here.

## Verification

`make build` succeeds · `make test` 2377 passed · `make lint` clean · `make test-app` 168 tests, one failure: `LiveMenuKeyEquivalentsTests.testUppercaseKeyEquivalentReportsImpliedShift()`, **pre-existing** and reproduced on unmodified `master` (macOS 26.5.2 / Xcode 26.6).

Six tests in `agtermTests/SidebarCopyNameTests.swift` drive a real outline view and Coordinator: the renamed and un-renamed session names, the workspace name, absence under a multi-selection, and the two cases that must leave the clipboard untouched — a blank workspace name, and a row closed between building the menu and choosing from it.


umputun:
your self-review covers the control-API half. The keymap half of the description still stands as written, and it doesn't hold for session rows.

`{AGT_SESSION_NAME}` is `Session.displayName`, the same string the row shows. `CustomCommandRunner.swift:265` puts `session.displayName` into the context and `WorkspaceSidebar+RowRendering.swift:214` renders that same property. The pane's OSC title is one step of that chain, not the whole of it (`Session.swift:383-389`): a manual rename wins over the title, and with neither the cwd basename does. So a renamed session yields its rename through the token, and an un-renamed one is never empty.

right-click does select the row. `WorkspaceSidebar.swift:775-777` calls `selectRowIndexes` when the clicked session row sits outside the current selection, which fires `outlineViewSelectionDidChange` (:654), which calls `store.selectSession` (:677). The right-clicked session is the active one by the time a menu action runs.

so one keymap line already does the session case:

```
command "Copy Name" cmd+shift+c printf %s "$AGT_SESSION_NAME" | pbcopy
```

what survives is narrower than the description argues:

- workspace rows. The guard at `WorkspaceSidebar.swift:775` is `node.kind == .session`, so right-clicking a workspace row changes no selection, and `{AGT_WORKSPACE_NAME}` resolves the active session's workspace (`CustomCommandRunner.swift:267`), which can be a different one.
- the batch. Tokens resolve one session, so one-name-per-line has no equivalent.
- being a menu item rather than a chord each user writes himself.

restate the case on those three and I'll take another look.

two things from the rework itself:

- dropping the two methods moved name resolution the wrong way. `CLAUDE.md` puts model and derivation in `agtermCore` and keeps the app target a side-effect adapter, continuing the #78 hoist series, so `store.workspaces.first { ... }` plus a blank-name filter inside a view goes against that. Owing a control command is not a reason to un-hoist.
- README still says Copy Names is "the first item on the menu" for a multi-selection. `WorkspaceSidebar+ContextMenu.swift:63-71` puts Clear Statuses and a separator above it whenever a selected session is non-idle.

skkap:
Sorry, deleted my earlier comment while tidying without realising you were mid-review — folded it into the description instead.

skkap:
Thanks — that was a useful correction, and all four points are addressed. Out of draft now.

**The keymap argument.** You're right, and I'd verified it the wrong way round: I'd read `menu(forRow:)` and never `menu(for event:)` directly above it. `{AGT_SESSION_NAME}` is `displayName` (`CustomCommandRunner.swift:265`) and the right-click selects the row (`:775`), so your one-liner covers the single-session case. The description no longer claims otherwise — it now rests on the three gaps you named: workspace rows (that guard is `node.kind == .session`, and `{AGT_WORKSPACE_NAME}` resolves the active session's workspace via `store.workspace(forSession:)`), the batch, and discoverability.

**The un-hoist.** Reverted. `sessionDisplayNames` / `workspaceName` are back in `AppStore+Naming.swift` with the blank-name gate in core, and the handler is two lines again. You're right that owing a control command was the wrong reason to move a derivation into a view.

**README.** Fixed — Clear Statuses and a separator sit above Copy Names whenever a selected session is non-idle, so it isn't the first item there either.

Two more I found while rechecking my own claims: I'd cited `:267` for `{AGT_WORKSPACE_NAME}` (it's `:268`), and I'd said `session.duplicate` had no read path, which is wrong — `tree` is its read-back. The control-API section now argues capability instead, and says plainly that it's yours to overrule.

Also tightened the tests: the ordering one passed for the wrong reason (it fed ids already in sidebar order, so raw order satisfied it too), and there's now one covering the composition the `\n` join depends on — the sanitize sites are each tested alone but were never tested together.

`make build`, `make test` (2377), `make lint` clean. `make test-app` is 169 with one failure, `LiveMenuKeyEquivalentsTests.testUppercaseKeyEquivalentReportsImpliedShift()`, which reproduces on unmodified `master` here — macOS 26.5.2 / Xcode 26.6, so possibly version-specific rather than something you'd see.

umputun:
go ahead with it, in the single-target form only.

drop **Copy Names**. Nothing needs a list of session names, and matching how Flag/Close/Move treat a batch is not a reason on its own. Gate it on `sessionCount == 1` next to Rename, Duplicate Session and Reveal in Finder.

keep both row kinds. The workspace half is the only part not reachable today, so it earns the item; the session half is there for reach, since a `keymap.conf` line only helps people who read the keymap docs.

no control command, though not on the capability argument. The socket does not write the user's clipboard: `session copy` returns the selection precisely so it does not touch it (`README.md:315`), and `session paste` reads the clipboard as input to a session. A `copy-name` that writes `NSPasteboard.general` would be the first command to break that, and one that returns the name is `tree`'s `name` field a second time. Neither shape is worth having. Put that in `.claude/rules/control-api.md` rather than sidebar.md, as a rule the next person can apply without re-arguing it: an action whose only effect is a system-pasteboard write gets no control command.

dropping the batch should also take out the newline join and what hangs off it, the sanitize composition test, the `copy-name operates on all targets` line in sidebar.md, and the "moves up under Clear Statuses" sentence in README.

one process thing. This was hard to follow from my side: opened ready for review, then pushed changes while I was reading it, then moved to draft, then kept discussing inside the draft. I could not tell whether draft meant work in progress and I should stay out of it, or that you still wanted my opinion.

pls keep it in draft while you iterate and mark it ready when you want eyes on it. I'll do the full pass then.

skkap:
Perfect, thank you! Moved to draft until ready for review.

skkap:
Addressed.

- **Copy Names removed.** Gated `sessionCount == 1` beside Duplicate Session and Reveal in Finder, so the item is absent under a multi-selection.
- **Newline join removed**, and what depended on it: the sanitize-composition test, the `copy-name operates on all targets` line in `sidebar.md`, the Clear Statuses sentence in README.
- **`sessionDisplayNames` removed** — a single row reads `session(withID:)?.displayName`. `workspaceName` stays for the blank-to-nil gate.
- **Both row kinds kept.**
- **Pasteboard rule added to `.claude/rules/control-api.md`**: an action whose only effect is a system-pasteboard write gets no control command. `sidebar.md` cross-references it, and Copy Name joined the single-target list there.
- **Tests: six.** The two batch ones removed, one added pinning the gating.
- Description rewritten.

`make build`, `make test` 2377, `make lint` clean. `make test-app` is 168 with the pre-existing `LiveMenuKeyEquivalentsTests.testUppercaseKeyEquivalentReportsImpliedShift()` failure.

umputun (approved):
lgtm

PR #388, opened 2026-08-07
cookbook: add same-dir recipe for syncing working directory across split panes

vladislav-yevtushenko:
## What it does

Adds a new recipe `same-dir` to the cookbook:
- Syncs the active pane's working directory (`$AGT_SESSION_PWD`) to the non-active opposite split pane (`cd <cwd>`).
- Bidirectional navigation: left pane -> right pane, right pane -> left pane.
- Auto-splits if no split is currently shown.
- TUI protection: checks `.foreground` / `.splitForeground` from `agtermctl tree --json` and aborts with an `agt notify` banner if the target pane is running a foreground process (such as `claude`, `lazygit`, or `nvim`).


umputun (commented):
commands, flags and JSON paths all check out against the current control API, and the 0.10.0 pin is right. Five things, only the first blocks the merge.

`cookbook/same-dir/same-dir.sh:63` - `printf 'cd %q\n'` reds cookbook CI. The `#!/bin/sh` shebang puts shellcheck in sh dialect where `%q` is SC3050, and that job has to come back clean. It runs fine as is, so this is the lint gate and nothing else. Fix: rename it `same-dir.zsh` with a zsh shebang, which moves it to CI's `zsh -n` check and keeps `%q`, or drop `%q` and single-quote the path yourself.

`cookbook/same-dir/same-dir.sh:33` - the TUI protection is off for a hidden split. `split` in the tree is `isSplit` (shown), not `hasSplit` (alive), and `session split off` hides the pane while keeping its shell and surface. So a hidden right pane running `claude` takes the `.split == false` branch, `BUSY` stays empty, the split is re-shown and `cd` goes straight into claude. Fix: read `.splitForeground` in that branch too rather than treating a hidden split as no split.

`cookbook/same-dir/same-dir.sh:63` - typing right after `session split on` races the pane. `toggleSplit` only flips the flag and the split surface is created lazily on the next layout pass; `session type --pane right` does not poll for realization the way the main pane does, it fails fast with `session has no split pane` or `session not realized`. Under `set -eu` with both streams on `/dev/null` the chord opens a split and then silently does nothing. Fix: retry the type a few times, and send it through `fail` so a real failure still gets a banner.

`cookbook/same-dir/README.md:15` - Requirements names 0.10.0 but not what shipped in it, which CONTRIBUTING asks for and the other recipes do. For reference: 0.10.0 is where custom commands began exporting `AGT_*`, and where `tree` gained `splitFocused`, `foreground` and `splitForeground`. `session type --stdin --pane` is older.

`cookbook/README.md:19` - the blank line after the new row ends the markdown table, so every row below it renders as literal pipe text. The index check greps `^|` and stays green, so CI will not catch it.

umputun (changes requested):
three of the five are fixed: the shebang and `%q`, the hidden-split check now reading `.splitForeground`, and the table row. One blocker left, and it is in the retry itself.

`cookbook/same-dir/same-dir.zsh:77` - `(( tries++ ))` under `set -eu` ends the script on the first pass. Post-increment evaluates to the OLD value, so with `tries` at 0 the arithmetic command returns 1 and `set -e` takes it. The loop therefore makes exactly one attempt, and when that one loses the race it exits with both streams on `/dev/null` and no banner - the silent no-op the retry was added to remove. Reduced to the bare loop it prints nothing past the first iteration and exits 1; `tries=$(( tries + 1 ))` runs all ten. Fix: use that form.

worth knowing why nothing caught it: renaming to `.zsh` was my suggestion, and it moved the file from shellcheck to `zsh -n`, which parses and does not lint. CONTRIBUTING says as much - read a `.zsh` recipe over yourself, since CI will not.

`cookbook/same-dir/README.md:15` - still open from last time. Requirements names 0.10.0 but not what shipped in it, which CONTRIBUTING asks for in the same line. 0.10.0 is where custom commands began exporting `AGT_*`, and where `tree` gained `splitFocused`, `foreground` and `splitForeground`.

the conflict is mine, not yours. I regrouped the cookbook index into four tables a few minutes ago, which rewrote the rows yours sits among. On rebase your row goes under **Panes, pickers and input**, last, after `overlay-and-split`. Nothing else about the change is affected.

umputun (approved):
lgtm

PR #390, opened 2026-08-07
fix: cap the substring band under the subsequence floor (#371)

### LINKED ISSUE #371: Fuzzy ranking: a literal substring match can rank below a non-match — the substring band is unbounded
`fuzzyScore`'s three bands are documented at `agtermCore/Sources/agtermCore/Fuzzy.swift:5-6`: "an exact prefix is `0`, a substring `5 +` its start offset, a scattered subsequence `40 +` the length gap." The ordering that comment implies — prefix < substring < subsequence — only holds for short targets, because the substring band is unbounded: `5 + offset` crosses 40 as soon as the match starts past character 35, at which point a row that literally contains the query ranks below a row that only matches by scattered letters.

## Repro

Verbatim copies of `termScore`/`fuzzyScore`/`fuzzyRank` from master (`6b74787`), query `stem`, two rows a `pick` list would plausibly contain:

```
score=52  /Users/dev/Code/monorepo/services/audio-assets/stems
score=46  /opt/steam
ranked:
  1. /opt/steam                                            <- no "stem" substring at all
  2. /Users/dev/Code/monorepo/services/audio-assets/stems  <- the directory literally named stems
```

`/opt/steam` scores as a subsequence (s·t·e·a·m ⊇ s,t,e,m): `40 + (10 − 4) = 46`. The stems path scores as a substring at offset 47: `5 + 47 = 52`. The user typed the exact name of the second row and got the first.

## Why it surfaces

The sole non-test caller is `agterm/Views/Palette.swift:163`, which ranks caller-supplied `agtermctl pick` rows — and the cookbook's own `native-dir-picker` recipe feeds it absolute paths from a `find` over a tree of varying depth, exactly the mixed-length input that trips the bands. The built-in palettes use short labels, which is probably why this has gone unnoticed.

## Why the test suite can't see it

`FuzzyTests.swift:33-39` (`prefixBeatsSubstringBeatsSubsequence`) draws all three scores from the same 11-character target (`"New Session"`), so the bands can never overlap inside it. The defect is cross-target ranking; the test only asserts within-target ordering.

## Possible fixes

Your call on the shape — two options that keep every existing test green:

- clamp the substring band so it can't reach the subsequence floor: `5 + min(offset, 34)`;
- or scale the offset into the band's width so deeper offsets still order among themselves.

Happy to send a PR with a cross-target regression test if either direction works for you.

x9x9x9x9x9x91:
Fixes #371.

`termScore`'s substring band was unbounded (`5 + offset`), so a literal substring in a long target could score past the subsequence floor of 40 and rank below a scattered match on another row. Clamped to `5 + min(offset, 34)` as discussed — a substring now lands at 39 at worst, under any subsequence score.

Per your notes on the issue: the `fuzzyScore` doc comment now states the cap, the tie at 39 for substrings starting at or past offset 34 (falling to the caller's tie-break), and scopes the guarantee per term — summed multi-term scores can still cross bands, and the doc no longer implies otherwise.

Two tests: the cross-target regression drives `fuzzyRank` with the probe pair from the report (`/opt/steam` against a literal `stem` at offset 43) and asserts the ranking — unclamped it fails with the report's exact 48-vs-46 scores; a second test pins the 39 tie for deep offsets.

Full suite (2405 tests) + `make lint` strict green.

### DIFF
--- a/agtermCore/Sources/agtermCore/Fuzzy.swift
+++ b/agtermCore/Sources/agtermCore/Fuzzy.swift
@@ -2,8 +2,10 @@
 /// The query splits on whitespace into terms (`"cap dev"` is two); EVERY term must match `target` and the
 /// score is their sum, so term order is irrelevant — `"cap dev"` and `"dev cap"` both match `caprica-dev`.
 /// An empty or whitespace-only query matches everything at `0`, keeping the unfiltered list's natural
-/// order. Case-insensitive. Per term: an exact prefix is `0`, a substring `5 +` its start offset, a
-/// scattered subsequence `40 +` the length gap.
+/// order. Case-insensitive. Per term: an exact prefix is `0`, a substring `5 + min(offset, 34)`, a
+/// scattered subsequence `40 +` the length gap. The cap keeps a single term's literal match (`39` at
+/// worst) ahead of any subsequence-only match; substrings starting at or past offset `34` tie and fall
+/// to the caller's tie-break. Summed multi-term scores can still cross bands.
 public func fuzzyScore(query: String, target: String) -> Int? {
     let terms = query.lowercased().split(whereSeparator: \.isWhitespace).map(String.init)
     guard !terms.isEmpty else { return 0 }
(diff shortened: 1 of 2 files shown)


umputun (approved):
lgtm

PR #391, opened 2026-08-07
docs(cookbook): add the kimi-session-resume recipe

x9x9x9x9x9x91:
Adds the missing member of the session-resume family: each tab reopens its own Kimi Code conversation after a restart.

The mechanism differs from the three existing recipes. Kimi Code has lifecycle hooks of its own, and its SessionStart hook receives the new conversation's `session_id` on stdin — so instead of a shell function wrapping the launch, a hook script pins the tab's restore command with `agtermctl session restore "kimi -r <id>"`. No wrapper, no mapping file, `kimi` itself is never shadowed; the pin is rewritten on every start, so it always points at the tab's current conversation. That makes the index sentence about session-resume recipes never calling the CLI wrong by one — reworded to scope it to the three shell-function ones.

The script guards the agent-pipeline case — a kimi worker spawned by a Claude Code session in the same pane would steal the pane's pin from the conversation that owns it, detected via `CLAUDE_CODE_SESSION_ID` and skipped — and carries the owning process's `-m` model flag into the pinned line, validated against a safe character set rather than quoted, since the pinned line is shell code.

Floor is 0.16.0: `session restore` including `--pane-id` and the sticky per-pane override shipped there (#271). Tested on kimi-code 0.31.0. The committed script is the sanitized twin of the copy that has been driving my own sessions.

shellcheck clean; the cookbook layout/index/headings check passes locally.

### DIFF
--- a/cookbook/README.md
+++ b/cookbook/README.md
@@ -19,6 +19,7 @@ Recipes come from other people as well as the maintainer. Every one is reviewed 
 | [claude-session-resume](claude-session-resume/) | each tab reopens its own Claude Code conversation after a restart | 0.3.1, zsh or fish, Claude Code |
 | [codex-session-resume](codex-session-resume/) | each tab reopens its own codex conversation after a restart | 0.3.1, zsh, Codex CLI |
 | [opencode-session-resume](opencode-session-resume/) | each tab reopens its own opencode conversation after a restart | 0.3.1, zsh, opencode |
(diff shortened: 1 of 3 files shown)


umputun (approved):
lgtm

one note for whenever you next touch the file, not a fix request: the `-m` value gets a charset filter because it becomes part of the pinned shell line, but `$session_id` comes from the same stdin JSON and is interpolated into that same line unfiltered. Nothing downstream catches it either, since `session restore` rejects control characters but allows shell metacharacters. Real kimi ids are tame, so this costs nothing today. The same `case` guard would cover it.

PR #397, opened 2026-08-08
feat: nested (child) sessions

sadensmol:
Lets a session be created as a child of another and stacked as an indented
sub-tree under its parent in the sidebar, at arbitrary depth. Headline case is
agent-driven: `agtermctl session new --parent "$AGTERM_SESSION_ID"` groups a
skill's helper sessions instead of scattering them as flat siblings. Reverses
the "no extra tree depth" stance in discussion #293 for this use case.

- Adjacency model (`Session.parentID`) + contiguity invariant → array order ==
  visual tree order, so every flat consumer (navigation, flagged view, focus,
  persistence, recency, reorder) is unchanged. `SessionTree` math,
  `AppStore+Nesting` ops.
- Control/CLI: `session new --parent`, `session move --parent/--unparent`,
  `--after/--before` parent inheritance, `session collapse/expand`; `tree`
  `parentID`/`collapsed` read-back. 74 → 76 commands.
- GUI: nested outline with collapse/fold + roll-up badge/status on a collapsed
  parent, drag-to-reparent (auto-expands the new parent), New Child Session
  menus, cascade-close confirm with descendant count and grouped undo/Reopen.
- Backward compatible: optional snapshot fields, no version bump, non-nested
  trees serialize byte-identically.

Keep-in-sync surfaces updated alongside the code:
- Bundled agterm skill — `plugins/agterm/skills/agterm/` `SKILL.md`,
  `reference.md`, `examples.md` (the source the Help ▸ Install bakes into the
  installed Claude/Codex copy): `--parent`/`--unparent`, `collapse`/`expand`,
  `parentID`/`collapsed` read-back, cascade close, and the command count.
- Engineering rules — `.claude/rules/control-api.md`, `.claude/rules/sidebar.md`.
- `README.md` and `site/commands.html` + `site/docs.html` (count 74 → 76 and the
  new commands/args).

`swift test`, `make build`, `make test-app`, `make lint`, and
`NestedSessionsUITests` all green.

### DIFF
--- a/.claude/rules/control-api.md
+++ b/.claude/rules/control-api.md
@@ -83,14 +83,14 @@ paths:
 
 ## Public catalog
 
-There are 74 public commands:
+There are 76 public commands:
 
 - `tree`, `events.read`
 - `workspace.new`, `.rename`, `.delete`, `.select`, `.move`, `.focus`, `.filter`, `.collapse`, `.expand`
 - `session.new`, `.duplicate`, `.close`, `.select`, `.rename`, `.reveal`, `.move`, `.type`, `.split`,
   `.scratch`, `.focus`, `.resize`, `.go`, `.copy`, `.paste`, `.selectall`, `.text`, `.search`, `.status`,
-  `.flag`, `.seen`, `.restore`, `.background`, `.overlay.open`, `.overlay.close`, `.overlay.resize`,
-  `.overlay.result`, `.hud.open`, `.hud.update`, `.hud.close`
+  `.flag`, `.seen`, `.collapse`, `.expand`, `.restore`, `.background`, `.overlay.open`, `.overlay.close`,
+  `.overlay.resize`, `.overlay.result`, `.hud.open`, `.hud.update`, `.hud.close`
 - `surface.zoom`, `dashboard`, `pick.open`, `pick.result`, `pick.cancel`
 - `quick`, `quick.type`, `quick.text`
 - `sidebar`, `sidebar.mode`, `sidebar.expand`, `sidebar.collapse`, `notify`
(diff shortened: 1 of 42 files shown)


umputun:
going to decline this one on the model, but there's a bug at the end that applies to any version of it, so read that part either way.

sessions inside a workspace are peers, and the workspace is already the thing that groups them. Nesting a session under another session adds a second grouping axis to the same tree and asserts an ownership relation that isn't generally true: a helper shell isn't owned by the shell that started it, it just happened to be started from there. #293 was a level above workspaces and this is a level below, so it isn't literally the same call, but the reason I gave there is the reason here. Same as #259 and #265.

to be clear about what the decline isn't. The contiguity approach is a real answer to the obvious objection, and keeping `Workspace.sessions` in preorder so navigation, flagged view, focus, recency and persistence read it unchanged is the right instinct. The mirror surfaces are all updated, count included. None of that is why it's a no.

the permanent surface is. Two more commands, `parentID` and `collapsed` on `Session` and `SessionSnapshot`, two more tree fields, new drag semantics, a cascade close with its own confirm, a new recent-closed kind with its own undo shape, and from here on every mutation path has to preserve an invariant that nothing enforces. That last part is the one I'd carry forever, and it's the same shape as what I said on #372.

for the underlying want, grouping an agent's helper sessions under the caller, a workspace per run already does it over the existing API: `workspace new "<run>" --collapsed`, then `session new --workspace <ws> --no-select` per helper, and `workspace delete` is the cascade close. `cookbook/park-and-resume/agt-resume.sh` builds exactly that shape. If that's genuinely not enough, an issue is the place to settle it before a PR this size, same as I asked on #265 and #372.

separate from the decision, and it applies to any version of this. The branch can delete live sessions off disk. `SessionTree.preorder` only walks from nil-parent roots, so a session whose parent is missing from its workspace never appears in the returned order, and `AppStore+Nesting.swift:66` and `:86` both then do `workspaces[i].sessions = order.compactMap { byID[$0] }`. That's a filter, not a permutation: the session is dropped with no teardown, no recent-closed record, and the next `save()` writes the loss. `repairContiguity`'s `guard order != before` doesn't catch it, since order differs precisely because the session went missing.

three paths get a workspace into that state, none of them exotic:

- `moveSessions` (`AppStore.swift:562`) was never migrated. It removes one element per id, never carries a subtree, never clears `parentID`. The sidebar's Move to menu calls it even for a single row, so right-clicking a parent and moving it strands the children pointing into another workspace.
- `duplicateSession` (`AppStore+Duplicate.swift:12`) inserts at `index + 1` with a nil parent, splicing a top-level row between a parent and its children. It was correct before this branch, the new invariant is what makes it wrong.
- `restore(from:)` copies `parentID` out of the snapshot with no check that it names a live session in the same workspace.

two smaller ones nearby. `session move --parent` returns `ok: true` when `reparentSession` silently no-ops on a cross-workspace anchor or a cycle, while the docs added in this branch say both are rejected. And the default close path (grace undo on) records one recent-closed item per subtree member instead of the group, so once the 3s grace expires, Reopen restores a single child whose parent is gone and which the sidebar can't render.

Issue #142, opened 2026-07-05
Voice dictation in Claude Code never gets microphone permission (missing NSMicrophoneUsageDescription)


dnncomp:
### Environment
- agterm version: 0.8.0 (build `cd5b0c6`)
- macOS: 26.5 (build 25F80)
- Bundle ID: `com.umputun.agterm`
- Install path: `/Applications/agterm.app`
- Code signature: valid, notarized (Developer ID Application: Brave Elk LLC, Team ID `H7K73622CK`)

### Problem
Voice dictation (`/voice` in Claude Code running inside an agterm session) never triggers the macOS microphone permission dialog, and agterm never shows up in **System Settings → Privacy & Security → Microphone**, even after:

- Fully quitting and relaunching agterm
- Resetting the TCC microphone approval state for the app:
  ```
  tccutil reset Microphone com.umputun.agterm
  ```

### Root cause
`Info.plist` inside the app bundle has **no `NSMicrophoneUsageDescription` key**:

```
$ /usr/libexec/PlistBuddy -c "Print" /Applications/agterm.app/Contents/Info.plist
Dict {
    CFBundleName = Agterm
    ...
    CFBundleVersion = 0.8.0
    CFBundleIdentifier = com.umputun.agterm
    ...
}
```
(no `NSMicrophoneUsageDescription` entry at all)

On modern macOS, an app is only allowed to request microphone access — and thus get a row created in the TCC database / show up in the Privacy & Security microphone list — if its `Info.plist` declares a usage-description string for that permission. Without `NSMicrophoneUsageDescription`, any attempt to access the microphone is refused at the OS level before a prompt can even be shown, and no TCC entry is ever created. This matches the observed symptom exactly: no dialog, no entry in the Microphone list, regardless of `tccutil reset`.

### Suggested fix
Add `NSMicrophoneUsageDescription` to `Info.plist` (via Xcode's Info tab / build settings, e.g. `INFOPLIST_KEY_NSMicrophoneUsageDescription`), for example:

```xml
<key>NSMicrophoneUsageDescription</key>
<string>Agterm needs microphone access to support voice dictation in terminal sessions (e.g. Claude Code voice input).</string>
```

### Steps to reproduce
1. Install agterm 0.8.0.
2. Run Claude Code inside an agterm session, try `/voice` (hold-to-record or `/voice tap`).
3. No microphone permission dialog appears.
4. Check System Settings → Privacy & Security → Microphone — agterm is not listed.
[...]

PR #398, opened 2026-08-08
fix: add missing TCC resource-access entitlements and usage strings

### LINKED ISSUE #396: Programs run inside agterm can never be granted Automation (and most other TCC permissions) — missing hardened-runtime resource-access entitlements
## Summary

agterm is signed with hardened runtime (`flags=0x10000(runtime)`) but carries only **1 of the 7** resource-access entitlements Apple requires under that runtime. Because agterm is the *responsible process* for everything it spawns, a program run inside a session cannot obtain those permissions — and for Apple Events the request is refused **without any prompt at all**, leaving no path to consent, not even manually in System Settings.

I hit this with Apple Events and verified that one end to end (details and repro below). The other six are the same mechanism and the same one-line-each fix, so I've included the full audit rather than filing them one at a time.

For comparison, **Ghostty, iTerm2 and kitty each ship all 7** — and agterm already does exactly this for the microphone, so the design intent is established.

## The verified case: Apple Events

`tccd` names the cause directly:

```
Prompting policy for hardened runtime; service: kTCCServiceAppleEvents
requires entitlement com.apple.security.automation.apple-events
but it is missing for responsible={TCCDProcess: identifier=com.umputun.agterm, ...},
accessing={TCCDProcess: identifier=com.apple.swift-frontend, ...},
requesting={TCCDProcess: identifier=com.apple.appleeventsd, ...}
-[TCCDPlatformMacOS promptingPolicyForIdentity:accessingService:withAttributionChain:]:
promptPolicy = 4; isApplePlatformBinary = 0
```

**Expected** (the behavior in Terminal.app, iTerm2, kitty and Ghostty): the first Apple Event to a given target shows the standard *"agterm wants access to control &lt;App&gt;"* prompt, and the answer is recorded under System Settings → Privacy & Security → Automation → agterm.

**Actual:** the call returns `-1743 errAEEventNotPermitted` immediately, no prompt appears, and **no TCC record is created**. Two consequences:

- agterm never appears under Privacy & Security → Automation, so the user cannot grant access by hand either. There is no path to consent at all.
- Since nothing is recorded, a later silent check returns `-1744 procNotPermitted` ("never asked"), so the state oscillates between "never asked" and "denied" and never converges. Tools that probe permission state see a target that is perpetually un-granted with no denial to explain it.

Automation grants agterm *already has* keep working — an existing allow record short-circuits the prompt path. Only the creation of **new** grants is impossible, which is what makes this easy to miss: whatever was granted before still works, everything new silently does not.

### Repro

Save as `aeprobe.swift` and run from an agterm session with Google Chrome (or any scriptable app) open:

```swift
import Foundation
import CoreServices

let bundleID = "com.google.Chrome"
var desc = AEDesc()
let data = Array(bundleID.utf8)
AECreateDesc(typeApplicationBundleID, data, data.count, &desc)
// askUserIfNeeded: true  -> should raise the Automation prompt
let st = AEDeterminePermissionToAutomateTarget(&desc, typeWildCard, typeWildCard, true)
AEDisposeDesc(&desc)
print(st)   // 0 = granted, -1743 = denied, -1744 = never asked, -600 = not running
```

```
$ xcrun swift aeprobe.swift
-1743
```

No prompt is shown. The identical script run from Ghostty, iTerm2 or kitty shows the Automation prompt as expected.

## The full audit

Apple's hardened-runtime **Resource Access** entitlements, compared across installed terminals:

| Entitlement | agterm 0.20.2 | Ghostty | iTerm2 | kitty |
| --- | :---: | :---: | :---: | :---: |
| `com.apple.security.device.audio-input` | **yes** | yes | yes | yes |
| `com.apple.security.automation.apple-events` | **missing** | yes | yes | yes |
| `com.apple.security.device.camera` | **missing** | yes | yes | yes |
| `com.apple.security.personal-information.addressbook` | **missing** | yes | yes | yes |
| `com.apple.security.personal-information.calendars` | **missing** | yes | yes | yes |
| `com.apple.security.personal-information.location` | **missing** | yes | yes | yes |
| `com.apple.security.personal-information.photos-library` | **missing** | yes | yes | yes |

`Info.plist` usage strings present in **all three** peers and absent from agterm:

```
NSAppleEventsUsageDescription        NSPhotoLibraryUsageDescription
NSBluetoothAlwaysUsageDescription    NSRemindersUsageDescription
NSCalendarsUsageDescription          NSSpeechRecognitionUsageDescription
NSCameraUsageDescription             NSSystemAdministrationUsageDescription
NSContactsUsageDescription           NSLocalNetworkUsageDescription
NSLocationUsageDescription
```

agterm ships one: `NSMicrophoneUsageDescription`. (Bluetooth, Local Network, Speech Recognition, System Administration and Reminders are TCC services that need only the usage string — they have no hardened-runtime entitlement.)

**Scope of what I actually verified:** the Apple Events case is confirmed end to end, with the `tccd` log above and a reproducible script. The other six entitlements I verified only as *absent from the signature* — I did not exercise each one. They go through the same `promptingPolicyForIdentity:` gate, so I expect the same refusal, but I'm flagging that as inference rather than a measurement.

## Why Terminal.app is not a counterexample

Terminal.app ships **zero** usage strings and none of these entitlements, which might suggest they're unnecessary. It gets away with it via a private Apple entitlement:

```
com.apple.private.tcc.allow-prompting
```

Third-party apps can't use that, which is exactly why iTerm2, kitty and Ghostty all carry the public entitlements and usage strings instead.

## Suggested fix

agterm embeds Ghostty, and **upstream Ghostty already ships precisely the right set** — so the cheapest correct fix is to adopt Ghostty's entitlements file and usage strings wholesale. That's 6 entitlement keys and ~11 `Info.plist` strings, and it can be verified in seconds with:

```
codesign -d --entitlements - /Applications/Ghostty.app
```

agterm's existing microphone handling is already the template — `com.apple.security.device.audio-input` plus:

> `NSMicrophoneUsageDescription` = "Command-line tools you run inside agterm (for example, voice dictation in Claude Code) request microphone access through agterm."

A parallel string for Apple Events would be something like:

> "Command-line tools you run inside agterm request permission to control other apps through agterm."

Note the three entitlements agterm has that the peers don't (`allow-jit`, `allow-unsigned-executable-memory`, `disable-library-validation`) are unrelated to TCC and nothing here suggests changing them.

## Why this matters

agterm's whole purpose is running other programs, and it already accepts the role of proxying TCC consent on their behalf. Automation in particular is how a lot of ordinary macOS tooling works — AppleScript in scripts and Makefiles, `osascript` one-liners, editor and browser integrations, and locally-built apps under development that talk to other apps over Apple Events. Under agterm none of these can obtain consent, and the failure mode is the unhelpful kind: no prompt, no error dialog, and no entry in System Settings to fix by hand — just a permission that silently never arrives.

## Environment

- agterm 0.20.2
- macOS 26.5.2 (25F84), Apple Silicon
- Compared against Ghostty, iTerm2 and kitty as installed on the same machine

---

Happy to open a PR for this if you'd like — it should be a small change to the entitlements plist and `Info.plist`. Just say the word and let me know whether you'd prefer the full Ghostty-parity set or only `com.apple.security.automation.apple-events` + `NSAppleEventsUsageDescription` (the one case verified here).

skkap:
Fixes #396. Same mechanism as #142, one layer down: that was a missing usage string, these are missing hardened-runtime entitlements.

agterm is signed with hardened runtime, under which Apple gates seven **resource-access** entitlements. agterm carried one (`device.audio-input`). For the other six, `tccd` refuses to *prompt* and records nothing — so a tool run inside a session cannot obtain the permission, and the user cannot grant it by hand either, because with no TCC record agterm never appears in the relevant Privacy & Security pane. Grants agterm already holds keep working, which is why this went unnoticed: only *new* grants are impossible.

## Changes

**`agterm/agterm.entitlements`** — the six missing resource-access entitlements:

```
automation.apple-events   device.camera   personal-information.addressbook
personal-information.calendars   personal-information.location   personal-information.photos-library
```

**`agterm/Info.plist`** — usage strings for the services with a real command-line path, worded to match the existing microphone string: AppleEvents, Camera, Contacts, Calendars (+ full-access, + write-only), Reminders (+ full-access), PhotoLibrary, Location, Bluetooth, LocalNetwork, SpeechRecognition.

Every key was checked against Apple's documented platform availability rather than copied from other terminals — that caught three keys iTerm2/Ghostty/kitty ship which are **iOS/iPadOS/visionOS only** and do nothing on macOS (`NSLocationWhenInUseUsageDescription`, `NSLocationAlwaysAndWhenInUseUsageDescription`, `NSPhotoLibraryAddUsageDescription`). The macOS key is `NSLocationUsageDescription` — 10.14+, not deprecated there.

The deprecated `NSCalendarsUsageDescription` / `NSRemindersUsageDescription` are kept deliberately. Both are deprecated at macOS 14, but agterm is the responsible app for third-party tools of unknown SDK vintage, and one linked against a pre-14 SDK still resolves against the legacy key.

**`project.yml`** — the reseal applied `--deep --entitlements`, which stamped the app's entitlement set onto the bundled `agtermctl`, a standalone CLI on the user's `PATH`. Measured before the fix, the helper carried all six TCC entitlements. Now `--deep` signs without entitlements and a second non-deep pass applies them to the app alone. Shipped artifacts were never affected — `release.sh` re-signs the helper itself — only `make build` / `make deploy` output.

## Verification

Each service was exercised from a CLI inside agterm 0.20.2 / macOS 26.5.2, against the **released build** — the failures this PR fixes, reproduced per service:

| Service | Result | `tccd` |
|---|---|---|
| Apple Events | `-1743`, no prompt | `kTCCServiceAppleEvents requires entitlement …automation.apple-events but it is missing` |
| Camera | denied | `kTCCServiceCamera requires …device.camera…` |
| Contacts | denied | `kTCCServiceAddressBook requires …addressbook…` |
| Calendars | denied | `kTCCServiceCalendar requires …calendars…` |
| Photos | denied | `kTCCServicePhotos requires …photos-library…` |
| Location | `notDetermined`, no prompt | none — handled by `locationd`, not `tccd` |

All carry `promptPolicy = 4` and `responsible={com.umputun.agterm}`.

**That the entitlements are the cure** was then shown with two minimal apps identical except for the entitlements passed to `codesign` — same binary, same usage strings, both `flags=0x10002(adhoc,runtime)`, distinct bundle ids:

| | no entitlements | with entitlements |
|---|---|---|
| Apple Events | `-1743`, no prompt | prompt → granted |
| Camera | denied, no prompt | prompt → granted |
| Location | `notDetermined` | prompt → `authorizedAlways` |

`tccd` un-entitled: `Policy disallows prompt … denied`. Entitled: `AUTHREQ_PROMPTING`, all three dialogs confirmed on screen. This also settles location, which produces no `tccd` line of its own.

Build: `make build` and the signing phase pass from an empty `DerivedData`, with `agtermctl` carrying no entitlements, the app all six, and `codesign --verify --deep --strict` clean.

## ✅ Resolved — `NSSystemAdministrationUsageDescription`

> **Decision reached and addressed.** Approved in review; the key is in the diff as of b5f641c. The
> follow-up asked alongside it — whether a missing purpose string terminates the process rather than
> denying — was checked: it denies. Details in the review thread.
>
> Original question kept below for the record.

**I think this should go in, and I'd like to add it.** It is not in the diff because you asked me to leave it out, and I'm not going to include it over that without checking first — but I don't think the reason it was excluded holds up:

1. **The exclusion tested a different subsystem.** It was ruled out because `CLIInstaller.elevatedSymlink` already elevates without it. That path is `do shell script … with administrator privileges` — Authorization Services, not TCC. This key gates a TCC service, so the installer working says nothing either way about it.
2. **Apple documents it as macOS-only and required.** macOS 10.14+, not deprecated: *"A message in macOS that tells people why the app is requesting to manipulate the system configuration"*, *"This key is required if your app uses APIs that manipulate the system configuration"*, naming `ODRecordSetValue` — the OpenDirectory API behind `dscl`. A tool running `dscl . -create` in a session lands exactly here, with agterm responsible.
3. **It is the same class as every other key here, and all three peers ship it.** Leaving this one out makes agterm the lone outlier for no mechanical reason, and the failure mode is the silent one this whole PR exists to remove.

~~One word and I'll push it.~~

**Still deliberately out of scope:** the Files & Folders family (`NSDesktopFolderUsageDescription` and friends). Those prompts already work — the keys would only replace Apple's generic wording with agterm's — so that is a copy improvement, not a fix, and belongs in its own change.

## ✅ Resolved — local `make test-app` failure

> **Decision reached and addressed.** Already fixed on master by 16d8cbc; this branch is rebased onto it
> and `make test-app` is green. The diagnosis below was wrong — see #400, which is closed as redundant.
>
> Original note kept below for the record.

~~`make test-app` fails locally on `LiveMenuKeyEquivalentsTests.testUppercaseKeyEquivalentReportsImpliedShift()`, on unmodified `master` too. Unrelated to this PR and fixed separately in #400 — the fixture uses the title "Paste and Match Style", which AppKit recognises as a standard item and rewrites to ⇧⌘V on `addItem`, before the code under test runs.~~

### DIFF
--- a/agterm/Info.plist
+++ b/agterm/Info.plist
@@ -62,6 +62,8 @@
 	<string>Command-line tools you run inside agterm request access to your reminders through agterm.</string>
 	<key>NSSpeechRecognitionUsageDescription</key>
 	<string>Command-line tools you run inside agterm request speech recognition through agterm.</string>
+	<key>NSSystemAdministrationUsageDescription</key>
+	<string>Command-line tools you run inside agterm (for example, dscl editing a directory record) request permission to change system configuration through agterm.</string>
 	<key>CFBundleDocumentTypes</key>
 	<array>
 		<dict>
--- a/agterm/Info.plist
+++ b/agterm/Info.plist
@@ -38,6 +38,8 @@
 	<string>Command-line tools you run inside agterm (for example, an osascript one-liner or a build script) request permission to control other apps through agterm.</string>
 	<key>NSBluetoothAlwaysUsageDescription</key>
 	<string>Command-line tools you run inside agterm request Bluetooth access through agterm.</string>
+	<key>NSCalendarsFullAccessUsageDescription</key>
+	<string>Command-line tools you run inside agterm request access to your calendars through agterm.</string>
 	<key>NSCalendarsUsageDescription</key>
 	<string>Command-line tools you run inside agterm request access to your calendars through agterm.</string>
 	<key>NSCameraUsageDescription</key>
@@ -46,14 +48,18 @@
 	<string>Command-line tools you run inside agterm request access to your contacts through agterm.</string>
 	<key>NSLocalNetworkUsageDescription</key>
 	<string>Command-line tools you run inside agterm (for example, connecting to a .local host) request local network access through agterm.</string>
-	<key>NSLocationUsageDescription</key>
+	<key>NSLocationAlwaysAndWhenInUseUsageDescription</key>
 	<string>Command-line tools you run inside agterm request your location through agterm.</string>
 	<key>NSLocationWhenInUseUsageDescription</key>
 	<string>Command-line tools you run inside agterm request your location through agterm.</string>
 	<key>NSMicrophoneUsageDescription</key>
 	<string>Command-line tools you run inside agterm (for example, voice dictation in Claude Code) request microphone access through agterm.</string>
+	<key>NSPhotoLibraryAddUsageDescription</key>
+	<string>Command-line tools you run inside agterm (for example, a script importing a screenshot) request permission to add to your photo library through agterm.</string>
 	<key>NSPhotoLibraryUsageDescription</key>
 	<string>Command-line tools you run inside agterm (for example, osxphotos) request access to your photo library through agterm.</string>
+	<key>NSRemindersFullAccessUsageDescription</key>
+	<string>Command-line tools you run inside agterm request access to your reminders through agterm.</string>
 	<key>NSRemindersUsageDescription</key>
 	<string>Command-line tools you run inside agterm request access to your reminders through agterm.</string>
 	<key>NSSpeechRecognitionUsageDescription</key>
--- a/agterm/Info.plist
+++ b/agterm/Info.plist
@@ -34,8 +34,30 @@
 	<string>public.app-category.developer-tools</string>
 	<key>NSHumanReadableCopyright</key>
 	<string>© 2026 Umputun</string>
+	<key>NSAppleEventsUsageDescription</key>
+	<string>Command-line tools you run inside agterm (for example, an osascript one-liner or a build script) request permission to control other apps through agterm.</string>
+	<key>NSBluetoothAlwaysUsageDescription</key>
+	<string>Command-line tools you run inside agterm request Bluetooth access through agterm.</string>
+	<key>NSCalendarsUsageDescription</key>
+	<string>Command-line tools you run inside agterm request access to your calendars through agterm.</string>
+	<key>NSCameraUsageDescription</key>
+	<string>Command-line tools you run inside agterm (for example, ffmpeg capturing video) request camera access through agterm.</string>
+	<key>NSContactsUsageDescription</key>
+	<string>Command-line tools you run inside agterm request access to your contacts through agterm.</string>
+	<key>NSLocalNetworkUsageDescription</key>
+	<string>Command-line tools you run inside agterm (for example, connecting to a .local host) request local network access through agterm.</string>
+	<key>NSLocationUsageDescription</key>
+	<string>Command-line tools you run inside agterm request your location through agterm.</string>
+	<key>NSLocationWhenInUseUsageDescription</key>
+	<string>Command-line tools you run inside agterm request your location through agterm.</string>
 	<key>NSMicrophoneUsageDescription</key>
 	<string>Command-line tools you run inside agterm (for example, voice dictation in Claude Code) request microphone access through agterm.</string>
+	<key>NSPhotoLibraryUsageDescription</key>
+	<string>Command-line tools you run inside agterm (for example, osxphotos) request access to your photo library through agterm.</string>
+	<key>NSRemindersUsageDescription</key>
+	<string>Command-line tools you run inside agterm request access to your reminders through agterm.</string>
+	<key>NSSpeechRecognitionUsageDescription</key>
+	<string>Command-line tools you run inside agterm request speech recognition through agterm.</string>
 	<key>CFBundleDocumentTypes</key>
 	<array>
 		<dict>
--- a/agterm/Info.plist
+++ b/agterm/Info.plist
@@ -42,20 +42,18 @@
 	<string>Command-line tools you run inside agterm request access to your calendars through agterm.</string>
 	<key>NSCalendarsUsageDescription</key>
 	<string>Command-line tools you run inside agterm request access to your calendars through agterm.</string>
+	<key>NSCalendarsWriteOnlyAccessUsageDescription</key>
+	<string>Command-line tools you run inside agterm request permission to add events to your calendars through agterm.</string>
 	<key>NSCameraUsageDescription</key>
 	<string>Command-line tools you run inside agterm (for example, ffmpeg capturing video) request camera access through agterm.</string>
 	<key>NSContactsUsageDescription</key>
 	<string>Command-line tools you run inside agterm request access to your contacts through agterm.</string>
 	<key>NSLocalNetworkUsageDescription</key>
 	<string>Command-line tools you run inside agterm (for example, connecting to a .local host) request local network access through agterm.</string>
-	<key>NSLocationAlwaysAndWhenInUseUsageDescription</key>
-	<string>Command-line tools you run inside agterm request your location through agterm.</string>
-	<key>NSLocationWhenInUseUsageDescription</key>
+	<key>NSLocationUsageDescription</key>
 	<string>Command-line tools you run inside agterm request your location through agterm.</string>
 	<key>NSMicrophoneUsageDescription</key>
 	<string>Command-line tools you run inside agterm (for example, voice dictation in Claude Code) request microphone access through agterm.</string>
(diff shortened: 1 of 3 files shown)


umputun:
add it.

your read on the exclusion is right: the CLI installer path is `do shell script ... with administrator privileges`, which is Authorization Services, not TCC, so it says nothing about `NSSystemAdministrationUsageDescription` either way. And `dscl . -create` in a session is exactly the "real command-line path" test the other strings pass, so leaving this one out is the outlier.

one thing worth checking while you have the scratch apps out: this key has no hardened-runtime entitlement behind it, unlike the six in the diff. If a TCC service that requires a purpose string and has none terminates the touching process instead of denying it, the failure mode here is the tool dying rather than the silent deny the rest of this PR fixes. I have not verified that, so treat it as a guess until you check.

push it and I'll merge.

skkap:
Rebased on master — you were right about #400, I should have pulled first.

## ✅ Decision reached and addressed — `NSSystemAdministrationUsageDescription`

Added in b5f641c, worded to match the rest:

> Command-line tools you run inside agterm (for example, dscl editing a directory record) request permission to change system configuration through agterm.

## ✅ Decision reached and addressed — missing purpose string

Checked it with the scratch apps: **it denies, it does not terminate.** Same binary in two bundles differing only in the usage string — the stringed one prompts and is granted, the string-less one gets a plain "Access Denied" and runs to completion with no crash. So the failure mode here is the same silent deny as the rest of this PR, not a dying tool. Full measurements on request.

Two limits, since you asked me to treat it as a guess until checked: I could not drive the SysAdmin service itself to a TCC decision — OpenDirectory rejects the write on credentials before TCC is consulted — and the behaviour may vary by framework. So that result is measured for Contacts and expected, not verified, for the rest.

## Gates

Clean on the rebase: `swift test` 2411 tests, `make test-app` 176 tests, `swiftlint --strict` 0 violations.

## Description

Both open items are marked ✅ Resolved with what was decided and where it landed, originals kept struck through.

umputun (approved):
lgtm

PR #400, opened 2026-08-08
fix(tests): stop AppKit rewriting the uppercase-key-equivalent fixture

skkap:
`make test-app` fails on `LiveMenuKeyEquivalentsTests.testUppercaseKeyEquivalentReportsImpliedShift()`:

```
XCTAssertEqual failed: ("cmd+shift+v") is not equal to ("cmd+opt+shift+v")
```

Product code is fine — the fixture is mutated before the code under test runs.

## Cause

The test builds its item with the title `"Paste and Match Style"`. AppKit recognises that as a **standard** menu item and, on `NSMenu.addItem`, rewrites it to the system-standard equivalent — so `"V"` + `[.command, .option]` becomes `"v"` + `[.command, .shift]`. `collectKeyEquivalents` then correctly reports `cmd+shift+v` for the item it was actually handed, and the assertion compares that against the `cmd+opt+shift+v` the test believed it built.

The rewrite keys off the **title alone**. Measured on macOS 26.5.2:

| fixture | after `addItem` |
|---|---|
| title `"Paste and Match Style"`, key `"V"`, mask `[cmd, opt]` | key `"v"`, mask `[cmd, shift]` |
| title `"Totally Custom Title"`, key `"V"`, mask `[cmd, opt]` | key `"V"`, mask `[cmd, opt]` — untouched |
| title `"Paste and Match Style"`, key `"Q"`, mask `[cmd, opt]` | key `"v"`, mask `[cmd, shift]` |

The third row is the telling one: a completely different key equivalent is still replaced with ⇧⌘V, because the title is what AppKit matches on. Before `addItem` the item holds exactly what the test set, so the rewrite happens at insertion.

It is not key-equivalent localization — `allowsAutomaticKeyEquivalentLocalization = false` makes no difference — and it is not the chord logic, which never sees the original values.

## Fix

Rename the fixture item to a non-stock title, and record why in a comment so it does not get "corrected" back to a realistic-looking stock name. One line of test data; `chordSyntax` is untouched.

After: `make test-app` reports **TEST SUCCEEDED**, and `swiftlint --strict` stays at 0 violations across 326 files.

## Why it presumably passes in CI

I did not confirm this, so treat it as a guess: the rewrite depends on AppKit's standard-item table, so a different macOS version — or a different system language, since the match is against localized standard titles — would not fire it. Reproduced here on macOS 26.5.2 / Xcode 26, where it fails identically on unmodified `master` (`4155ed9`).

Split out of #398, which ran into it.

### DIFF
--- a/agtermTests/LiveMenuKeyEquivalentsTests.swift
+++ b/agtermTests/LiveMenuKeyEquivalentsTests.swift
@@ -78,8 +78,13 @@ final class LiveMenuKeyEquivalentsTests: XCTestCase {
 
     // AppKit matches an uppercase key equivalent against the SHIFTED chord even with shift absent from
     // the mask, so lowercasing without adding shift would name a chord the item can never fire.
+    //
+    // The title must NOT be one AppKit recognises as a standard menu item ("Paste and Match Style" was,
+    // and is why this used to fail): `NSMenu.addItem` rewrites such an item to the system-standard
+    // equivalent, turning ⌥⌘V into ⇧⌘V before the code under test ever sees it — a fixture that
+    // silently stops testing what it says. The rewrite keys off the title alone, not the chord.
     func testUppercaseKeyEquivalentReportsImpliedShift() throws {
-        let file = menu("Edit", [item("Paste and Match Style", key: "V", mods: [.command, .option])])
+        let file = menu("Edit", [item("Vendor Paste Special", key: "V", mods: [.command, .option])])
 
         let found = try XCTUnwrap(ControlServer.collectKeyEquivalents(in: file, menu: "Edit").first)
 


umputun:
this is already fixed on master by 16d8cbc, merged a day before your commit. Your branch forks from 4155ed9 and is 15 commits behind, so it doesn't carry the fix.

the cause isn't a standard-item table. AppKit substitutes `NSUserKeyEquivalents`, the App Shortcuts bound by hand in System Settings > Keyboard, and it matches on the menu-item title the moment the item joins a menu, detached menus included. That's why it fails on your machine and nowhere else, and it's neither macOS-version nor language specific. `defaults read -g NSUserKeyEquivalents` should show "Paste and Match Style" bound to ⇧⌘V on yours. The row in your table where key `"Q"` still comes back as ⇧⌘V is the giveaway: the system-standard equivalent for that command is ⌥⇧⌘V, which is what the test asserts and would have passed.

renaming the fixture also doesn't generalize. The substitution keys on any title, not only stock ones, so `"Vendor Paste Special"` is rewritten just the same on a machine where that string is bound. It leaves `"Zoom"` and `"Close"` untouched in the same file, and `CloseSessionChordTests` needs the real `"Close"` to test chord ownership against agterm's own item, so that one can't be renamed at all.

16d8cbc suppresses the substitution for the duration of each test class instead, `NSMenuItem.usesUserKeyEquivalents = false` in `setUp` with a restore in `tearDown`, in both classes. Titles stay as they are.

closing this one. Rebase #398 on master and the failure goes with it.

PR #428, opened 2026-08-13
docs(cookbook): add container-agent-status recipe

nquo:
## Summary
  - Adds `container-agent-status`, a recipe for reporting a hook-driven agent's
  status onto its sidebar row when the agent runs inside a container instead of on
  the host — the installer's own hooks can't reach the local control socket from in
  there.
  - A small script inside the container (`container-status-notify.sh`) turns each
  Claude Code hook event into a flat JSON line and sends it to the host over TCP; a
  listener on the host (`agterm-remote-receiver.sh`) turns each line back into the
  same `agtermctl session status` call the installer's own hooks would make
  directly.
  - Zero extra dependencies: the whole path runs on plain `nc` on both ends — no
  `jq`, no `socat`, nothing to `brew install`. macOS ships `nc`, and so does most
  Linux base imagery a Claude Code container would use, so there is usually nothing
  to install at all.
  - Tested end to end on macOS with Apple's `container` CLI (`container machine
  run`) running a Linux Claude Code container — see the README's note on why that
  runtime's networking (host ports reachable from the container by default, no
  publish/rebuild needed for the reverse direction) makes the recipe simpler than
  on Docker/Podman.
  - Adds the index row to `cookbook/README.md` under **Agent status and
  workflows**.

## Test plan
  - [x] `bash -n` on both scripts
  - [x] Manual end-to-end run: container script → `nc` → receiver on host →
  `agtermctl session status`, confirmed in the receiver's log and the sidebar row
  - [x] README carries all six required headings (What it does / Requirements /
  Setup / Usage / How it works / Limits)
  - [x] Index row in `cookbook/README.md` matches the directory name
  - [x] No absolute home-directory paths, hostnames, IPs, or other personal data in
  the scripts or README (checked with a full grep pass)

### DIFF
--- a/cookbook/README.md
+++ b/cookbook/README.md
@@ -39,7 +39,7 @@ its *needs* column, so searching this page for `claude` or `kiro` finds those di
 | [claude-clear](claude-clear/) | one chord sends /clear to the pane's Claude Code run, and nothing when it is not running | 0.13.0, python3, Claude Code |
 | [claude-conversation-picker](claude-conversation-picker/) | pick a past Claude Code conversation by what it was about and resume it in the pane | 0.21.0, python3, Claude Code |
 | [claude-recap](claude-recap/) | one key lists what the Claude Code run in a session was working on | 0.10.0, zsh, jq, Claude Code |
-| [container-agent-status](container-agent-status/) | a containerized agent reports status onto its sidebar row via a TCP notification to the host | 0.7.1, nc, Claude Code |
+| [container-agent-status](container-agent-status/) | a containerized agent reports status onto its sidebar row via a TCP notification to the host | 0.7.1, nc, timeout, Claude Code |
 | [kimi-agent-status](kimi-agent-status/) | Kimi Code sessions report agent status onto their sidebar row | 0.3.1, Kimi Code |
 | [kiro-agent-status](kiro-agent-status/) | Kiro CLI sessions report active, blocked, and completed onto their sidebar row | 0.7.1, Kiro CLI |
 | [status-announcer](status-announcer/) | demo: speak agent status changes from a dedicated session | 0.16.0, jq |
--- a/cookbook/README.md
+++ b/cookbook/README.md
@@ -39,6 +39,7 @@ its *needs* column, so searching this page for `claude` or `kiro` finds those di
 | [claude-clear](claude-clear/) | one chord sends /clear to the pane's Claude Code run, and nothing when it is not running | 0.13.0, python3, Claude Code |
 | [claude-conversation-picker](claude-conversation-picker/) | pick a past Claude Code conversation by what it was about and resume it in the pane | 0.21.0, python3, Claude Code |
 | [claude-recap](claude-recap/) | one key lists what the Claude Code run in a session was working on | 0.10.0, zsh, jq, Claude Code |
+| [container-agent-status](container-agent-status/) | a containerized agent reports status onto its sidebar row via a TCP notification to the host | 0.7.1, nc, Claude Code |
 | [kimi-agent-status](kimi-agent-status/) | Kimi Code sessions report agent status onto their sidebar row | 0.3.1, Kimi Code |
 | [kiro-agent-status](kiro-agent-status/) | Kiro CLI sessions report active, blocked, and completed onto their sidebar row | 0.7.1, Kiro CLI |
 | [status-announcer](status-announcer/) | demo: speak agent status changes from a dedicated session | 0.16.0, jq |
--- a/cookbook/container-agent-status/README.md
+++ b/cookbook/container-agent-status/README.md
@@ -1,0 +1,105 @@
+# Container agent status
+
+Report a Claude Code (or any hook-driven agent) session's status onto its sidebar row when that agent runs inside a container instead of on the host.
+
+## What it does
+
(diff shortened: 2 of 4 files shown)


umputun (commented):
three blockers, rest are notes.

**`agterm-remote-receiver.sh:75-107` drops `--blink` and `--auto-reset`.** `process_event` greps only `cmd`/`state`/`session_id`/`pane` and builds a fixed argv, so the `"args"` array `container-status-notify.sh:27-34` sends is never read. Both flags are `@Flag = false`, and `clearAutoResetIndicator` gates on `autoReset`, so the row never pulses and `completed` never clears on visit. Three README lines say otherwise: "The row pulses active while the agent works", the Usage line promising a completed flash "clearing when you visit the session", and "precisely the call the installer's own hooks make". Fix: allowlist the two flags in `process_event` and append them to `args` before the call.

**`agterm-remote-receiver.sh:109-115` closes the port for a second after every event, not just after a failure.** `sleep 1` is unconditional, so a `Stop` or `blocked` arriving inside that gap gets ECONNREFUSED and `container-status-notify.sh:37` swallows it with `|| true`, leaving the row asserting a state that is no longer true. Fix: `nc -k -l "$PORT"` piped into one `while read` loop, dropping the outer `while true` and the sleep; macOS nc has `-k`. Limits currently says "a very rare missed event, never a wrong one", which needs rewording either way.

**`agterm-remote-receiver.sh:54-63` logs raw socket bytes without stripping control characters.** `log()` interpolates `$*` straight into `echo`, and :110 passes the whole unparsed line through, while agterm sets no `clipboard-write` key so ghostty's `allow` default stands and `GhosttyCallbacks.swift:250` writes the pasteboard unprompted. A line carrying an OSC 52 sequence replaces the clipboard of whoever has the receiver's output on screen, and `--log-file` stores the bytes so `tail`/`cat` replays them later. Fix: strip once in `log()`, `line="[$(date '+%Y-%m-%d %H:%M:%S')] [$level] $(printf '%s' "$*" | tr -d '[:cntrl:]')"`.

notes:

- `README.md:12` and `cookbook/README.md:42` omit `timeout`, which `container-status-notify.sh:37` calls on the only delivery path. CONTRIBUTING wants every external tool named in Requirements, so `0.7.1, nc, timeout, Claude Code`.
- `container-status-notify.sh:25` forwards `AGTERM_PANE` but not `AGTERM_PANE_ID`, so the promoted-survivor case (#199) puts the status on the wrong pane. Forwarding it conditionally is self-gating, a 0.7.1 host never exports it, so this does not move the version floor.
- `README.md:100` states the trust model correctly but never says `nc -l` binds every interface, so "do not expose the port beyond that" reads like advice the reader can act on when he has no lever. One sentence in Limits covers it; a `--bind ADDR` passed as `nc -l "$BIND" "$PORT"` if you want the knob.
- `container-status-notify.sh:22-23` splits `AGTERM_REMOTE_HOST` with `%:*`/`##*:`, so a colon-less value silently becomes host == port and fails with nothing in either log. A two-line `case` guard turns that into an early exit.

umputun (approved):
all three blockers are fixed, and the four notes with them. Merging.

a few things for later, none of them holding this up.

**the recipe fails silently on a host with Homebrew's `netcat`.** That formula installs `/opt/homebrew/bin/nc`, which shadows `/usr/bin/nc` and has no `-k`, so it answers `nc: invalid option -- k` and exits 1. `agterm-remote-receiver.sh:130` sends that to `2>/dev/null`, so the reader gets `listening on 127.0.0.1:9993` and then `listener exited`, exit code 0, and never a status. Setup starts it with `&`, so it looks like a background job that came up fine. My own box is set up this way, which is how I hit it. Requirements at `README.md:12` says "`nc` (netcat)", which does not distinguish the two: Apple's has `-k`, the Homebrew formula of the same name does not. Dropping the `2>/dev/null` would have named it, and the same silence covers a port already in use or a `--bind` address that is not on the host.

**the receiver discards every failure signal it has.** Same shape three times: `nc`'s stderr above, `agtermctl`'s exit status at `:117` (`>/dev/null 2>&1 || true`), and no presence check for `$AGTERMCTL` even though `nc` gets one at `:120`. So `agtermctl` not installed, a custom `AGTERM_STATE_DIR`, a session id that stopped resolving, or agterm not running all log the same `[INFO] event: ...` line, which reads as a delivery confirmation. `--verbose` prints `forwarding: ...` before the call rather than after it, so it confirms a send that failed. Since Setup tells the reader to run with `--log-file`, a healthy-looking log and no row change is the hardest case to unpick.

**Setup's copy-paste line still binds every interface.** Limits explains it now and `--bind` exists, but `README.md:29` is the line people paste. Worth putting `--bind 127.0.0.1` in it and letting Limits explain when to widen. Exposure is small either way: the allowlist takes only `--blink`/`--auto-reset`, so nothing composes arbitrary argv. `state` does reach agtermctl as a token, so a value like `--sound` gets there, but ArgumentParser rejects it, and what is left is a LAN peer setting indicator states on session ids it guesses.

**`README.md:95` calls the wire format the same shape the control socket takes.** It is not: you send `cmd:"session-status"` with top-level `state` and `session_id`, the native request is `cmd:"session.status"` with `target` and a nested `args.status`. Someone who reads that and sends a real native request gets it dropped with no complaint.

two smaller ones: `nc -k -l` handles one connection at a time, so a peer that connects without sending a newline parks the listener until the senders' `timeout` gives up, and `:90` logs the extracted `session_id` before `:92` validates it, so a bad line can grow the log. Both are inside the trust model the README already states.

Issue #445, opened 2026-08-17
Quit confirmation blocks a system-initiated quit, so an unattended restart loses every captured command


ssgreg:
Leave sessions running with *Restore running commands on restart* enabled, then let the machine restart with nobody at the keyboard. A policy-scheduled macOS update at night is the usual way this happens. After the reboot every tab is back with its directory and its name, and every pane sits at a shell prompt. Nothing was re-run.

Here is why. `applicationShouldTerminate` puts up a blocking `NSAlert` whenever a window is open, and it makes no exception for a quit the system asked for:

```swift
guard counts.windows > 0 else { return .terminateNow }
let alert = NSAlert()
alert.messageText = "Quit Agterm?"
...
return alert.runModal() == .alertFirstButtonReturn ? .terminateNow : .terminateCancel
```

With nobody there to click Quit, the app is killed once the shutdown timeout runs out, so `applicationWillTerminate` never runs. That is the only place `captureForegroundCommands` runs on a normal exit, so the state is saved with no captures.

The docs already say a force-quit or crash captures nothing, and that is fair. What is not fair is that an ordinary system restart cannot be anything but a force-quit: the app blocks its own clean exit and there is nobody there to unblock it. The setting is named "on restart", and an update-driven restart is the most ordinary restart there is.

What I measured, on 0.22.0:

- the update installed overnight, the machine booted at 03:06
- all 50 sessions came back with their cwd, 21 of them with their custom names
- `windows/<id>.json` carried no `foregroundCommand` on any session
- a deliberate ⌘Q two days earlier, same build, restored 48 of 50 commands, so the feature itself works. The exit path is what fails.

The fix looks small: skip the confirmation when the quit comes from the system, so `willTerminate` still runs and the capture happens exactly as it does for ⌘Q. Either an `NSWorkspace.willPowerOffNotification` observer that sets a flag, matching how `SystemWakeObserver` and `SystemAccessibilityObserver` already bridge workspace notifications, or the quit-reason check on the current Apple event inside `applicationShouldTerminate`. Not holding up a system shutdown with a modal nobody can answer is worth having on its own.

This reads like the same class as #370, where quitting by closing the last window lost every captured command: another exit path that never reached the capture. Happy to send the patch if you want it done this way.

ssgreg:
Worth adding: ghostty's own macOS app already handles this case, so there is a working reference for the mechanism. `macos/Sources/App/AppDelegate.swift`:

```swift
[...]

PR #447, opened 2026-08-17
fix: read the quit reason with kAEQuitReason, not a keyword built from a string

ssgreg:
Thanks for turning #445 around within the hour. The shape of #446 is right, and it does not go far enough: as merged, the check cannot fire.

`AEKeyword` is `FourCharCode`, which is `UInt32`, so `AEKeyword("why?")` resolves to `UInt32.init?(String)`, the decimal parser, and is always nil. The `if let` chain stops on its second condition, the alert is built anyway, and an unattended restart still loses every captured command.

So this does not read as a claim, here is a throwaway test in this repo's own suite, not part of this diff:

```
✘ Test theKeywordTheAppBuildsResolves() recorded an issue at QuitReasonTests.swift:10:9:
  Expectation failed: (AEKeyword("why?") → nil) != nil
```

Nothing in the suite could have caught it, which is worth saying out loud: the read sat inline in the app target, out of reach of the host-free tests, and the hosted suite returns `.terminateNow` under XCUITest before the check. `QuitReasonTests` was green because the classifier is fine. It is the keyword that never resolved.

**The fix.** The read moves into `agtermCore` as `QuitReason.isSystemQuit`, which is what makes the seam testable, and the keyword comes from `kAEQuitReason`, which Foundation exposes: 0x7768793F, the same number your string algorithm produces. `AppDelegate` keeps one line, and the three codes stay exactly as you built them.

The tests build the quit event AppKit delivers and assert through the code under test: shut, rest and rlgo skip the prompt, a scripted quia keeps it, an event with no reason attribute or no event keeps it too. A keyword that fails to resolve fails all three positive cases.

`make test` 2569 tests green, `make lint` clean, `make build` succeeds. The manual check is unchanged from #446: whether macOS delivers the event before `applicationShouldTerminate` on a real restart.

Two notes, neither touched here. The trap is an easy one to hit and not yours alone: ghostty carries the same dead guard in `macos/Sources/App/AppDelegate.swift`, so the implementation we both read never reached its switch either. And `kAELogOut` (logo, 1819240303) is still outside the set, only rlgo is in it: I have not established which of the two macOS sends for an ordinary logout, so I left the set alone.

Follows #446, reported in #445.

### DIFF
--- a/.claude/rules/windows.md
+++ b/.claude/rules/windows.md
@@ -67,8 +67,10 @@ session drag are out of scope.
   `Debouncer`; structural mutations save synchronously and cancel pending saves.
 - Quit uses `applicationShouldTerminate` and a warning alert with host-free `openCounts` and
   `QuitPrompt.message`. Skip it for system shutdown/restart/logout, no open windows, XCUITest, or an
-  unwired library during the first roughly four seconds. The GUI-only prompt is keep-in-sync exempt and
-  manually verified.
+  unwired library during the first roughly four seconds. The system-quit half is host-free in
+  `QuitReason.isSystemQuit`, which reads `kAEQuitReason` off the current Apple event, so the read itself
+  is under test: a keyword built as `AEKeyword("why?")` is `UInt32.init?(String)` and always nil. The
+  GUI-only prompt is keep-in-sync exempt and manually verified.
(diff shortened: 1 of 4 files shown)


umputun (approved):
lgtm

PR #452, opened 2026-08-18
feat: add restore capture, filling the capture slot on demand

### LINKED ISSUE #454: A control byte in a captured argv breaks the restored line: the pane comes back stuck at `quote>`
A captured argv keeps its bytes exactly as the process had them, and `CommandRestore.shellQuotedLine` wraps each argument in single quotes. That protects the shell's parser, and the restored line never reaches the parser as text: it is typed into the pane through `initial_input`, so the line editor sees it first. A C0 byte there is an editing command, not a character.

Take an ordinary one-liner, an ANSI stripper reading stdin:

```
sed $'s/\x1b//'
```

zsh expands `$'\x1b'` into a real ESC byte, so `argv[1]` holds `s/ESC//`. The capture is faithful, the window snapshot keeps the byte:

```json
"foregroundCommand": ["sed", "s/\u001b//"]
```

The next launch renders that back as `'sed' 's/ESC//'` and types it, and ESC starts a meta sequence in ZLE. The pane comes up like this:

```
greg@host ~ % 'sed' 's/^[//' /'
quote>
```

The command never runs. The pane sits at a continuation prompt waiting for a quote nobody will close, and the user has to notice and hit ^C. Verified end to end in an isolated instance (own `AGTERM_STATE_DIR`) on current master: type the line, `restore capture`, `kill -9`, relaunch.

Same shape for the other C0 bytes, each with its own damage. 0x09 triggers completion, so an `awk -F` with a literal tab restores with a filename spliced in. 0x03 and 0x15 kill the line outright.

This predates `restore capture` (#452), it is the quit-time capture too, and it needs no unusual input, only a program whose argument holds a control byte. Two directions, both reasonable:

- render an argument containing a byte below 0x20, or 0x7f, as `$'...'` with `\xNN` escapes. Exact and reversible, but `$'...'` is a bash/zsh extension, so a login shell that is `/bin/sh` would take it literally.
- refuse to capture such an argv and leave the pane a plain shell, the way the denylist already declines to restore what it cannot restore faithfully. Less faithful, never wrong.

Same class as #347 and #95, different sink.

ssgreg:
Adds `restore capture`, which fills the restore slot with every pane's running command now instead of at the next clean quit.

```
$ agtermctl restore capture
captured 1 pane
$ agtermctl restore capture --json
{"result":{"count":1,"text":"captured 1 pane"},"ok":true}
```

`restore clear` drops those captured commands and `session restore` pins one, but nothing created one. #447 took the shutdown and restart path off the table, agterm quits normally there and captures on its own. What stays uncovered is a force quit, a crash, a hard reset, a power loss, and each of them brings every pane back a plain shell.

It runs the same `captureForegroundCommands` the quit path runs and persists through the same `saveAllOpen`: same slots, replay stays launch-only and one-shot, the denylist still decides what re-runs, app-global so no `--window`. Bind it or run it from a scheduled job and an exit nobody was there for restores like a deliberate quit.

**It refuses while the master setting is off**, rather than writing argv a user who opted out will never see replayed:

```
$ agtermctl restore capture
error: "Restore running commands on restart" is off, nothing was captured
```

That is the one place this command does not follow `session.restore`, which succeeds with a note in the same state. A pin outlives the toggle and is worth keeping; a capture nothing will replay only goes stale.

**`count` is the slots the call itself wrote**, not a tally taken once it returns: a session whose split is hidden or gone can still hold argv from an earlier capture, and a tally reports that as fresh. The response carries its own `result.text` too, or the shared formatter prints the total as "N diagnostic(s)".

**Two invariants held by construction until a capture could run mid-run**, both fixed here. Closing a split left its captured command in the slot for the next ⌘D to fire. A non-last window close left argv in that window's file, which the never-windowless reopen fallback could replay. Unrelated but in the way: an unknown `cmd` now names itself in the error instead of answering "the data couldn't be read".

The two UI tests start their foreground `tee` through `session.type` rather than `typeText`. Event synthesis needs the machine running the tests to be allowed to post events, `session.type` needs nothing, and a capture test only wants a live foreground process. The existing typing tests are untouched.

Tested with `swift test` (2581), `make test-app` (250), `make lint`, `make build`, and `-only-testing:agtermUITests/RestoreCommandUITests/testRestoreCaptureSurvivesForceQuit` plus `testRestoreCaptureRefusesWhenTheSettingIsOff`. Checked on a live isolated instance rather than reasoned about: `sleep 12345` in a pane, capture reports 1, the window snapshot on disk carries `["sleep","12345"]`, then `kill -9`, and the relaunch re-runs it with the slot cleared after consumption.

Docs updated in `site/commands.html` and `site/docs.html`, the bundled skill (`SKILL.md`, `reference.md`, `examples.md`), `.claude/rules/control-api.md` and `.claude/rules/settings.md`, which record the gate, the refusal and why the count is what it is.

Rebased on master, so the capture inherits 131dfc6: an argv carrying a control byte or lossy UTF-8 is refused at replay and the pane comes back a plain shell. That closes #454 for an on-demand capture too, and the `site/docs.html` paragraph now states both that refusal and this command.

Follows #447.

### DIFF
--- a/.claude/rules/control-api.md
+++ b/.claude/rules/control-api.md
@@ -110,8 +110,16 @@ paths:
   may have quit. Its `stop()` returns early without unlinking, leaving the owner's socket intact.
 - One newline-delimited JSON request and response uses each connection, capped at 1 MiB. Unknown commands
   return structured errors. Mutations may return `result.id`; trees use `result.tree`.
+  A decode failure reports the `DecodingError`'s CONTEXT `debugDescription`, not `localizedDescription`, so
+  the error NAMES the rejected `cmd`. That is the only signal a caller gets that its agterm predates its
+  agtermctl, and the reason a new command needs no version handshake. Read the context, never the error:
+  `DecodingError.debugDescription` is macOS 26.4+, so at the 14.0 deployment target `String(describing:)`
+  degrades to a reflection dump that hides the sentence inside `DecodingError.Context(...)`.
 - Human output shows IDs only for created session/workspace/window, retains them in JSON, uses
   `result.affected` for session counts, and reserves `result.count` for diagnostics/search.
+  A command that reuses `count` for something else must carry its own `result.text`, which the shared
+  formatter prefers over every count spelling; `restore.capture` does, or its pane total would print as
+  "N diagnostic(s)".
 - Targets accept active, case-insensitive UUID, or unique prefix. Batch targets resolve within the first
   target's store, deduplicate, and fail atomically. Preserve the first target in the legacy top-level field
   so old servers degrade to it rather than active.
@@ -135,7 +143,8 @@ renumbering. Do not reintroduce a count anywhere.
 - `font.inc`, `font.dec`, `font.reset`
 - `window.new`, `.list`, `.select`, `.close`, `.rename`, `.delete`, `.resize`, `.move`, `.zoom`,
   `.fullscreen`, `.minimize`
-- `keymap.reload`, `keymap.list`, `config.reload`, `theme.set`, `theme.list`, `restore.clear`, `version`
+- `keymap.reload`, `keymap.list`, `config.reload`, `theme.set`, `theme.list`, `restore.capture`,
+  `restore.clear`, `version`
 
 `debug.appearance` is a private `Command` case, absent from the list above, used only by `AppearanceFlipUITests`.
 It accepts light/dark, sets `NSApp.appearance`, posts `.agtermSystemAppearanceChanged`, echoes the effective
@@ -619,6 +628,12 @@ side, and reads `lastAppliedIsDark` when bare. Refuse it outside XCUITest; provi
 
 ## Restore commands
 
+- `restore.capture` fills those same captured main/split slots on demand, from every open window's live
+  panes, saves immediately, and captures no hidden split. It is app-global. It reports the slots it
+  actually WROTE in `result.count` plus its own `result.text`; counting the slots afterwards instead would
+  read a stale split capture as a fresh one. It REFUSES while `restoreRunningCommand` is off, which is the
+  one place this API does not follow `session.restore`'s note-and-succeed: see [[settings]] for why the
+  two differ and for the exits the command exists for.
 - `restore.clear` clears captured main/split foreground commands across open windows and saves immediately.
   It never clears durable `initialCommand`; it is app-global.
 - `session.restore` pins per-session, per-pane next-launch behavior for discussion #264:
--- a/.claude/rules/settings.md
+++ b/.claude/rules/settings.md
@@ -126,14 +126,23 @@ paths:
   libghostty diagnostics across all sources, clear all session zoom, post appearance change, and notify
   non-zero diagnostics. A config-directory change reloads both co-located files. Launch also reports
   cached diagnostics.
-- **Restore running commands is opt-in, and both capture and replay are exit-scoped.**
-  `AppDelegate.captureForegroundCommands` runs at two points: `applicationWillTerminate` before
-  `saveAllOpen()`, and the LAST window's `willClose` before its surface teardown, which precedes
+- **Restore running commands is opt-in. Replay is launch-scoped; capture runs at two exits and on demand.**
+  `AppDelegate.captureForegroundCommands` runs at three points: `applicationWillTerminate` before
+  `saveAllOpen()`, the LAST window's `willClose` before its surface teardown, which precedes
   `applicationWillTerminate` and is therefore the only point where a close-the-last-window exit's
-  commands are still readable.
-  Guarded by `openIDs() == [windowID]`, skipped under `isTerminating`.
-  A NON-last close captures nothing: a launch restore can't tell that window's file from one open at
-  exit, so its argv could replay via the never-windowless reopen fallback.
+  commands are still readable, and `restore.capture` on demand, which exists for the exit that reaches
+  neither: a force quit, a crash, a hard reset, a power loss. A system shutdown/restart/logout is NOT in
+  that set — since #447 it reaches `applicationWillTerminate` like any quit — so do not re-motivate the
+  command with an OS update. The on-demand arm changes nothing else: it fills the same
+  slots, persists through the same `saveAllOpen`, and replay stays launch-only and one-shot.
+  All three arms are gated on the setting, and only the on-demand one SAYS so: it refuses while the
+  setting is off rather than capturing what nothing would replay. Deliberately unlike a `session.restore`
+  pin, which succeeds with an explanatory note in the same state, because a pin outlives the toggle and
+  a capture only goes stale.
+  The `willClose` arm alone is guarded by `openIDs() == [windowID]` and skipped under `isTerminating`.
+  A NON-last close captures nothing AND clears both persisted slots plus the pending pair: a launch
+  restore can't tell that window's file from one open at exit, so its argv could replay via the
+  never-windowless reopen fallback, and on demand a capture can now have written argv there mid-run.
   Argv comes from `ghostty_surface_foreground_pid`, `sysctl(KERN_PROCARGS2)`, and host-free parsing.
   Capture no hidden split.
   Strip login `-` before shell recognition; a known shell with only flags is idle and omitted, while
--- a/agterm/AppDelegate.swift
+++ b/agterm/AppDelegate.swift
@@ -333,23 +333,36 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
     /// Capture the given panes' foreground commands (main + split) into their `Session` fields for the
     /// snapshot save. `ForegroundProcess` returns nil for a pane at its shell prompt, so plain shells stay
-    /// plain. Two callers on different lifecycle edges: `applicationWillTerminate` passes every open
-    /// session, `WindowAccessor`'s `willClose` passes one closing window's — on a close-the-last-window
-    /// exit the quit-time capture runs after that teardown, too late to see any surface. Both sites and
-    /// the launch-only replay gate are stated in `.claude/rules/settings.md`.
+    /// plain. Three callers on different edges: `applicationWillTerminate` passes every open session,
+    /// `WindowAccessor`'s `willClose` passes one closing window's — on a close-the-last-window exit the
+    /// quit-time capture runs after that teardown, too late to see any surface — and `restore.capture`
+    /// passes every open session on demand. All three sites and the launch-only replay gate are stated in
+    /// `.claude/rules/settings.md`.
+    ///
+    /// Returns how many slots it actually WROTE a command into, which is what an on-demand caller reports.
+    /// Counting the slots afterwards instead would include a value this call never touched: the split slot
+    /// of a session whose split is hidden or gone still holds whatever an earlier capture put there.
     @MainActor
-    static func captureForegroundCommands(sessions: [Session]) {
+    @discardableResult
+    static func captureForegroundCommands(sessions: [Session]) -> Int {
         let shellBasename = ProcessInfo.processInfo.environment["SHELL"].map(CommandRestore.basename)
+        var captured = 0
         for session in sessions {
             if let view = session.surface as? GhosttySurfaceView {
                 session.foregroundCommand = ForegroundProcess.command(for: view, shellBasename: shellBasename)
+                if session.foregroundCommand != nil { captured += 1 }
             }
             // only a SHOWN split is recreated on restore, so gate on isSplit — a hidden split's captured
-            // command would sit stale until the next ⌘D fires it.
+            // command would sit stale until the next ⌘D fires it. Clearing it in the else keeps that stale
+            // value out of the snapshot now that a capture can run more than once per launch.
             if session.isSplit, let split = session.splitSurface as? GhosttySurfaceView {
                 session.splitForegroundCommand = ForegroundProcess.command(for: split, shellBasename: shellBasename)
+                if session.splitForegroundCommand != nil { captured += 1 }
+            } else {
+                session.splitForegroundCommand = nil
             }
         }
+        return captured
     }
 
     func applicationShouldTerminateAfterLastWindowClosed(_: NSApplication) -> Bool {
--- a/agterm/Control/ControlServer.swift
+++ b/agterm/Control/ControlServer.swift
@@ -90,6 +90,21 @@ final class ControlServer {
         }
     }
 
+    /// The sentence inside a `DecodingError`, read off its `Context` rather than off the error: the error's
+    /// own `debugDescription` is macOS 26.4+, so at this app's 14.0 deployment target `String(describing:)`
+    /// falls back to a reflection dump that buries the same sentence inside `DecodingError.Context(...)`.
+    /// Every case carries a context, and the `@unknown default` keeps a future case readable rather than
+    /// silent.
+    nonisolated private static func decodeDetail(_ error: DecodingError) -> String {
+        switch error {
+        case .dataCorrupted(let context), .keyNotFound(_, let context),
+                .typeMismatch(_, let context), .valueNotFound(_, let context):
+            return context.debugDescription
+        @unknown default:
+            return String(describing: error)
+        }
+    }
+
     /// Cap on a request line, shared with the client via `ControlWire` so the two sides can't drift; over
     /// it the line is rejected and the connection closed, so a bad client can't grow the buffer unbounded.
     nonisolated private static let maxLineBytes = ControlWire.maxRequestLineBytes
@@ -334,7 +349,12 @@ final class ControlServer {
         do {
             request = try JSONDecoder().decode(ControlRequest.self, from: line)
         } catch {
-            writeResponse(conn, ControlResponse(ok: false, error: "invalid request: \(error.localizedDescription)"))
+            // the decode CONTEXT over `localizedDescription`: the localized string is the generic "data
+            // couldn't be read", while the context names the offending value — for an unknown `cmd` that is
+            // the difference between a mystery and "Cannot initialize Command from invalid String value
+            // restore.capture", which tells a caller its agterm is older than its agtermctl.
+            let detail = (error as? DecodingError).map(Self.decodeDetail) ?? error.localizedDescription
+            writeResponse(conn, ControlResponse(ok: false, error: "invalid request: \(detail)"))
             return
         }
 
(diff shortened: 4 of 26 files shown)


umputun (changes requested):
two blockers, the rest is minor. The verb itself is right, `restore clear` and `session restore` had no counterpart that fills the slot.

first, the framing. The policy-driven macOS update case is already covered since #446/#447, your own fix: `QuitReason.isSystemQuit` returns `.terminateNow` for shutdown/restart/logout, so that exit reaches `applicationWillTerminate` and captures normally. What is left uncovered is a crash, `kill -9`, power loss. That is still worth a command, but it needs to be what the docs say.

**blocking**

`agterm/Control/ControlServer.swift:507-514` - the pane count goes out in `result.count`, and the CLI renders a bare count as `count == 0 ? "ok" : "N diagnostic(s)"` (`SocketClient.swift:205`). So three captured panes print `3 diagnostic(s)`, and with the setting off `text` wins and the count never prints at all. `.claude/rules/control-api.md:111` reserves `count` for diagnostics/search. Fix: build the human string and always set `result.text`, keep `count` for `--json`, same shape as `session.search` at `ControlServer+SurfaceIO.swift:445`.

`agtermCore/Sources/agtermCore/AppStore+Panes.swift:79-103` - `closeSplit` clears `splitRestoreCommand` and `pendingSplitRestoreCommand` but not `splitForegroundCommand`, while `closePrimaryPane:148` does clear it. That field used to be written only at teardown, this PR makes it live mid-run. Capture with a split running `tail -f app.log`, close the split, ⌘D a new one later, then crash: the next launch types `tail -f app.log` into a pane that never ran it. Fix: nil it next to the two pin clears at :92-93.

**minor**

`agterm/Control/ControlServer.swift:507-509` - `count` sums non-nil slots after the capture, not what the capture wrote, so a split captured earlier and since hidden or closed still counts. Your own doc comment at :502 says "main and shown split each counting one". Fix: gate the split term on `$1.isSplit && $1.splitForegroundCommand != nil`.

`MiscCommands.swift:65-74`, `reference.md:1169`, `SKILL.md:485`, `site/commands.html:2616` - all four lead with the macOS-update case above. Fix: drop that example, keep crash / `kill -9` / power loss.

`MiscCommands.swift:63-77` - the "run it from a scheduled job every so often" advice has no staleness note. A capture is only as fresh as its last run, so a pager or a build that finished since still re-runs after a crash. One clause covers it, plus a pointer at `restore clear` and `restore-denylist.conf`.

tests - dispatcher routing and CLI parse only, nothing reaches `captureRestoreCommands` and nothing renders its response, which is why the `diagnostic(s)` output went unnoticed. `restore.clear` has `restoreClearRoundTrips` (`ControlProtocolTests.swift:1305`) and `testRestoreClearSucceeds` (`ControlAPIUITests.swift:144`). The one that pays here is a hosted case in `ControlServerSessionActionsTests.swift`, it already stands up a real `ControlServer` over a real `WindowLibrary` and `SettingsModel`, so the count and the setting-off text are testable without a UI test.

**one question**

both existing capture arms gate on `restoreRunningCommand` (`AppDelegate.swift:322`, `WindowAccessor.swift:154`), this one captures regardless and reports it in `result.text`. Is that deliberate? It means argv gets persisted for someone who turned the feature off, and it replays if he flips the setting on before the next launch.

on the read-back you offered: skip it. Matching `restore.clear` is the right call, and whether those slots become a public read surface is a decision for both commands at once, not something to bolt onto this one.

ssgreg:
Both blockers and the count are already in the branch, and your line numbers say why they read as open: `ControlServer.swift:507-514` is `captureRestoreCommands` as it stood in the first push, before the gate moved the function down. c1fbdae went up at 10:50Z, the review is stamped 17:58Z, so GitHub attached the newer head to a read of 982698d. Head is now 0a8f9e5, with one thing added that you were right about.

**The count.** `ControlServer.swift:520-521` builds the sentence and always sets `result.text`; `count` stays for `--json`.

```
$ agtermctl restore capture
captured 1 pane
$ agtermctl restore capture --json
{"result":{"count":1,"text":"captured 1 pane"},"ok":true}
```

It also counts what the call wrote rather than the slots holding argv once it returns: `AppDelegate.swift:349-361` increments per assignment and nils the split slot when `isSplit` is false, so a hidden or closed split cannot inflate the number.

**closeSplit.** `AppStore+Panes.swift:95-96` clears `splitForegroundCommand` and `pendingSplitForegroundCommand` beside the two pin clears, pinned by `AppStorePaneTests.closeSplitDropsTheCapturedSplitCommand`.

**Your question.** Deliberate, and the opposite way round from what you read: with the master setting off the command now refuses and captures nothing.

```
$ agtermctl restore capture
error: "Restore running commands on restart" is off, nothing was captured
```

Exit 1. A `session.restore` pin outlives the toggle, so note-and-succeed fits there; a capture nothing will replay only goes stale, and persisting argv for someone who opted out was the part worth removing. The difference is recorded in `.claude/rules/settings.md`.

**The framing.** Gone from all four places you listed plus `site/docs.html`, which promised "a force-quit or crash captures nothing" with no way out. They now name a force quit, a crash, a hard reset, a power loss, and say a shutdown, restart or logout captures by itself since #447.

**Staleness.** That one was genuinely missing. `MiscCommands.swift:76` carries it after the scheduled-job advice, with the pointer to `restore clear` and `restore-denylist.conf`, and `reference.md` the same clause.

**Tests.** Hosted coverage over a real `ControlServer`, `WindowLibrary` and `SettingsModel` sits in `agtermTests/ControlServerRestoreCaptureTests.swift`: refuses with the setting off, writes nothing to disk in that state, and reports the pane count in its own text. I put it beside `ControlServerPickTests` and `ControlServerWorkspaceCommandsTests` rather than inside the 654-line `ControlServerSessionActionsTests`, since those two set the same per-feature precedent; say the word and I fold it in. Two UI tests cover the SIGKILL round trip and the refusal, and they start the pane's foreground through `session.type` instead of `typeText`: event synthesis needs the machine running the tests to be allowed to post events, `session.type` needs nothing, so they run where a typing test cannot.

`make test` 2581, `make test-app` 250, `make lint`, `make build`, both UI tests green.

Read-back skipped, and agreed on the reasoning: that is a decision for both commands together.

ssgreg:
Rebased onto master (2683dbd), single commit 8686a84.

That is also why the previous push showed no CI: the PR had gone CONFLICTING, so GitHub could not build the merge ref and never created a run. It is MERGEABLE again and the run is going.

The only conflict was `site/docs.html`, where your 131dfc6 sentence about a control byte or lossy UTF-8 starting a plain shell met my clause about `restore capture` covering the force-quit case. The resolved paragraph carries both. 131dfc6 also closes #454, and wider than I reported it, so the capture inherits that refusal at replay rather than needing its own guard.

`make test` 2601, `make test-app` 250, `make lint`, `make build`, both UI tests green on the new base.

umputun (changes requested):
head 0a8f9e5 breaks the clean quit path, and not just for `restore.capture` callers: every ⌘Q with the setting on now restores plain shells. `testRestoreReRunsForegroundCommand` passes on master and fails here at :51.

**blocking**

`agterm/Views/WindowAccessor.swift:156-167` - the new `else` fires whenever ANY of the three conditions on :153-154 is false, and one of them is `!library.isTerminating`. Clean quit order: `applicationWillTerminate` sets the flag, captures every pane, and `saveAllOpen`s it (`AppDelegate.swift:319-327`); `willClose` then runs with the store still loaded, since `closeWindow` no-ops under that flag, so `library.isOpen` on :147 is true; the else nils `foregroundCommand`, `splitForegroundCommand` and the pending fields, and `store.save()` on :167 writes nulls over what :327 just persisted. `testRestoreCaptureSurvivesForceQuit` cannot see it, `kill -9` never fires `willClose`.

three runs of that one test: master green, head red at :51, head plus `} else if !library.isTerminating {` green. That is also the fix, the terminating path has to keep the quit-time capture. Worth a regression case that exits with ⌘Q rather than SIGKILL, since nothing today covers this branch on the quit path.

**minor**

`agterm/Control/ControlServer.swift:519-522` - `saveAllOpen` goes through `AppStore.save`, which throws away `saveChecked`'s Bool (`AppStore.swift:920-922`), so the command answers ok and prints a count with nothing on disk. `AppStore.swift:924-926` states that contract and names `setRestoreCommand` as the caller keeping it today. A checked `saveAllOpen` does not exist yet, so this is more than a one-liner, your call whether the one command whose whole point is that the write reached disk should carry it.

`agtermTests/ControlServerRestoreCaptureTests.swift:53-60` - `testWritesNothingWithTheSettingOff` presets every `foregroundCommand` to nil, then asserts they are nil. A hosted store has no `GhosttySurfaceView`, so nothing can write one and the case stays green with the gate deleted. `testRefusesWithTheSettingOff` already pins the gate.

`agtermUITests/RestoreCommandUITests.swift:357` - `testWindowCloseCapturedCommandDoesNotReplayOnMidRunReopen` never calls `restore.capture`, so the new clearing branch is only ever seen nil to nil, and deleting the `else` leaves it green. Capture first, assert the argv reached the window file, then close.

merge - conflicts with master in `site/docs.html`, the rest auto-merges.

everything from round 1 checks out at head.

ssgreg:
The blocker is real and it was mine. Head is ab9dc3f.

`WindowAccessor.swift:156` is now `} else if !library.isTerminating {`, exactly your fix, and the comment records why the terminating path belongs to neither arm.

Two proofs, since "every ⌘Q restores plain shells" deserves more than a diff. First an isolated instance (`AGTERM_STATE_DIR`, own socket) quit by an Apple Event addressed to that pid alone, `sleep 12345` running in the pane:

```
before the fix:  foregroundCommand on disk: []
after the fix:   foregroundCommand on disk: [["sleep","12345"]]
```

Second a regression case on the quit path, `testRestoreCaptureSurvivesCleanQuit`: capture, `gracefulQuit`, assert the argv survived on disk, relaunch, assert it re-ran. It fails on the old code at the on-disk assertion.

I also owe you a correction on my last comment. I said the typing tests could not run on my machine; a stale `testmanagerd` from an earlier run was eating the synthesized events, and after restarting it the whole class runs: `agtermUITests/RestoreCommandUITests`, 23 tests, 0 failures, 220s, `testRestoreReRunsForegroundCommand` among them. Running that class is what would have caught this before you did, and it is on me that I concluded "environment" instead of digging. The `session.type` seam stays in the two new tests, since it is faster and needs no event synthesis, but not as a workaround for something broken here.

On the minors:

`saveAllOpen` swallowing the result: agreed, and worth doing rather than documenting. `WindowLibrary.saveAllOpenChecked()` attempts every store and reports whether all of them landed, `saveAllOpen()` is that with the result discarded so the two cannot drift, and `restore.capture` now answers `ok: false` with "captured N panes but the save failed, so nothing reached disk". `WindowLibraryTests.saveAllOpenCheckedReportsAFailedWrite` pins it with the unwritable-windows-directory lever your stale-file test already uses.

The vacuous hosted case: rewritten as `testTheRefusalLeavesAnEarlierCaptureAlone`. It presets `["sleep","12345"]` in every slot, so deleting the gate runs the capture over a store with no surfaces, assigns nil, and fails the test. An all-nil fixture could not.

`testWindowCloseCapturedCommandDoesNotReplayOnMidRunReopen`: it now calls `restore.capture` first and asserts the argv reached the window file before the close, so the clearing branch is exercised full-to-empty rather than nil-to-nil.

The `site/docs.html` conflict is gone: I rebased onto 2683dbd before this review arrived, and the resolved paragraph carries both your 131dfc6 sentence and this command. 131dfc6 also closes #454, wider than I filed it, so the capture inherits that refusal at replay.

`make test` 2602, `make test-app` 250, `make lint`, `make build`, `RestoreCommandUITests` 23/23.

umputun (commented):
two minors, both one-liners. Last round's items check out at ab9dc3f.

**minor**

`agterm/Control/ControlServer.swift:522-523` - "nothing reached disk" is false in two cases. `saveAllOpenChecked` returns false when ANY store fails, so with two windows open the first one's write can already be on disk. And the captured argv stays in the live `foregroundCommand` fields either way, so the next ordinary `store.save()` (session select, rename, split) writes exactly what the caller was told was never written. The `ok: false` itself is right, only the clause is wrong. Fix: say what happened, e.g. "captured N pane(s) but at least one window's save failed; the captured argv stays in memory and will be written by the next save".

`agtermTests/ControlServerRestoreCaptureTests.swift:53-64` - `testTheRefusalLeavesAnEarlierCaptureAlone` still does not pin the gate. Line 55 presets `foregroundCommand` only, and the fixture's seeded session has `surface == nil`, so `captureForegroundCommands` never reaches the assignment inside `if let view = session.surface as? GhosttySurfaceView`. Delete the guard at `ControlServer.swift:513` and this test stays green, which is the same shape as the all-nil version from round 2. The comment at :59-61 is wrong for the same reason: only `splitForegroundCommand` is written unconditionally, by the new `else`. Fix: preset `session.splitForegroundCommand = ["sleep","12345"]` as well and assert it survives, since that one does go red without the guard, then reword :59-61 to name the split slot as the one the else nils.

ssgreg:
Both right, both fixed at 497b7d3.

The error clause now says what happened: "captured N pane(s) but at least one window's save failed; the argv stays in memory and the next save writes it". Your two cases are exactly why the old wording was a lie: the verdict is an AND across stores, and nothing rolls the fields back.

The hosted case now presets `splitForegroundCommand` too and asserts that one first, since it is the slot the capture's `else` nils unconditionally for a session with no shown split. The main slot never reaches its assignment without a `GhosttySurfaceView`, so you were right that it pinned nothing, and the comment now says which slot does the work.

Checked by mutation rather than by reading: with the guard at `ControlServer.swift:513` deleted, `make test-app` reports 4 failures and `testTheRefusalLeavesAnEarlierCaptureAlone` is among them. With it back, 250 pass.

`make test` 2602, `make test-app` 250, `make lint`, `make build`.

umputun (commented):
two minors, both one-liners in files you already touched here. Round 3's items check out at 497b7d3, and the refusal test does pin the gate now: deleting the guard at `ControlServer.swift:509` turns it red on the split-slot assertion, so that oracle is real this time.

**minor**

`agterm/Control/ControlServer.swift:512` - an interactive capture records itself. While `agtermctl restore capture` blocks on the response it is that pane's foreground process group, so `command(for:)` reads its argv. `isIdleShell` keeps it (`argv[0]` is not a shell), the denylist has no built-ins and the seeded starter file is only `tmux`/`screen`/`zellij`, and `shouldRestore` refuses nothing here, so the slot gets `["agtermctl","restore","capture"]` and the pane comes back running it, printing a stray "captured N panes". It re-arms itself too: at launch the replayed invocation is the foreground group again when the capture it triggers runs. Typing it at a prompt is how anyone tries the command before wiring the keybind or the job. One narrower case behind it: a `session.new --command` pane whose command has already exited gets `hadForeground: true` from this, and `restorePlan` then suppresses the `initialCommand` that would otherwise re-run.

An agent calling it from inside a session is unaffected, a tool-spawned child never `tcsetpgrp`s, so the group leader is still `claude`/`codex`, which is the argv you want captured.

Fix: docs, not code. A built-in `agtermctl` filter contradicts "there is NO built-in list" and would suppress a legitimate `agtermctl events read --follow`, and per-connection pane attribution does not exist for an app-global command. One clause in `MiscCommands.swift:63-82` and in `reference.md`'s restore section: a capture typed at a prompt records the invocation in that pane, so prefer a keybind or a scheduled job, or `restore clear` after an interactive one.

`agterm/Control/ControlServer.swift:334` - the decode-error drive-by does not do what its comment and `.claude/rules/control-api.md:110-112` say. `DecodingError`'s `debugDescription` is `@available(macOS 26.4, *)` and we deploy to 14.0 (`project.yml:18`), so below 26.4 `String(describing:)` falls back to the reflection dump, `dataCorrupted(Swift.DecodingError.Context(codingPath: [...], debugDescription: "Cannot initialize Command from invalid String value restore.capture", ...))`. The rejected `cmd` is still in there so the point survives, it just arrives as a dump rather than a sentence. It renders as the sentence on 26.x, which is why it looks right locally. `swiftc -target arm64-apple-macosx14.0` on `e.debugDescription` is a hard error, so the rules bullet asserts something nobody can call at our deployment target.

Fix: pull the `Context` out of the four cases and use its `debugDescription`, version-independent and the same sentence on 14. Or keep `String(describing:)` and reword both the comment and the rules bullet to say so.

ssgreg:
Both confirmed and fixed at d93e894, rebased onto 94ab03f.

**The self-capture.** Reproduced before documenting it, in an isolated instance with the setting on, typing the command at the pane's own prompt:

```
greg@host ~ % .../agtermctl restore capture --socket /tmp/agterm-sc.yjxojk/agterm.sock
captured 1 pane

foregroundCommand on disk:
[["/…/agtermctl","restore","capture","--socket","/tmp/agterm-sc.yjxojk/agterm.sock"]]
```

Docs, as you said, not code: a built-in `agtermctl` filter would contradict the empty denylist and would take `events read --follow` with it. The clause is in `MiscCommands.swift`, `reference.md` and `site/commands.html`, each naming the pane it records, the keybind-or-job preference and `restore clear` after an interactive run. The `--command` corollary you raised follows from `hadForeground`, which `CommandRestore`'s own doc already states, so I left that as is.

**The decode detail.** You are right that it says one thing and does another. `ControlServer.decodeDetail` now pulls the `Context` out of the four cases and returns `context.debugDescription`, and the rules bullet says context, never the error, with the reason. Checked against the deployment target rather than against this Mac:

```
$ swiftc -target arm64-apple-macosx14.0 -typecheck a.swift
error: 'debugDescription' is only available in macOS 26.4 or newer
   func f(_ e: DecodingError) -> String { e.debugDescription }

$ swiftc -target arm64-apple-macosx14.0 -typecheck b.swift
   func g(_ c: DecodingError.Context) -> String { c.debugDescription }   # compiles
```

`project.yml:5` is `macOS: "14.0"`, so the old line was a 26.x-only sentence.

**On the rebase.** Four conflicts, all the same shape: `version` and `restore.capture` arriving in the same case lists and the same catalog line, both kept. `ControlServer.init` gained `identity`, so the hosted fixture takes `AppIdentity(version: "9.9.9", commit: "testsha")` like `ControlServerSessionActionsTests` does.

`make test` 2629, `make test-app` 255, `make lint`, `make build`, `RestoreCommandUITests` 23/23.

umputun (approved):
lgtm, merging.

both round 4 items check out at d93e894. `decodeDetail` pulls the `Context` out of the four cases, so a rejected `cmd` prints as a sentence at the 14.0 deployment target instead of a reflection dump. The self-capture clause is in the CLI help, `reference.md` and `site/commands.html`.

rebase onto 94ab03f is clean. `restore.capture` reaches all four sites a `Command` case needs, and `formatResponse`'s new `app` branch sits above `text`, so the round 1 count fix survives it.

nothing new blocking. Five rounds is a lot of back and forth, and the reproductions you attached each time are what kept it moving.

PR #461, opened 2026-08-18
fix(agent-status): guard the Claude hooks against a spawned agent repainting the spawner's row

x9x9x9x9x9x91:
The defect from the ownership-guard thread in [#456](https://github.com/umputun/agterm/discussions/456): a worker agent spawned from inside a session — a headless `claude -p` from a tool call, a second CLI agent — inherits the spawner's `AGTERM_*` environment, so its own four hooks address the SPAWNER's row. A coordinator's session goes green the moment a worker's Stop fires, and blinks with every worker tool call.

Per the discussion verdict: the guard lives in a Claude-specific adapter, the shared wrapper is untouched byte for byte, and ownership is decided from process topology, never terminal state.

**Commit 1 — the adapter.** `agterm-claude-status.sh`, beside the Codex adapter and following its shape: walks its ppid chain toward the pane boundary counting agent binaries, then delegates argv verbatim to the generic wrapper. Exactly one agent found (its own) → the pane's agent, report; a second → another agent spawned mine, stay silent. The walk passes *through* pty wrappers — `script`/`expect`/`unbuffer` add a pty but don't break parentage, so the spawner is still found — and a deliberately headless reporter with an exported session id has no agent ancestor, so it still reports. Both directions the rejected tty test got wrong come out right. It fails OPEN whenever the chain is unreadable or severed (a detached worker whose spawner already exited): the worst case is today's behavior, never a silent false-silence. Cost ~25 ms per hook fire on a ~1000-process machine.

**Commit 2 — the migration, without which this shouldn't merge.** An installer change alone reaches fresh installs only: `mergeClaudeSettings` skips an event whose entries already invoke us, and the Claude side has no `refreshManagedCodexBlock` counterpart — the same objection that routed the Stop classifier to the cookbook. The merge now first rewrites an entry whose command is byte-exactly what this installer generated for that event, swapping the wrapper path for the adapter path and touching nothing else; a hand-edited entry, an entry with extra flags, and a user hook that merely mentions the wrapper all fail the comparison and are left alone. The idempotency probe accepts either path, so a declined entry still reads as installed and never gains a stock duplicate beside it.

**Commit 3 — hermetic tests for the shipped script.** `ClaudeStatusHookTests` (5 tests) builds real process chains with `exec -a` per level and a recording stub as the wrapper, fixtures wrapped behind a `login` boundary — so a maintainer running `swift test` from inside a Claude session, with a real `claude` in the runner's own ancestry, can't have the owner case flipped. Covers owner-delegates-verbatim, nested-worker silence, a `node`-hosted agent counted through argv[1], the walk stopping at the pane boundary, and the outside-agterm no-op.

2605 tests in 97 suites pass; `swiftlint --strict` clean. shellcheck reports the same SC1007 false positive the shipped Codex adapter carries on its byte-identical `CDPATH=` line; left consistent with the sibling. The uncovered case is named in the shim's header rather than discovered later: a worker that outlives its spawner reparents to launchd and reads as an owner — fails open, and `env -u AGTERM_SESSION_ID` stays the advice for deliberate detachers (the tmux paragraph in `docs/troubleshooting.md` could name this carrier too; happy to add the sentence here if wanted).

Long-term, `foregroundPid` on the tree node (offered in #456) would delete the agent-name table and the severed-chain hole entirely; this PR is the fix that works without an app release.

### DIFF
--- a/agterm/AgentHooksInstaller.swift
+++ b/agterm/AgentHooksInstaller.swift
@@ -127,6 +127,8 @@ enum AgentHooksInstaller {
 
     // bake the bundled agtermctl's absolute path into the installed wrappers so the hooks fire even when the
     // CLI was never symlinked into PATH. The transform itself is host-free in `AgentHooksInstall`.
+    // `claudeWrapperName` is deliberately absent: the Claude adapter never calls agtermctl, it `exec`s the
+    // generic wrapper — which carries the baked path — so there is nothing to bake into it.
     private static func bakeAgtermctlPath() throws {
         guard let tool = bundledTool else { return } // no bundled CLI: leave the PATH fallback in place
         for name in [AgentHooksInstall.wrapperName, AgentHooksInstall.codexWrapperName] {
--- a/agterm/Resources/agent-status/agterm-claude-status.sh
+++ b/agterm/Resources/agent-status/agterm-claude-status.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+# Claude Code status adapter installed by agterm's Help ▸ Install Agent Status Hooks command.
+#
+# A worker agent spawned from inside a session (headless `claude -p` from a tool call, a second CLI
+# agent) inherits the spawner's AGTERM_* environment, so its own hooks would repaint the SPAWNER's
+# row. This adapter keeps that ownership question inside the installed hook package: decide from
+# process topology, not terminal state. The hook is a descendant of the agent that fired it, so
+# exactly one agent binary between here and the pane means "I am the pane's agent"; a second one
+# means another agent spawned mine, so stay silent. A tty test cannot answer this — a headless lane
+# that legitimately owns its pane has none, and a worker under script/expect gets a fresh pty anyway.
+#
+# It fails OPEN (reports) whenever the chain is unreadable or severed, e.g. a detached worker whose
+# spawner already exited: a missed guard is the behavior without this adapter, while a false silence
+# is a bug with no symptom. The argv is forwarded to the shared wrapper verbatim, so the adapter adds
+# a guard and changes nothing else; it never reads stdin, so a payload another hook wants is intact.
+set -u
+
+[ -n "${AGTERM_SESSION_ID:-}" ] || exit 0
+
+script_dir=$(CDPATH= cd -- "$(dirname -- "$0")" 2>/dev/null && pwd)
+status_wrapper=${AGTERM_STATUS_WRAPPER:-"$script_dir/agterm-agent-status.sh"}
+
+# exact names, not a glob: `claude*` also matches wrapper scripts such as claude-opus-worker.
+is_agent() {
+  case "$1" in claude | codex | kimi | opencode | pi) return 0 ;; esac
+  return 1
+}
+
+agents=0
+p=$PPID
+for _ in 1 2 3 4 5 6 7 8; do
+  [ -n "$p" ] && [ "$p" -gt 1 ] 2>/dev/null || break
+  line=$(ps -o ppid=,command= -p "$p" 2>/dev/null) || break
+  [ -n "$line" ] || break
+  line=${line#"${line%%[![:space:]]*}"} # ltrim the ppid column's padding
+  ppid=${line%% *}
+  cmd=${line#* }
+
+  base=${cmd%% *}
+  base=${base##*/} # argv[0] basename
+  base=${base#-}   # login shells report -/bin/zsh
+  base=${base#\(}
+  base=${base%\)}  # macOS renders (name) when argv is unreadable
+  case "$base" in # node/bun-hosted CLIs present as `node .../cli.js`, so test argv[1] too
+    node | bun | deno | python3 | python)
+      rest=${cmd#* }
+      hosted=${rest%% *}
+      hosted=${hosted##*/}
+      is_agent "$hosted" && base=$hosted
+      ;;
+  esac
+
+  is_agent "$base" && agents=$((agents + 1))
+  [ "$agents" -gt 1 ] && exit 0        # a second agent above mine: I am a worker, not the pane's agent
+  case "$base" in login | agterm) break ;; esac # reached the pane boundary
+  p=$ppid
+done
+
+exec "$status_wrapper" "$@"
--- a/agterm/Resources/agent-status/agterm-claude-status.sh
+++ b/agterm/Resources/agent-status/agterm-claude-status.sh
@@ -21,8 +21,12 @@ script_dir=$(CDPATH= cd -- "$(dirname -- "$0")" 2>/dev/null && pwd)
 status_wrapper=${AGTERM_STATUS_WRAPPER:-"$script_dir/agterm-agent-status.sh"}
 
 # exact names, not a glob: `claude*` also matches wrapper scripts such as claude-opus-worker.
+# The list is the union of the hook-driven agents and shell/integration.sh's AGTERM_AGENT_RE
+# default — one agent set split by detection mechanism, and the walk must count all of it: an
+# integration-driven agent (say gemini) that spawns `claude -p` is a spawner too, and missing it
+# here would let the worker's reset repaint the gemini pane. Keep the two lists in sync.
 is_agent() {
-  case "$1" in claude | codex | kimi | opencode | pi) return 0 ;; esac
+  case "$1" in claude | codex | kimi | opencode | pi | gemini | cursor-agent | aider | crush | goose) return 0 ;; esac
   return 1
 }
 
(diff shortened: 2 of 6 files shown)


umputun (commented):
two things on the ownership walk.

`is_agent` in `agterm/Resources/agent-status/agterm-claude-status.sh:25` covers `claude|codex|kimi|opencode|pi`, but `shell/integration.sh:29` defaults `AGTERM_AGENT_RE` to `gemini|cursor-agent|aider|crush|goose`, and its header says claude/codex/pi/opencode are left out of that one because their own hooks drive finer state. The two lists are one set split by detection mechanism, and the walk got half of it. `login -> gemini -> claude -p` counts a single agent and reports, so the worker's `completed --auto-reset` turns the gemini pane green mid-run. Same with aider, crush, goose or cursor-agent as the outer agent.

not a regression, the unguarded wrapper did this for every coordinator before. Fix is the five names in the case:

```bash
case "$1" in claude | codex | kimi | opencode | pi | gemini | cursor-agent | aider | crush | goose) return 0 ;; esac
```

I checked the reverse direction on a patched copy: `login -> claude` and `login -> zsh -> claude` still report, `login -> goose -> claude` goes silent.

second, `ClaudeStatusHookTests` has no case for the `script`/`expect` topology. That is the one the tty test was rejected over and the claim the walk rests on, so worth pinning: chain `["login","claude","script","claude"]` expecting silence. A chain long enough to exhaust the 8-hop bound, expecting delegation, would cover the fail-open half.

AnotherSava:
@x9x9x9x9x9x91 — another carrier for the ownership guard, live on macOS today (Claude Code 2.1.238).

```
72349  .../claude/versions/2.1.238 --resume .../beta/...       ??
72194  ClaudeCode.app/.../claude --bg-pty-host                 ??
4517   ~/.local/bin/claude daemon run --origin transient \     ??
         --spawned-by {"cwd":".../alpha/web","pid":93331}
93331  claude --continue                                       ttys005
90537  zsh -> login -> agterm
```

A desktop-app session in `~/Projects/beta`, parented by a transient daemon that alpha's agent spawned two days earlier. Paneless, tty-less, holding all seven of alpha's AGTERM_* values.

Neither documented remedy has an application point: the daemon is emitted by claude, not typed, and the only typed ancestor is alpha's own `claude --continue` — scrubbing there kills alpha's legitimate reporting. Pane restart doesn't clear it either, since the daemon outlives its spawner.

Your walk hits `claude` at 72194 and 4517 → silent; single-agent sessions here still count 1. The reparent case you flagged in the PR body doesn't bite: the second match is the daemon itself, so the exit precedes the parent deref.

Version-invariant — env-export source is blob `f75fd565` at v0.24.0, v0.25.0 and master, and no tag carries `agterm-claude-status.sh`.

Can hand over full `ps` dumps or the chain as a topology fixture. Hooks uninstalled here, so this is chain evidence, not an observed mis-paint.

umputun:
both round-1 points are answered. `is_agent` carries the ten names, and `workerUnderScriptStaysSilent` pins `["login","claude","script","claude"]` with the exhausted-walk and integration-driven cases beside it. Two code findings remain, one blocking.

**the walk cannot see a custom `AGTERM_AGENT_RE`**

`integration.sh:6-7` documents overriding the regex and uses `my-agent` as its own example, but `is_agent` at `agterm-claude-status.sh:29` is ten literals, so an override is invisible to the walk. `login -> my-agent -> claude` counts one agent and delegates, and the worker's `completed --auto-reset` repaints the my-agent pane. Reproduced with `AGTERM_STATUS_WRAPPER=/usr/bin/false` as a delegation sentinel: that chain exits 1, the `gemini` control exits 0.

the ten-name change covers the shipped defaults, but the documented override stays outside the classifier. A user who exports his own regex gets no coverage from it at all.

I do not think adding names fixes it, and consuming the existing regex has a compatibility wrinkle. The adapter is Bash, so it can match with `[[ =~ ]]` and needs no `grep`. Bash at `integration.sh:43` uses ERE. Zsh at `:33` uses ERE by default, though `RE_MATCH_PCRE` switches the same expression to PCRE. Fish at `integration.fish:32` uses PCRE2. So a PCRE-only override can be valid in the shell that set it and invalid in the adapter, and Bash returns 2 on an invalid regex, which is a case that has to keep delegating.

on the documented bash and zsh path the value already reaches the adapter. `integration.sh:29` uses `:=`, so an `export AGTERM_AGENT_RE=...` set before sourcing keeps its value and its export attribute, which is what the reproduction above used. The adapter never reads it. Fish is the one side that would also need plumbing, since `integration.fish:28` and its example at `:7` both use `set -g` rather than `set -gx`. Relaxing `is_agent` to a glob is out for the reason your own comment at `agterm-claude-status.sh:23` gives.

**the migration misses the pre-blink `UserPromptSubmit` entry**

`a9e678d9` changed that state from `active` to `active --blink`. The byte comparison at `AgentHooksInstall.swift:369` tests against the current `active --blink`, so an entry a pre-`a9e678d9` install wrote is not rewritten, `entryUsesWrapper` at `:403-408` then sees the generic wrapper and reports installed, and the merge adds no adapter entry. That hook keeps firing unguarded and nothing says so.

narrow: no tagged release carries the bare form. `17c8a914` introduced the hooks at 22:32, `a9e678d9` added `--blink` at 00:19, both first appear in v0.3.1, and no tag exists between them, so it is a source build from a window of under two hours. `PostToolUse` came later in `0d5fedb1` and is appended fresh, `Stop` and `Notification` migrate, `UserPromptSubmit` alone is left behind. One extra comparison covers it, and that is the whole historical set: those four states are the only variants across the file's history.

**two things outside the code**

the branch is conflicting against master, and not incidentally: `AgentHooksInstaller.swift`, `AgentHooksInstall.swift` and `agterm-agent-status.sh` have all changed there since `2683dbdc`, which is the installer you edit and the wrapper the adapter `exec`s. Worth updating before the next look, since the review above is against a tree that will not merge.

CI has not run on `c5c05f5b`. Run 32194734792 passed on `0df57725`, and the fix commit came after, so the ten-name case and the three new tests have never been through it.

**and the parts that hold**

the `script` topology, the hosted-argv branch, the pane-boundary stop and the exhausted-walk fail-open all check out against the tests. AnotherSava's chain traces correctly too, including the ordering claim: on the `4517` iteration the count reaches two and the script exits before `p=$ppid`, so a dead `93331` is never dereferenced. His pair is right for a reason worth knowing, though, and it is not obvious from the chain: `72349` is `.../claude/versions/2.1.238`, whose argv[0] basename is `2.1.238`, so `is_agent` does not count it. Harmless there because the two below it are plain `claude`, and I have no evidence a pane-owning session presents that way. Worth a thought when you touch the matching for the regex fix.

x9x9x9x9x9x91:
both in with 88092f2, rebased onto master first so it merges clean now.

went with consuming the regex like you sketched: is_agent falls through to an exported AGTERM_AGENT_RE, matched as ERE against the same argv[0] basename the literals see, so an override that inspects arguments won't extend the walk. an invalid expression returns 2 and lands back on the literal list, so the PCRE case degrades to fail-open instead of erroring. fish sets the variable with -gx now, that was the missing plumbing on that side. the migration accepts the bare active form for UserPromptSubmit and rewrites it onto the adapter and the current state in one go. both cases are pinned in tests that fail without their fix, and CI is green on the new head this time.

the 2.1.238 argv[0] I left alone for now, counting version-shaped names without evidence a pane-owning session presents like that felt like guessing. noted it for when the matching gets touched again.

umputun:
the fish side of the regex fix has a gap left, and nothing pins it.

**the export only runs when the variable is absent**

`integration.fish:29` guards with `if not set -q AGTERM_AGENT_RE`, so the `set -gx` on `:30` fires only for someone who has no override at all. A user who already set one in the form this file documented until this commit, `set -g AGTERM_AGENT_RE ...` before sourcing, satisfies `set -q`, keeps the export bit off, and the value still never reaches the adapter. `[ -n "${AGTERM_AGENT_RE:-}" ]` at `agterm-claude-status.sh:37` stays false and the walk falls back to the ten literals.

the value that does get exported cannot change anything either. The default is `^(gemini|cursor-agent|aider|crush|goose)([[:space:]]|$)`, and all five names are already in the literal `case` at `:36`. So the branch exports the value that needs no export and skips the one that does. The doc line moving to `set -gx` is still worth having for anyone copying the new form.

fix is to export on both branches, preserving an existing value rather than only setting the default.

**and no test catches it**

`shippedShellIntegrationsOmitLifecycleAgentsFromDefaultRegex` at `AgentHooksInstallTests.swift:475-481` accepts either spelling, and its other assertions are about the regex content, so reverting `:30` to `set -g` leaves the suite green. An `else` branch re-exporting the existing value adds a second line with the same `set -gx AGTERM_AGENT_RE ` prefix and breaks `#expect(defaultLines.count == 1)`, so key the filter on the default's literal regex rather than dropping the `set -g` alternative.

the migration change resolves the earlier blocking point. No tag carried the bare `active`, and a false-positive rewrite of an identical hand edit is recoverable from the `.bak`.

x9x9x9x9x9x91:
right, I exported the branch that didn't need it. both branches export now, the existing-value one by re-exporting what it finds, and I checked it against real fish this time: a set -g override from before sourcing comes through to a child process and the my-agent -> claude chain stays silent end to end, while the previous commit's file drops it to empty. pinned with a test on the re-export line, and the default-regex guard filter is keyed on the literal default like you suggested so the second set -gx line doesn't trip it. CI is green on 1e01674.

PR #462, opened 2026-08-18
docs(cookbook): add truthful-agent-lights recipe

x9x9x9x9x9x91:
The classifier from [#456](https://github.com/umputun/agterm/discussions/456), in the shape the verdict asked for: a cookbook recipe, with the wrapper dependency pinned and stated in Limits, and the sub-state color/shape vocabulary as something the reader opts into.

What it does: replaces the Stop hook's unconditional `completed --auto-reset` with a classifier over the tool subtrees still alive under the agent — a dispatched worker keeps the pulse, machinery gets the work tint, a lock queue reads as queued rather than working — and adds a launchd sweeper that enforces "no glyph means idle" in both directions: it clears a claim no process backs and re-lights a row that is quietly working. Stuck detection reads transcript progress, never hook silence, per the discussion.

Points from the discussion, so review can check them off:

- **Limits leads with the undocumented dependency**: the classifier keys on Claude Code's snapshot-wrapper shape, pinned to the version it was verified against (2.1.235), with the silent failure mode spelled out — if the wrapper moves, everything reads as "nothing running". The leaf-shape misreads are the second bullet (`rg` over a big tree reads as waiting, `tail -f` on a build log as work), with the lists placed where a reader edits them.
- **Per-call overrides beat Settings ▸ Agent Status** — said in What-it-does, in Limits, and in the variables file. Every color and shape is a variable, and blanking one posts the state *without* the override, so a reader whose Settings vocabulary collides opts out per shape rather than fighting the recipe.
- **The version floor is derived, not assumed**: 0.17.0, from `session status --shape` plus the `statusShape` read-back on `tree` (#292); everything else the recipe calls is older.
- **The spawned-worker repaint is named, not papered over**: a worker inheriting `AGTERM_*` writes the spawner's state files and paints its row — Limits states exactly which files and which self-heal. That inheritance is the stock-package defect from the #456 guard thread (fix proposed in #461), not something a recipe can fix.
- **State lives under `$HOME/.local/state`, and `XDG_STATE_HOME` is deliberately ignored**: the hooks see a shell-rc export, launchd does not, and honoring it would split the two halves into different directories on exactly the machines that set it. The reason is in the file where the variable is set.

On the "workflow you actually run" bar: the scripts these were distilled from have been driving ~20 concurrent agent sessions on the machine this was written on since the #456 filing. The recipe copies themselves were exercised against stub suites for every turn-end decision path and sweeper branch, plus a live read-only pass — the classifier over 14 live sessions across ten process topologies, the sweeper parsing the real tree with its writes recorded instead of issued. What has *not* run end to end is the recipe copies wired in as a session's actual hooks; saying that here rather than having it discovered.

### DIFF
--- a/cookbook/README.md
+++ b/cookbook/README.md
@@ -43,6 +43,7 @@ its *needs* column, so searching this page for `claude` or `kiro` finds those di
 | [kimi-agent-status](kimi-agent-status/) | Kimi Code sessions report agent status onto their sidebar row | 0.3.1, Kimi Code |
 | [kiro-agent-status](kiro-agent-status/) | Kiro CLI sessions report active, blocked, and completed onto their sidebar row | 0.7.1, Kiro CLI |
 | [status-announcer](status-announcer/) | demo: speak agent status changes from a dedicated session | 0.16.0, jq |
+| [truthful-agent-lights](truthful-agent-lights/) | the row reports what is still running after the turn ends, and stops claiming work that is gone | 0.17.0, jq, Claude Code |
 
 ### Panes, pickers and input
 
--- a/cookbook/truthful-agent-lights/README.md
+++ b/cookbook/truthful-agent-lights/README.md
@@ -1,0 +1,184 @@
+# Truthful agent lights
+
+A Claude Code session's row reports what is still running when the turn ends, and a sweeper clears every glyph nothing backs.
+
+## What it does
+
+The stock hooks map Stop to `completed --auto-reset` unconditionally, which is right for a session that finished and wrong for one that did not. Three cases go wrong in agent-heavy use:
+
+- **The turn ends, the work does not.** A session that started a test run in the background, or dispatched a second agent as a worker, goes green at Stop while that work runs on. Visit the row once and the glyph clears, so it is now indistinguishable from an idle shell.
+- **The agent dies, the glyph lives.** A killed or crashed agent leaves its last `active` pulsing forever, because a pushed status is trusted until something pushes another one.
+- **Thinking, machinery and waiting look identical.** A thirty-minute test run, a model composing a reply, and a session waiting on a lock are all `active`, so "is it working, waiting, or wedged?" cannot be answered from the sidebar.
+
+This recipe replaces the Stop hook with a classifier that looks at what is still alive under the agent process and picks the glyph that matches, and adds a scheduled sweeper that enforces the same truth from the outside, in both directions: it clears a claim no process backs, and re-lights a row that is quietly working.
+
+The other agent-status recipes — `kimi-agent-status`, `kiro-agent-status`, `container-agent-status` — wire one more agent's lifecycle hooks to the stock four-state vocabulary. This one goes the other way: it keeps that package installed and replaces only the decision Stop makes, then adds a sweeper that runs out of band, on a timer, with no hook involved. If you also run `status-announcer`, note that it speaks `blocked`, `completed` and `idle`: the sweeper's own clears are status changes like any other, so a row it corrects at three in the morning is a sentence spoken out loud.
+
+Sub-states are drawn with the per-call `--color` and `--shape` overrides:
+
+| glyph | means |
+|---|---|
+| pulsing default | a dispatched worker or subagent is still running, or the model is producing |
+| work color, running shape (square) | machinery is executing: tests, a build, a remote run |
+| work color, mixed shape (diamond) | machinery is executing *and* something else is queued behind it |
+| work color, queued shape (triangle) | only waiting: a lock queue, a poll loop, a log watcher |
+| stuck color, stuck shape (star), pulsing | the row claims to be working, nothing is running under it, and the transcript has not moved |
+| completed | the turn ended and the scan found nothing running |
+
+A per-call override beats what you picked in Settings ▸ Agent Status, for as long as that status stands. If a shape here collides with the vocabulary you configured there, change the shape variables in `lights-common.sh` rather than your settings — the recipe is the thing that should give way.
+
+## Requirements
+
+- agterm 0.17.0 or later — `agtermctl session status --shape` and the matching `statusShape` field on `tree` shipped there (#292, selectable status-glyph silhouettes picked per status in Settings or per call). Everything else the recipe uses is older: `--color` in 0.7.1 (#129), `--pane` and the injected `AGTERM_PANE` in 0.7.1 (#130), `--pane-id` in 0.13.0, the `statusBlink` read-back on `tree` in 0.10.0 (#169), and `--blink`, `--auto-reset`, `window list --json` and `tree --json` all predate the earliest tagged release.
+- `jq`, and a BSD userland: the scripts use `ps -axo`, `ps eww`, `stat -f %m` and `find -mmin` as macOS provides them.
+- Claude Code. The classifier reads a Claude Code implementation detail — see *Limits* — verified against 2.1.235.
+- The agent-status hooks package (Help ▸ Install Agent Status Hooks…), optional but recommended: when its `agterm-agent-status.sh` is present the recipe posts through it, so socket, pane and pane-id handling stay upstream's. Without it the recipe calls `agtermctl` itself.
+- `launchd` for the sweeper, or any scheduler that can run a command every couple of minutes.
+
+## Setup
+
+Copy the recipe somewhere stable and make the scripts executable:
+
+```sh
+mkdir -p ~/.config/agterm/truthful-lights
+cp truthful-agent-lights/* ~/.config/agterm/truthful-lights/
+chmod +x ~/.config/agterm/truthful-lights/*.sh
+chmod -x ~/.config/agterm/truthful-lights/lights-common.sh
+```
+
+`lights-common.sh` is sourced by the others, never run, so it is the one file that does not want the executable bit. Everything machine-specific lives in it as a variable with a default:
+
+| variable | default | what it is |
+|---|---|---|
+| `AGTERMCTL` | `agtermctl` | the CLI that talks to the control socket |
+| `AGT_LIGHTS_STATE` | `~/.local/state/agterm-lights` | pid notes, turn stamps, heartbeats |
+| `AGT_STATUS_SCRIPT` | the hooks package's `agterm-agent-status.sh` | the stock script to post through, when it exists |
+| `AGT_AGENT_PATTERN` | `claude\|codex\|kimi\|opencode\|pi` | command names that count as an agent |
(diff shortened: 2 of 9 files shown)


umputun (commented):
two worth fixing before this goes in, both in the sweeper. The rest are small. The #456 asks are all there, and the 0.17.0 floor holds against every flag and read-back field the scripts touch.

**the row can go green over live work**

`status-sweep.sh:308-314` - the "work drained" branch never checks that the scan found nothing, it fires whenever `work_state` came back empty, which includes machinery alive but under the `subtree_busy` cpu gate. End a turn with `rsync`, `pip install` or `docker pull` in a background tool call and the next sweep posts `completed --auto-reset` over it. Fix: gate on `[ "$total" -eq 0 ]`, and gate the `idle` branch at :315 the same way, otherwise the same case falls through and clears the row instead. README:181 needs rewording after that, it describes the outcome as leaving the row dark.

**a split session is classified from one pane**

`status-sweep.sh:85-100` - both panes share `AGTERM_SESSION_ID` and both have a tty, so `sort -n | awk '!seen[$1]++'` keeps the lower pid and the whole row is judged from that pane's tree. With an agent in each half, an idle pane A can green pane B's live work. Limits:179 doesn't cover it, the "lower pid is the pane's own agent" reasoning is about a pty-wrapped worker and doesn't hold when both are real pane agents. Fix: keep every pane-attached pid per session and combine the counts before deciding, no state file has to change shape. `annotate-claude-replies` keys by session and pane for the same reason.

**smaller**

`README.md:173` - "readable by you rather than by everyone on the machine" isn't what the code does, both creation sites use a bare `mkdir -p`, so 0755/0644 under the default umask. Fix: `(umask 077; mkdir -p ...)` at both, or drop that half of the sentence. `park-and-resume/agt-park.sh:11` does the former.

`README.md:68` - the socket wrapper has to append `--socket` after the subcommand. It's on `BasicOptions`, not the root command, so the obvious wrapper fails on every call and `set_status` swallows it with `|| true`. `new-session-in-workspace/agt-new-session.sh:32` has the working form.

`README.md:144` - "every reason it did nothing at all" isn't true, the no-op branches at :294, :322 and :325 log nothing. The next sentence in the same paragraph states it correctly. Fix: drop the clause.

`status-sweep.sh:222` - no `pipefail`, so the pipeline takes jq's status and jq exits 0 on empty input. A dead socket or a wrong `AGTERMCTL` wrapper gives an empty window list and a completely empty log, which is the failure a reader is most likely to hit. Fix: log when `$windows` is empty, ideally off the `agtermctl` call's own status so a closed agterm doesn't write a line every 120s.

`work-scan.sh:10-11` - the legend puts a long-lived ssh in `machinery`, the classifier at :116-118 puts every ssh/scp in `remote` or `waiting` and never in `realwork`. Limits sends readers into this file to edit the lists. Fix: drop that clause, the `remote` line below already has it right.

`status-sweep.sh:282-286` - the `own_turn_live` branch reposts `active --blink` whenever the row isn't blinking, which wipes what `machinery-paint.sh` just set, since that one posts no `--blink`. The PreToolUse glyph survives a single sweep. Fix: skip the repaint when the row already carries the work color, but keep the stuck-shape escape.

`README.md:49-64` - `AGT_LIGHTS_LOG` is defined in `status-sweep.sh:29`, not in `lights-common.sh` as the table's intro says. Fix: annotate the row the way `AGT_MACHINERY_PATTERN` is.

`README.md:3` and `:173` - "clears every glyph nothing backs" is unqualified, but `blocked` on a session with no recorded agent is never touched. Code is right there, clearing a prompt-for-input it knows nothing about would be worse. Fix: narrow the two claims, the script header at :17-20 has the wording.

`shellcheck` clean, modes right, index row correct, nothing personal committed. Nothing was executed here per CONTRIBUTING, so the snapshot-wrapper shape is unverified against a live 2.1.235, which Limits already calls out.

your Cookbook checks job has been running about an hour, looks stuck on the zsh apt install rather than on anything in the diff.

x9x9x9x9x9x91:
Thanks, both of those were real and both are fixed, along with the rest.

The green-over-live-work one was the worse of the two. The drained branch and the idle branch below it now both insist the whole scan came back empty, not just that `work_state` returned nothing, so machinery sitting under the cpu gate no longer falls through to `completed --auto-reset`. With a subtree alive and nothing proven about it the sweeper now posts nothing at all and leaves the glyph where it is, which is a new outcome, so it is in the header table and the Limits line about CPU evidence has been rewritten to describe what actually happens in both directions.

For the split session, discovery keeps every pane-attached pid per session now and a new `scan_session` adds the counts up before anything is decided. A pty-wrapped worker that also claims the session gets counted in too, and its subtree lands in the total twice, once under itself and once as a running worker under its parent. That overstates the work rather than understating it, which seemed the right way round for a light that means something is alive here, and the Limits bullet says so instead of the old pid-order reasoning.

The smaller ones: both state directories are created under `umask 077` now, so the README claim is true rather than aspirational. The socket wrapper in Setup is a worked example with `--socket` after the arguments, plus a note on why the obvious order fails silently. The "every reason it did nothing" clause is gone. The window list is read off `agtermctl`'s own exit status and logs the dead-socket case apart from the no-open-window case, both through a helper that stays quiet while the last log line already says the same thing, so a closed agterm writes one line rather than one every two minutes. The ssh clause is out of the `work-scan.sh` legend. The `own_turn_live` repaint is skipped while the row carries the work color, so the PreToolUse glyph survives, and the stuck shape still always gets the pulse back. `AGT_LIGHTS_LOG` is annotated the way `AGT_MACHINERY_PATTERN` is, and the two "clears every glyph" claims now carve out `blocked` on a session with no recorded agent.

On the version floor: I rechecked every flag and read-back field against current master and they are all unchanged, so 0.17.0 still holds. I did notice `statusChangedAt` on the tree session node from 0.25.0, which would answer most of the unbacked-glyph question in one field. I have not built on it, since the floor is lower and the field is not persisted across a restart, but there is a sentence in Limits pointing anyone adapting this at 0.25.0 or later at it.

Rebased onto master. Nothing was executed here either. One thing worth knowing: the Cookbook checks job on the previous push hung on the zsh and fish apt install and never reported, so that run told us nothing about the diff.

umputun (commented):
one thing to fix before this goes in, the rest are notes. Last round's items check out, apart from the split-pane claim below.

**Setup undoes #461 on 0.26.0 and later**

`README.md:108` / `lights-common.sh:30` - since 0.26.0 the installer wires the Claude hooks to `agterm-claude-status.sh`, which stays silent when a spawned `claude -p` worker fires them. Setup tells the reader to replace those entries with `set-status.sh`, and `AGT_STATUS_SCRIPT` defaults to the generic `agterm-agent-status.sh`, so the guard is gone again. A worker's posts land on the spawner's row, and because it inherits the pane, its `active` or `completed` replaces the spawner's `blocked`. Fix: default to `agterm-claude-status.sh` when it is installed, fall back to the generic script for older versions, keep the `AGT_STATUS_SCRIPT` override. The adapter only guards status posts, the worker still writes the pid, transcript and turn-start notes, the turn-end stamp and the heartbeat, so rewrite Limits `:186` to say that rather than dropping it, and adjust `:35`.

**notes, not blocking**

`README.md:187` - the counts are combined now, but the turn stamps, the transcript note and the Stop hook are still per session. With an agent in each half, pane A's Stop posts `completed` over pane B's live turn, and its `turnend` stamp makes the drained branch at `status-sweep.sh:397` repeat that on any sweep that catches B between tool calls. No need for per-pane state in a recipe, just narrow the claim: only the process counts are combined, and a turn ending in one half can green the other.

`status-sweep.sh:368` - the repaint skip recognises the machinery glyph by color only. With `AGT_WORK_COLOR` blanked, which the recipe documents as the opt-out, `machinery-paint.sh` posts a shape with no color and the next sweep during the turn wipes it with `active --blink`. Fix: also count the row as painted when `shape_now` is one of the running, mixed or queued shapes.

`status-sweep.sh:430-442` - the tree is read once per window and the writes come after the per-row scans and cpu samples, a few seconds later. A permission prompt posted in that gap gets replaced by the sweep's `active`, the same goes for the active-branch writes. Skipping `set_status` when `hb/<sid>` is newer than the sweep start narrows it, it won't close it.

x9x9x9x9x9x91:
The adapter one was a real regression on my side, fixed.

`AGT_STATUS_SCRIPT` now resolves to `agterm-claude-status.sh` when it is installed and falls back to `agterm-agent-status.sh` when it is not, and an explicit override still wins. Checked all four cases by sourcing the file against a fake `~/.config/agterm/agent-status`. Setup no longer says the repointing costs the guard, since the recipe posts through the same adapter the stock entries named. The Limits entry is rewritten the way you describe: the adapter guards the status post and nothing else, so a worker still overwrites the pid and transcript notes, stamps a turn start and a turn end, and touches the heartbeat on every post, which is why the sweeper reads the spawner's hooks as alive while the worker runs. Requirements says which script is meant on which version.

The split-pane claim is narrowed to what holds: only the process counts are combined, the stamps, transcript note, heartbeat and Stop hook are per session, and a turn ending in one half can green the other, including the repeat from the drained branch on a sweep that catches the other half between tool calls.

The repaint skip now also counts a row as painted when `shape_now` is the running, mixed or queued shape, so a blanked `AGT_WORK_COLOR` no longer gets its machinery glyph wiped on the first sweep of the turn. An empty shape never matches, so a blanked shape variable cannot make an unpainted row look painted.

For the write gap, `set_status` skips the row when its heartbeat is at or after the sweep's start time and logs that it did. Every hook post touches the heartbeat before it posts, so that is the signal that something changed the row after the tree was read. The comment says it narrows the window rather than closing it: a post between the check and the CLI call still loses, and a session without the recipe's hooks writes no heartbeat to check.

Nothing executed against a live agterm here either, shellcheck and the resolution and skip paths were exercised by hand.

umputun (approved):
lgtm. Two notes for whenever you next touch this, neither blocking.

`status-sweep.sh:367` and `:395` - the stuck-shape comparisons never got the guard the work-shape `case` right above them has, so with `AGT_SHAPE_STUCK` blanked every row carrying no shape override compares equal to the stuck mark, and the branch pulses over the machinery tint anyway. `[ -n "$AGT_SHAPE_STUCK" ] && [ "$shape_now" = "$AGT_SHAPE_STUCK" ]` at both sites, and `:416` is already right as it stands.

`README.md:186` - "never replaces the spawner's glyph" is stronger than the adapter promises: it fails open and posts when the chain above the worker is unreadable or gone. The reachable case is a spawner that already exited, not `nohup` or `disown`, which don't change PPID. The eight-hop bound is not in play, the deepest chain here is six.

last round's items check out.

PR #464, opened 2026-08-19
cookbook: add copilot-agent-status recipe

rychkov:
Add a cookbook recipe wiring GitHub Copilot CLI's native hooks to the bundled agent-status wrapper, giving Copilot CLI sessions the same active/blocked/completed sidebar glyph the installer already provides for Claude Code, Codex, Pi, and OpenCode. Co-authored-by: Copilot


umputun:
merged.

which Copilot CLI version did you run this against, and did you see all four states fire in agterm, blocked in particular? `cookbook/CONTRIBUTING.md` asks for a recipe you actually run, and the README records a check against the published hooks reference rather than a session. Every sibling names a version - `tested on kimi-code 0.31.0`, `tested on kiro-cli 2.16.1`. Send the version and I'll put it in *Requirements*.

two README bits are mine, follow-up: *Requirements* should say what shipped in 0.7.1 the way the container and kiro recipes do, and *Limits* is missing the no-prompt case - under `--allow-all-tools`, `--yolo` or `-p` nothing asks for permission, so `notification` never fires and the row goes active to completed with no blocked in between.

the JSON ships as a file, which the layout rule in `cookbook/CONTRIBUTING.md` did not allow. Amended the rule rather than asking you to inline 34 lines a reader can mistype.

rychkov:
I'm using `GitHub Copilot CLI 1.0.80`

PR #479, opened 2026-08-23
feat(nav): add cmd+1..cmd+9 session slots

stpkys:
⌘1–⌘8 select the 1st–8th session in the sidebar's flattened order; ⌘9 selects the 9th, or the **last** session once there are more than nine. A slot naming no session leaves the selection alone.

Numbering is global across workspaces and uses the same `navigableSessions` scope as ⌥⌘↑/↓, so flagged mode and the focus filter confine the slots identically. Landing behavior is the arrow path (`navigatePlain(.slot(n))`).

### Surfaces

- Nine rebindable built-ins `select_session_1`…`select_session_9`, default ⌘1…⌘9.
- No menu item and no palette row — the fuzzy session palette is the by-name surface. Chords are matched by `CustomCommandRunner` like `toggle_fullscreen`'s and run by `paletteLessHandler`; conflict model and `uiActionsEnabled` gate unchanged.
- `agtermctl session go --to 1`…`9`.

### Notes

- Existing keymaps are safe: a file already binding ⌘3 keeps it and `select_session_3` stays keyless. The two ad-hoc ⌘⇧D/⌘⇧G compatibility blocks became one `vacatesNewDefault` helper the slots share.
- `SessionSlot.index` owns the overflow rule: slot 9 means "last" only past nine sessions.
- `AppStore.swift` was at the 1000-line limit, so the navigation vocabulary types moved unchanged into `SessionNavigation.swift`.
- Docs updated: `site/docs.html`, `site/commands.html`, the bundled skill, and the `keymap` / `menu-actions` / `control-api` rules.

### Testing

- `swift test` — 2632 pass. `make lint` clean, `make build` succeeds.
- `make test-app` — passes except `AgentHooksInstallerTests.testAlertWithoutDocsKeepsTheSingleDefaultButton`, which fails on master too (Xcode 26.3 `NSAlert` reports no buttons when none were added).
- XCUITest **not yet run**: `SessionNavUITests/testSessionNavigationFollowsFocus` and `ControlAPIUITests/testSessionGoJumpsToASlot`.

### DIFF
--- a/.claude/rules/control-api.md
+++ b/.claude/rules/control-api.md
@@ -189,8 +189,11 @@ side, and reads `lastAppliedIsDark` when bare. Refuse it outside XCUITest; provi
   `--grow-left|right|primary|split|top|bottom` delta, defaulting an unset ratio to 0.5. Require a split,
   clamp through store limits, persist, then post the object-scoped live-divider notification. Hidden split
   stores for next show. Return clamped ratio as `%.3f`; read `splitRatio`.
-- `session.go --to next|prev|first|last|next-attention|prev-attention` operates on current selection in
-  the placement store, wraps within filtered scope, and returns selected ID. It has no target.
+- `session.go --to next|prev|first|last|next-attention|prev-attention|1-9` operates on current selection in
+  the placement store, wraps within filtered scope, and returns selected ID. It has no target. A digit is
+  the absolute slot the ⌘1…⌘9 chords select, one vocabulary with them ([[keymap]] owns the rule); a slot
+  naming no session leaves the selection where it was and still answers ok with it, exactly as a one-session
+  `next` does.
 - `notify` requires body, defaults title and session, skips OSC focus suppression, increments unseen, and
   uses click identity. When banners are disabled, return `ControlNotify.bannersOffNote` (#286); delivered
   requests omit text. Read through `unseen`.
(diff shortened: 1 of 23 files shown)


umputun:
saying this while it is still a draft rather than waiting for you to mark it ready, so you don't sink the remaining XCUITest work into something I would most likely reject.

the preference is to use the existing extension mechanics wherever they reach. Several issues and PRs have been answered that way already: if a keymap custom command can express it, that is where it goes, and the app stays out of it. `tree` plus `session select` reaches this one, so nine new built-in actions with nine shipped defaults is not a trade I want to make.

the whole feature is a script and nine one-line bindings:

```sh
#!/bin/sh
# agt-slot.sh N — select the Nth session in the sidebar's flattened order
id=$(agtermctl tree --json --window "$AGT_WINDOW_ID" --socket "$AGT_SOCKET" |
  jq -er --argjson n "$1" '
    .result.tree as $t |
    (if $t.sidebarMode == "flagged"
     then [$t.workspaces[].sessions[] | select(.flagged)]
     else ([$t.workspaces[] | select($t.workspaceFilter != true or .focused == true)]
           | (if length == 0 then $t.workspaces else . end) | [.[].sessions[]])
     end)[$n - 1].id | select(type == "string")
  ') &&
agtermctl session select --target "$id" --window "$AGT_WINDOW_ID" --socket "$AGT_SOCKET"
```

```conf
command "session 1" cmd+1 ~/bin/agt-slot.sh 1
```

that jq is the `navigableSessions` rule, so the slots stay confined by flagged mode and the focus filter exactly as ⌥⌘↑/↓ are. An absent position exits non-zero and the selection is left alone. Pick a different rule if you want different behavior, which is the part the app should not be deciding for you.

if you want to share your version, [cookbook/](https://github.com/umputun/agterm/tree/master/cookbook) is where this kind of thing lives, and `cookbook/CONTRIBUTING.md` has the format.

PR #488, opened 2026-08-27
feat(cookbook): add the close-tab-when-done recipe

andr81:
## What is the problem?

A tab that finished its task keeps a finished agent in it. Closing it is manual, and the sidebar fills up with tabs whose only remaining content is a reply you have already read.

Asking the agent to close its own tab does not solve it. The close runs from inside the turn, so the final reply is killed half-drawn, and it happens only if the model remembers to run the command and is allowed to.

## How does this solve it?

A chord marks the session; the agent's `Stop` hook is what closes it. The hook runs after the turn ends, so the reply is complete on screen first, and no instruction, tool permission or model cooperation is involved. The marker is an empty file named after the session id, consumed on the first turn that ends, with a `⏻` in the sidebar name while it is armed.

The recipe is one POSIX `sh` script plus a README and the index row, in the `Agent status and workflows` group.

## Notes for review

Things the recipe deliberately does, since they are the parts worth checking:

- The session id is validated as a UUID and upper-cased before it is used as a file name. It is an argument the README tells you to pass, so `on ../../notes.md` is a plausible typo, and agterm reports ids upper-case, so a hand-typed lower-case id would otherwise arm a marker the hook never looks for.
- `arm` is fail-closed: it proves the CLI, socket and `jq` answer before writing a marker, and removes the marker if the badge rename fails. An unbadged marker is a tab that closes with no warning.
- `fire` exits 0 on every path and is a silent no-op outside agterm. Removing the marker is what claims the close — plain `rm`, not `rm -f`, so a disarm racing the hook wins.
- The close is detached with `nohup` and re-enters the script, because `session close` kills the hook's own parent.
- A run hosted by Claude Code's daemon (`claude daemon run`, `bg-spare`, `bg-pty-host`) does not fire: such a worker keeps the `AGTERM_SESSION_ID` of whichever tab started the daemon, which is not the tab it runs in. Verified live — a background agent launched with a victim tab's id left that tab open, and a probe from inside that context reported the ancestor chain and the guard's verdict.
- The badge walk reads every open window, since `tree` reports only the frontmost one while `--target` addresses ids globally; `window list` also reports closed windows, whose tree is null.

`Limits` states the destructive behavior plainly: firing kills every process in the session with no undo, and a `Stop` at the end of a turn that ended with a question to the user closes the tab with the question on screen.

Ran `shellcheck`, `sh -n` and the cookbook CI steps (layout, index both directions, headings, shebang) locally; all green. It is installed in my own setup in exactly the form committed here — same script, same chord, same hook line — rather than written for the collection.

### DIFF
--- a/cookbook/README.md
+++ b/cookbook/README.md
@@ -39,6 +39,7 @@ its *needs* column, so searching this page for `claude` or `kiro` finds those di
 | [claude-clear](claude-clear/) | one chord sends /clear to the pane's Claude Code run, and nothing when it is not running | 0.13.0, python3, Claude Code |
 | [claude-conversation-picker](claude-conversation-picker/) | pick a past Claude Code conversation by what it was about and resume it in the pane | 0.21.0, python3, Claude Code |
 | [claude-recap](claude-recap/) | one key lists what the Claude Code run in a session was working on | 0.10.0, zsh, jq, Claude Code |
+| [close-tab-when-done](close-tab-when-done/) | arm a tab with a chord and it closes itself when the agent stops replying | 0.22.0, jq, Claude Code |
 | [container-agent-status](container-agent-status/) | a containerized agent reports status onto its sidebar row via a TCP notification to the host | 0.7.1, nc, timeout, Claude Code |
 | [copilot-agent-status](copilot-agent-status/) | Copilot CLI sessions report active, blocked, and completed onto their sidebar row | 0.7.1, Copilot CLI |
 | [kimi-agent-status](kimi-agent-status/) | Kimi Code sessions report agent status onto their sidebar row | 0.3.1, Kimi Code |
--- a/cookbook/close-tab-when-done/README.md
+++ b/cookbook/close-tab-when-done/README.md
@@ -1,0 +1,128 @@
+# Close tab when done
+
+Arm a tab with one chord and it closes itself the moment the agent stops replying.
+
+## What it does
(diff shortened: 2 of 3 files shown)


umputun:
merged, thx. Recipe reads well and Limits does what that section is for.

one thing worth a follow-up, not a blocker.

`arm` can write a marker with no badge on it, which is the case its own comment says must never exist. `session_name` deliberately separates a read failure (returns 1) from a session that simply is not there (returns 0 with no output), and then `badge_on` folds the second case back into success:

```sh
case $name in
"$BADGE"* | '') return 0 ;;
esac
```

so `arm` writes the marker, badges nothing, and exits 0 with no output. `status <that-id>` then reports `on` while the sidebar shows no `⏻`.

the reachable path is the id form Usage recommends for scripts and agents. One wrong digit still passes `canonical_id` and the liveness probe, so you get a marker naming no session. That one is inert.

the destructive direction is narrower. `session_name` excludes closed windows with `select(.open)`, so a valid session id from one comes back as not found and `arm` still leaves its marker behind. After the window is reopened and an agent finishes a turn in that restored session, the tab closes without the badge warning.

fix is small: with a non-empty badge, treat empty `session_name` output as a failure, so `arm` removes the marker and reports the session left disarmed. An intentionally empty badge keeps its documented invisible behavior.

no rush, it is already in.

PR #490, opened 2026-08-27
refactor: let Session own the captured-command reset

ssgreg:
Moves the captured-command reset into `Session`, so `restore.clear` and the non-last-window close stop spelling out the same four fields inline. No behaviour change.

This is the `capture-slot-reset-hand-rolled-in-two-places` backlog item, and the second copy is mine: #452 added it. Your note says the scattering is what made `closeSplit` miss `splitForegroundCommand` in the first place, which is the argument for doing it now rather than at the third slot.

**Session owns the pair-of-pairs.** `clearCapturedForegroundCommands()` sits beside the `clearPendingForegroundCommands()` it calls and drops the persisted pair with the pending one. The doc line names why they go together and points at the rules file instead of restating it.

**Both call sites keep their scope.** `ControlServer.clearRestoreCommands` still walks `library.allOpenSessions()`, the `willClose` arm still walks the closing window's store, and neither gained or lost a field. `.claude/rules/settings.md`'s sentence about who cancels an armed replay stays true as written, so no rules change: only the mechanism moved.

**Test and its mutation.** `SessionTests.clearCapturedForegroundCommandsDropsThePersistedAndPendingPairsTogether` presets both pairs plus a `session.restore` pin, then asserts the persisted fields are nil, both pending slots take as nil, and the pin survives. Removing the pending call from the new method turns it red on the split slot:

```
✘ Test clearCapturedForegroundCommandsDropsThePersistedAndPendingPairsTogether() recorded an issue at
  SessionTests.swift:564:9: Expectation failed:
  (session.takePendingForegroundCommand(pane: .right) → ["tail", "-f", "/var/log/x"]) == nil
```

`make test` 2632, `make test-app` 258, `make lint`, `make build`. For UI I ran the two cases that exercise the call sites, `RestoreCommandUITests/testWindowCloseCapturedCommandDoesNotReplayOnMidRunReopen` and `ControlAPIUITests/testRestoreClearSucceeds`, both green. The full `RestoreCommandUITests` class I ran green twice today, but on the base before #485 and #487, not on this one.


umputun (approved):
lgtm

PR #492, opened 2026-08-27
Reduce memory usage

paul-nameless:
Bumps the ghostty pin to 683d8db, which carries upstream's hidden-surface GPU release (ghostty#14017), and starts reporting occlusion for panes that aren't actually on screen. Hidden shells stay live while their Metal swap chains get released after a 1s reparenting grace. A new deckOnScreen flag, wider than deckVisible, keeps dashboard cells and passive HUDs painting.

The pin bump also brings the new multi-mime clipboard API, so the clipboard callbacks are rewritten against it. Denying a request now uses ghostty_surface_deny_clipboard_request instead of the old empty-confirmed-read workaround.

Measured on a Debug build with 12 sessions all printing output every 0.5s: 1.1 GB peak while every surface was realized, settling to 199 MB once the 11 hidden ones released (IOSurface resident down to 62 MB). Previously hidden surfaces held their swap chains forever.

This is the hidden-surface GPU cost from discussion #196, fixed without carrying a ghostty patch now that upstream released GPU resources for hidden surfaces itself.


umputun (changes requested):
glad to see this moving, the problem is real. The approach is right, and it is the outcome I argued for in #196: upstream owns the release, we just report occlusion, no ghostty patch to carry. I checked `683d8db` against the two renderer bugs from that thread and both are fixed there, `swap_chain_rebuilt` forces a redraw and `size_changed` updates the screen-size uniforms from the current surface size before the rebuilt frame is uploaded. So a quiet pane repaints and a hidden-resized one comes back at the right geometry. The deny migration to `ghostty_surface_deny_clipboard_request` is correct too.

three things need changing before this can merge.

**1. summoning the quick terminal occludes panes that are still visible.** `WindowContentView+Detail.swift:53` builds `visible = deckInteractive && isActive && !hideForOverlay && !quickTerminal.holdsKey`, and that expression reaches `deckOnScreen` because the pane host (:183), the scratch host (:87) and the pane-overlay host (:276) pass no `onScreen:`, so `TerminalView` falls back to `deckOnScreen = deckVisible`.

`holdsKey` is not a visibility term. `QuickTerminal.swift:21` documents it as focus ownership and says the panel stays on screen after agterm loses key, and `targetFrame` at :228 makes it a centered inset rectangle sized by `QuickTerminalMetrics` (configured 40-90%, or the capped default), non-activating, nothing ordered out.

so: pane running `tail -f`, hotkey, type in the panel for more than a second. `holdsKey` flips, `deckOnScreen` goes false, the 1s grace expires, and `ghostty_surface_set_occlusion(surface, false)` releases the swap chain for a pane the user is looking at, including panes in other windows and on a display the panel does not touch. A program using DEC mode 2033 is also told it is not visible, `Surface.occlusionCallback` queues a visibility report and `Termio.visibilityReport` sends `CSI ? 999 ; 2 n`. It recovers on dismiss, which is the only reason this is not critical.

your own line in `.claude/rules/libghostty.md:34` says to drive this from `deckOnScreen` rather than `deckVisible`. Two hosts that genuinely differ got the override, this third case did not. The new doc at `TerminalView.swift:38` ("nil follows `deckVisible`, which is correct for every interactive pane host") is the part that is wrong.

**2. Kitty clipboard writes of non-text are dropped and reported as succeeded.** `writeClipboard` keeps the first `text/plain` and returns for anything else. `write_clipboard_cb` is void, so `setClipboard` cannot fail back, and `Surface.zig:6004` sends `kittyClipboardStatus(.write, .DONE)` right after it, with `EIO` only on a Zig-side error we can never produce. `clipboard-write` defaults to `allow` and `kittyClipboardWrite` gates on nothing else, so an image write reaches us on a stock install, is discarded, and the program is told it worked.

the `ask` case is worse. `completeKittyClipboardWrite` reaches the confirmation callback on `UnauthorizedPaste`, we map Kitty write to `.write`, the user approves, the completion goes back with `confirmed = true`, and core then calls the void write callback with the authoritative representations, which we drop. So the user approves and still gets a false `DONE`.

the text-only assumption came from the old code, where it was fine because OSC 52 is always text. Kitty write is new at this pin, so the assumption needs to go with it. There is no cheap version of this, the callback is void and we cannot fail the request once ghostty has it, so the fix is to write every representation to the addressed pasteboard. Ghostty.app at the same pin maps each MIME through `NSPasteboard.PasteboardType(mimeType:)`, falls back to the raw MIME identifier, then calls `declareTypes` and `setData` per representation, which is a working shape to copy.

**3. an OSC 52 read gets no reply when the clipboard holds no text.** `pasteboardText` ends `flatMap { !$0.isEmpty ? $0 : nil }` (`GhosttyCallbacks.swift:195`), so an empty pasteboard is nil, `contents` comes back empty, and `readClipboard` returns `GHOSTTY_CLIPBOARD_READ_UNAVAILABLE`. Upstream discards that for OSC 52, `_ = try self.startClipboardRequest(.standard, .{ .osc_52_read = clipboard })`, and unavailable means the request never started, so nothing is written back to the pty.

the old code did `readPasteboardText() ?? ""` and always completed. Upstream's own `completeClipboardReadOSC52` says it replies even when empty because the client is waiting, and Ghostty.app keeps an empty string as zero-length data rather than dropping it. A client waiting for the OSC response gets none and has to time out.

to see it: `pbcopy </dev/null`, then an OSC 52 read query. On the stock `ask` default the working path raises the confirmation sheet first, the buggy path returns `UNAVAILABLE` before any sheet appears. With `clipboard-read = allow` no response comes back at all. Fix is to serve a zero-length `text/plain` representation whenever text was requested, even when `pasteboardText()` is nil.

**two smaller ones, not blocking.**

`showsOnScreen` is `deckOnScreen && window != nil`, and `window` does not go nil when a window is miniaturized, ordered out, or the app is hidden. So each hidden window keeps the swap chains of its selected session's on-screen hosts, which is both panes of a split plus a floating overlay or HUD if one is up, rather than every session. Inactive sessions are already released. `observeWindowVisibilityChanges` at `GhosttySurfaceView.swift:395` already subscribes to miniaturize, deminiaturize, app hide and app unhide, so most of the hook exists, `orderOut` would need a separate signal.

`QuickTerminalController.hide()` only calls `orderOut`, which leaves `NSView.window` set, and `QuickTerminalPane` never touches `deckOnScreen`. So the quick terminal's surface keeps its swap chain while the panel is ordered out between summons, potentially for the rest of the app's life, on the surface that spends the most time hidden.

**docs, mine to fix, listed so you know they are seen.**

- `GhosttySurfaceView.swift:204` says `deckOnScreen` is "narrower than `deckVisible`". It is wider, every call site agrees: no site has `deckOnScreen` false with `deckVisible` true. The PR body has the same inversion.
- `ClipboardPromptController.swift:4` and the alert text at :62 still say OSC 52. A Kitty program now raises that sheet and it names a protocol it is not using.
- `.claude/rules/libghostty.md:289` still says denied reads must complete with empty text and `confirmed = true`. Your change correctly replaced that with `deny_clipboard_request`, but an agent reading the rule as authoritative would revert it. Same paragraph's "identifies OSC 52 read versus paste" is now incomplete.
- `GhosttySurfaceViewTrackingTests.swift:15` still says "other two invisible contracts" after the third was added.
- `docs/backlog/hidden-panes-keep-drawing.md` was rewritten to `worth: no` instead of being deleted. `no` means a real item deliberately not worth fixing, so the backlog walk will keep reporting finished work as outstanding. The convention is `git rm` in the commit that fixes it.

the reduction you measured is consistent with what upstream reports, and the Zig 0.16 formula matches the pinned minimum. Happy to take another look once the three above are in.

paul-nameless:
All three are in (81bc119), plus most of the smaller ones.

1. New onScreen gate without the holdsKey term, wired through DeckPaneGates to deck panes, scratch, pane overlays, and the session overlay. Verified live: with the quick terminal summoned, the pane behind it keeps its surfaces resident.
2. writeClipboard keeps every representation and stores all of them with setData via a PasteboardType(mimeType:) mapping copied from Ghostty.app's shape. setData for text/plain too: the contract is binary-safe and a String round-trip would corrupt invalid UTF-8.
3. Empty pasteboard now answers a text read with a zero-length text/plain.

Also: the quick terminal clears deckOnScreen on hide, since orderOut leaves window set and its surface is the one that spends the most time hidden (verified: 97.8M resident drops to 62.4M on dismiss). The miniaturized/ordered-out window case is filed in docs/backlog/hidden-windows-keep-swap-chains.md. Of the doc items, I took the ones my diff introduced: the narrower/wider inversion (PR body too), the prompt text past OSC 52, the deny rule in libghostty.md, the test comment, and the backlog file is deleted instead of rewritten.

Running it as my daily terminal since this morning; footprint matches the patched branch it replaces (I had a forked version with patched libghostty before they fixed it)

paul-nameless:
Hold off on re-review: found a leak while daily-driving this build. After a day of real use the footprint grew from ~470 MB to 3.5 GB — hidden panes stop releasing after some re-hosting flow leaves the visibility flag stuck, and clicking through every session frees it all again. Running an instrumented build now to catch the exact path; will push a fix once I have it.

umputun:
reproduced the leak here. Split creation, split hide, and split close all retain IOSurface memory; dashboard and zoom return to baseline.

setup: Debug build at `683d8db`, isolated state dir, seven sessions with six hidden, each running a `date` loop so every surface paints. Numbers are `vmmap -summary` IOSurface resident, taken at least 3s after each action so the grace period has expired.

```
6 hidden sessions, at rest            18.1M   <- correct, hidden panes release
dashboard open (6 cells) -> closed    21.5M -> 18.1M
zoom show/hide x6, three cycles       18.1M   flat
split ON x6                           37.4M   <- hidden splits retain 19.3M
split OFF x6                          73.3M   <- hide retains another 35.9M
select every session                  37.4M   <- only the mounted panes release
remount both panes, select away       18.1M   <- full recovery
churn round 2                        100.9M
churn round 3                        150.1M
```

the growth is monotonic and settling does not recover it. That matches the shape of your 470 MB to 3.5 GB day-long run.

the measurements show retained resources on live surfaces that a real visibility transition can release. From 150.1M, remounting both panes with `split on`, then selecting every session and selecting away, returns exactly to the 18.1M baseline. That proves the missing true-to-false edge matters; it does not prove that `deckOnScreen` or `rendererVisible` is stuck at true.

the static path to instrument is Core Animation display. Metal's `IOSurfaceLayer` callback calls `renderer.drawFrame(true)` directly, bypassing `Thread.flags.visible`. After `false` releases the swap chain, attach, reparent, or bounds work can trigger a display and rebuild it while upstream `Surface.visible` and `rendererVisible` still read false. Later `false` writes then dedupe at both layers.

split hide rehosts the focused split pane as the sole pane and unmounts the primary. Selection cycles only the mounted split pane, which explains the partial recovery until both panes are remounted. `closeSplit` destroys the right pane and rehosts the surviving primary out of the `HSplitView` into the single-pane `deckPane`; the same display path can account for close adding one full primary per hidden session. The reparent is confirmed from the view structure. The callback timing is still a hypothesis for your instrumented build to check.

we ruled out an inherited `dropGhosttyLayer` leak before sending this. Removing its `contents` copy did not change the PR measurement, and master returned to its own baseline in three open-close rounds. That path explains none of this retention.

the pin itself checks out as upstream's merge commit for ghostty#14017, on `main` and a strict descendant of the old pin. The font-size-increase scrollback case also passes at `683d8db`.

one non-blocking wording follow-up in this PR: the OSC 52 generalisation stopped short. `ClipboardAccess` and `ClipboardPromptPolicy` in agtermCore, plus `ARCHITECTURE.md:15`, still say OSC 52 even though `confirmReadClipboard` now routes `KITTY_READ` and `KITTY_WRITE` through that policy. Please generalize those descriptions when you send the visibility fix. Wording only, no separate merge condition.

paul-nameless:
Pushed the fix for the bug I mentioned: after a big session restore the app sat at ~2.2 GB until you clicked through sessions.

Cause: we set occlusion off ~1s after launch, before the restore paints anything. The renderer's release is edge triggered and draws aren't gated on visibility, so those late presents rebuild every hidden pane's swap chain and nothing ever releases it again. Window resizes do the same during the day. Both reproduce on ghostty main, I'll file it upstream.

Fix: each hidden pane sweeps every 30s. Non-nil layer.contents means it was painted while hidden, so the sweep toggles occlusion to force a re-release and drops the retained frame. A settled pane costs one nil check. Measured: 2.29 GB at launch, 710 MB at +34s, flat after.

A single sweeper walking all panes would be tidier than a task per pane, but there's no global pane list (surfaces live under windows, splits, scratch, quick terminal, overlays), and the worst offenders are unmounted views a window walk can't reach, so it would need a hide/reveal registry. Can do that if you prefer the shape.

paul-nameless:
Our comments crossed: the fix I pushed at 7d95c23 covers your split churn too, same root cause (CA display rebuilding the chain after the edge, later false writes deduping). Confirmed against your exact setup, vmmap IOSurface resident:

```
baseline (6 hidden)   42.1M
split ON x6          127.7M
split OFF x6         211.4M
churn round 2        337.6M
wait 35s, no input    42.2M   <- first sweep, back to baseline
wait 95s              42.2M   flat
```

The sweep runs per hidden pane and survives unmount, so the rehosted split cases recover without any selection. Also pushed the wording fix: ClipboardAccess, ClipboardPromptPolicy and ARCHITECTURE.md now say OSC 52 or Kitty.

umputun:
the memory result is convincing and the root cause reads right to me: edge-triggered release plus draws that aren't gated on visibility, so late presents rebuild a chain nothing re-releases. Your launch numbers and the run against my churn both hold up. One thing blocks the janitor as written, and it is not the memory behavior.

**the re-arm toggle tells a hidden program it became visible.** `Surface.occlusionCallback` (Surface.zig:3309 at `683d8db`) sets `io.terminal.flags.visible` and, when mode 2033 is on, queues a `visibility_report` for every edge. The only guard in front is a same-value dedupe. So

```swift
ghostty_surface_set_occlusion(surface, true)
ghostty_surface_set_occlusion(surface, false)
```

emits visible then hidden to a program that stayed hidden throughout, on every sweep that finds a late present. That happens at least once for the leaked panes this janitor is meant to repair, and again whenever a hidden present repopulates `contents`. A TUI that dims or pauses while hidden can resume and pause again while the user is looking elsewhere. The public occlusion API cannot be used as a renderer-only release edge: it is a terminal-visibility API that also releases GPU resources.

the upstream report should state the invariant rather than prescribe a new C API: an invisible surface must not rebuild its swap chain through the Core Animation display callback, and releasing Metal resources must drop the layer's retained target. An internal upstream fix may remove the janitor entirely. If upstream instead exposes a renderer-only force-release operation, the janitor can use it, but agterm cannot implement either route through the current ABI.

**second, smaller: nilling `contents` has no freshness barrier before reveal.** the comment says the reveal path refreshes before anything shows, but `refreshCallback` (Surface.zig:3426) only calls `queueRender`, whose own comment says a render isn't guaranteed to happen immediately. The ordinary draw presents with `sync: false`, and `IOSurfaceLayer.setSurface` dispatches the contents assignment to the main queue. A session switch doesn't change the layer bounds, so nothing forces a synchronous Core Animation display either. So SwiftUI can commit the reveal while `contents` is still nil, and that retained frame is exactly what used to cover the rebuild. I have not reproduced a flash, so I would not call it frequent. The ordering hole is still real: there is no barrier between clearing `contents` and exposing the view, and a one-frame blank on session switch is a reachable result.

**tests.** `81bc119..a2683e2` adds none. The 30s cadence isn't worth testing, but the wiring is and it is all reachable in the hosted target: the janitor starts on the hide edge, does no work after reveal and retires on its next wake, and `destroySurface` cancels it and clears the task. Worth having whatever the mechanism ends up being, since the lifecycle is the part that will get broken by a later edit.

**logging.** `visibilityLogger.notice` with a public session UUID fires on every logical visibility edge for every surface, so a restore of 63 surfaces writes 63 default-visible lines before anyone has done anything. `.debug` fits: the UUID is only useful when you have that category turned on anyway.

**on your question, keep the task per pane for now.** the task lives beside the renderer state it checks and needs no separate registry; Session ownership keeps an unmounted pane available, and the task's long sleep captures it weakly. A global sweep could traverse the stores and reach detached primaries, so that is not the distinction. Its real benefit would be one timer instead of a group of 30s wakeups. Unless those wakeups measure as a problem, the per-pane version has less shared machinery and fewer stale-entry paths.

the reproduction and diagnosis are solid. Keep the per-pane ownership shape if upstream provides a renderer-only re-arm; if upstream makes invisible draws unable to rebuild resources, the janitor should disappear instead.

paul-nameless:
Pushed two commits addressing the review.

b3d6373 removes the toggle: you're right that occlusion is terminal visibility, so the sweep now only drops the layer's retained frame. Reveal draws synchronously via ghostty_surface_draw (the same main-thread drawFrame the CA display callback uses) closing the blank-frame window. Janitor lifecycle has hosted tests now, and the edge log went to debug.

e15f9c1 replaces what the toggle was doing at launch, legally this time: the hide edge now defers until the pane's first present (60s cap). The edge lands after the chain exists, so the one release a pane already gets actually frees it. No extra visibility edges, a pane still spawns visible and hides exactly once, just a couple of seconds later.

Measured: sandbox restore (13 hidden) spikes to 224M IOSurface and settles at the 42M baseline in ~30s untouched; my 45-session daily instance now launches to ~770MB where the toggle-less build sat at 2.2G. What remains for upstream is the invariant you stated: resize can still rebuild a hidden pane's chain through the CA display path, and that heals only on a real visit.

paul-nameless:
Turns out the resize rebuild was fixable app-side: Ghostty's layer sets needsDisplayOnBoundsChange, so every resize made CA display hidden panes and rebuild their freed chains. b5f4312 drops that flag while a pane is renderer-hidden and restores it on reveal; a22dec6 makes reveal always draw synchronously so a stale-size frame can't flash. Six scripted resizes over a settled sandbox leave IOSurface flat. Launch, retained frames and resizes are all handled in the PR now.

umputun:
this answers both, and the resize fix is a better outcome than the upstream wait I was expecting. The upstream paths at `683d8db` support both changes.

`ghostty_surface_draw` does draw inline: `Surface.draw` (Surface.zig:894) calls `renderer.drawFrame(true)` directly with the comment "Renderers are required to support `drawFrame` being called from the main thread, so that they can update contents during resize", so this is the sanctioned path rather than a liberty. It holds `draw_mutex`, so it serializes against the render thread's own draws, waits for GPU completion and writes `layer.contents` through `setSurfaceSync` before returning. The blank window really is closed, and one synchronous reveal draw is the API's own contract.

the `needsDisplayOnBoundsChange` drop is coherent too: ghostty sets it once in `Metal.init` and never reads it again, so nothing upstream desynchronizes from the flag. You keep sending size and content-scale updates while hidden, and the reveal draw reads current bounds, so a visible pane's live-resize behavior is unchanged and only the implicit hidden-pane display callback goes away. I checked whether anything upstream reinstalls that layer mid-life, since the flag is written once per edge and never re-applied: nothing does at this pin, one layer per surface lifetime.

**one thing before merge, and it is small.** the two tests you added are the right shape, but they cover the janitor, which is now the simplest part. What has no coverage is the part that changed four times and carries the correctness: first present exits the poll and then hides, expiry hides a pane that never presented, a reveal or cancellation prevents the deferred hide, `delayHide: false` skips the grace but still waits for first present, and hide drops `needsDisplayOnBoundsChange` with reveal restoring it.

that reads like more than it is, because one seam unlocks most of it. `hiddenJanitorInterval` is currently a `static var` whose only writer anywhere is the test that shortens it, which is shipping mutable state that exists to be reassigned. If the cadence goes back to a `static let` and both the janitor and the first-present wait take their intervals as parameters defaulting to it, the tests can pass small values and none of them has to wait 60s. Same change fixes the mutable global and makes the deferral testable.

the bounds-flag assertion is nearly free once you are in there: hide, assert false, reveal, assert true.

two smaller notes, both mine rather than asks. `GhosttySurfaceView+Visibility.swift:67` says the synchronous draw makes the first shown frame current. `drawFrame(true)` doesn't call `updateFrame`, and the renderer deliberately skips it while invisible, so what that draw guarantees is a nonblank frame at the current bounds; terminal output produced while the pane was hidden arrives when the queued visible render runs. The runtime behavior is correct; the sentence claims one step more than it delivers. And I will add the pin-coupling to `libghostty.md`: the flag mutation is tied to Metal.zig still using that property, so a future `GHOSTTY_REV` bump has to recheck it. Your rule about never toggling occlusion to force a re-release is exactly the right thing to have written down, and that is where it belongs.

everything else is settled and I am not looking for more. Send the coverage and this goes in.

paul-nameless:
Coverage pushed in 03a1d92. The cadence statics are lets again and the deferral/janitor take intervals as parameters, so the tests pass small values. Covered: first present ends the wait and hides, expiry hides a never-painted pane, reveal cancels the deferred hide and restores needsDisplayOnBoundsChange, an already-painted pane hides without deferral, plus the existing janitor pair. The bounds-flag write moved ahead of the surface guard to make that wiring testable host-free. Reveal-draw comment reworded per your note.

PR #495, opened 2026-08-28
feat: report the pane session.restore wrote

ssgreg:
Echoes the pane `session.restore` wrote, so a caller that addressed it by token can read back where the pin landed.

```
$ agtermctl session restore "claude --resume abc" --pane-id $AGTERM_PANE_ID
ok
$ ... --json
{"ok":true,"result":{"id":"EB8AE210-992C-44E9-81AF-CB575D4B0241","pane":"right"}}
$ agtermctl session restore "npm run dev" --json          # no selector, the main-pane default
{"ok":true,"result":{"pane":"left","id":"EB8AE210-992C-44E9-81AF-CB575D4B0241"}}
$ agtermctl session restore --clear --pane right --json
{"ok":true,"result":{"id":"EB8AE210-992C-44E9-81AF-CB575D4B0241","pane":"right"}}
```

`--pane-id` names a surface by its spawn token rather than by a role, which is the point of it: a hook in a promoted-then-re-split pane pins the right slot without knowing which slot that is. The cost was that the caller could not find out either, short of re-reading `tree` and diffing `restoreCommand` against `splitRestoreCommand`. `result.ratio` already does this for `session.resize`, so the shape is not new.

**On every success, not only for `--pane-id`.** The `--pane` and default-to-main paths report it too, so a caller never has to branch on how it addressed the pane to know how to read the answer. It costs one field and removes the only reason to special-case the token form.

**The rule this exposed, recorded in the contract.** `ControlArgs` is synthesized `Codable` with no `CodingKeys`, so an app that predates a field drops it and runs the command without it, answering an ordinary ok. That is unlike an unknown *command*, which fails to decode and names itself, and the asymmetry was not written down anywhere. `.claude/rules/control-api.md` now says it, and says that a field deciding *where* a mutation lands wants a read-back rather than leaving the two outcomes indistinguishable. It also notes the exposure is narrow, since `agtermctl` ships inside the bundle: the CLI that sends a field and the app that reads it are the same build, so only a stale running process across an upgrade can split them.

Surfaces synced: `site/commands.html`, the bundled skill reference, and the `--pane-id` paragraph in `control-api.md`, which said this addressing "adds no read-back field" and is now true only of `session.status`.

`make test` 2632, `make test-app` 259, `make lint`, `make build`, all green. The transcript above is a scratch instance with its own state dir and socket. `ControlServerSessionActionsTests/testRestoreReportsThePaneItActuallyWrote` covers the split, the explicit-pane and the default paths; with `pane:` dropped from the result it goes red on both panes rather than only the interesting one:

```
ControlServerSessionActionsTests.swift:669: error: XCTAssertEqual failed: ("nil") is not equal to ("Optional("right")")
ControlServerSessionActionsTests.swift:677: error: XCTAssertEqual failed: ("nil") is not equal to ("Optional("left")")
```

No UI test: this path has none.

### DIFF
--- a/.claude/rules/control-api.md
+++ b/.claude/rules/control-api.md
@@ -117,6 +117,13 @@ paths:
   never the error:
   `DecodingError.debugDescription` is macOS 26.4+, so at the 14.0 deployment target `String(describing:)`
   degrades to a reflection dump that hides the sentence inside `DecodingError.Context(...)`.
+- An unknown ARGUMENT is not symmetric with an unknown command. `ControlArgs` is synthesized `Codable` with
+  no `CodingKeys`, so a server that predates a field drops it and runs the command without it, answering ok:
+  there is no "unknown field" error and no way to ask. A new field that only narrows or decorates is fine
+  that way. One that changes WHERE a mutation lands is not: give it a read-back so a caller can see what the
+  server did, rather than leaving the two outcomes indistinguishable. Since `agtermctl` ships inside the
+  bundle, the CLI that sends a field and the app that reads it are the same build, so the exposure is a
+  stale RUNNING process across an upgrade, not a mismatched install.
 - Human output shows IDs only for created session/workspace/window, retains them in JSON, uses
   `result.affected` for session counts, and reserves `result.count` for diagnostics/search.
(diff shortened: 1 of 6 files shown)


umputun (approved):
lgtm

PR #496, opened 2026-08-28
Fix peer chat for labelled Claude composers

paskal:
Claude Code can place a context label inside the top composer rule, for example `──── e2e ─`. The peer-chat parser required that rule to contain line characters only, so it could type a message but could not verify the composed text and deliberately withheld Return.

This change accepts the optional context label while retaining the closing-rule check that distinguishes the composer from ordinary terminal output.

A standard-library regression script covers plain rules, labelled rules, and a prompt missing its closing rule. The cookbook CI job now executes every `test_*.py` recipe check in its own process, alongside the existing Ruff gate.

Checks completed:

- `python3 cookbook/two-agent-chat/test_peer_chat.py`
- `ruff check --ignore EXE001 cookbook/two-agent-chat/peer-chat.py cookbook/two-agent-chat/test_peer_chat.py`
- `actionlint .github/workflows/ci.yml`
- `make test`
- live Claude send through the corrected parser: `{"sent": 354}`

### DIFF
--- a/.claude/rules/ci.md
+++ b/.claude/rules/ci.md
@@ -50,9 +50,10 @@ paths:
 - The cookbook job builds nothing. It compares the `cookbook/README.md` table and recipe directories in
   both directions; requires kebab-case directories, a `README.md` with all six exact
   (`grep -qxF`) headings, and shebangs for `.sh`/`.zsh`/`.py`; runs `shellcheck` on `.sh`; parses `.zsh`
-  with `zsh -n`; and runs `ruff check` on `.py`. `shellcheck` is preinstalled. Install absent `zsh` and
-  `ruff` in separate steps immediately before their own; shellcheck cannot lint zsh, and `ruff` needs
-  `pipx` because the runner's python is externally managed.
+  with `zsh -n`; runs `ruff check` on `.py`; and executes every `test_*.py` regression script directly.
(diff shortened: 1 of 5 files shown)


umputun (approved):
parser fix is good and the regression script pins it. Two nonblocking notes on the new CI gate, both in the sentence you added at `cookbook/CONTRIBUTING.md:17`.

`cookbook/CONTRIBUTING.md:17` - CI runs the file as `python3 <file>`, which executes its top-level statements and stops there. Defining test functions is not running them, so a `test_*.py` written pytest-style, or without the `if __name__ == "__main__": unittest.main()` guard, exits 0 having asserted nothing and the step stays green. `test_peer_chat.py` gets this right, but nothing written asks the next one to. Worth stating the contract: a `test_*.py` is a standalone entry point, so running it executes its checks and exits non-zero on failure.

`cookbook/CONTRIBUTING.md:17` - the cookbook job runs on `ubuntu-latest` with stock runner Python, no install step for a recipe's own dependencies, no `agtermctl` and no live agterm to talk to. That same line already lets a recipe reach past the standard library when it genuinely must, so a test for one of those can pass locally and red on the runner with the constraint written nowhere. The doc could say a `test_*.py` has to supply its own fakes for whatever the runner does not have.

Issue #502, opened 2026-08-28
two-agent-chat: sends to Claude Code are never verified, message stranded in the composer


mazanur:
The `two-agent-chat` recipe's `peer-chat.py` does not complete a send to Claude Code as shipped. Every send is typed into the pane and then refused at the verification step with `message was typed but not verified in the target composer; submit withheld`, leaving the message stranded in the composer.

Tested with agterm 0.24.0 and a current Claude Code. I patched a local copy until sends went through; the three causes are below. `#d6a5c06` (labelled composer rules) is a different problem and does not address these.

### 1. The live composer is not found

`claude_prompt_text` requires the rule to sit on the line immediately above the `❯` prompt:

```python
if not match or not RULE_RE.match(lines[index - 1]):
    continue
```

Current Claude Code does not draw it that way. It frames the live composer with a rule **both above and below** the prompt, and the line directly above the prompt is not always the rule. So the live composer is never identified, and `prompt_text` returns nothing for a composer that is plainly on screen.

The lower border turned out to be the useful signal: it is what separates the live composer from an already-submitted prompt higher up the transcript. I ended up scanning a small window above for a rule and requiring a rule below.

### 2. Long messages become a paste placeholder

Post-typing verification asserts the composer content starts with the label:

```python
return content.startswith(profile.label)
```

Claude Code collapses a long pasted-in message into a placeholder, so the composer reads `[Pasted text #1 +12 lines]` and never starts with `Chat from Codex:`. `PASTED_RE` only covers `[Pasted Content N chars]`, not the `[Pasted text #N +N lines]` form. Verification fails on exactly the messages the recipe is for.

### 3. The message is accepted before verification looks

On longer messages Claude Code can move the composer content into the transcript before `wait_for_composed` reads the pane. The composer is empty by then, so the script reports "typed but not verified" and withholds submit — but the message was already accepted. Sending the submit key at that point would submit a second time.

What worked: count matching prompts in the transcript before typing, and after typing treat "composer empty + caret at rest + a new matching transcript prompt" as delivered, returning without injecting a submit key.

### Also

`BOX_LINES = 40` is not enough pane text to see both composer borders in current Claude Code, and `PROBE_TIMEOUT = 0.6` is tight — a slow `agtermctl` probe reads as an empty composer. I run 240 and 3.0.

Happy to send a PR if the approach looks right. The Codex direction was fine throughout; all of this is Claude-side only.

umputun:
tickets like this one won't be accepted.

[...]

PR #505, opened 2026-08-29
Harden peer chat composer detection and message input

paskal:
Codex draws its composer prompt as `›`, but the glyph changes to `»` at ultra reasoning effort. The peer-chat parser matched only `›`, so it could type a long message into an ultra session, fail to recognise the compacted composer content and withhold submission.

This change:

- recognises both Codex prompt glyphs and configurable two-column-indented status lines;
- treats a newer column-zero `!` shell prompt, a multi-row shortcut overlay or an unknown trailing UI row as a pre-write refusal;
- reads pane text before the caret check, leaving the caret as the final pre-write sample;
- keeps master's Claude post-write parser byte-identical while applying stricter dialog refusal through a separate live-prompt parser;
- preserves master's 20-character Claude body-fragment verification while typing the routing label separately;
- accepts both agents' compacted-paste forms and allows two seconds for a long body to render.

The branch also adds private file-backed input for Codex callers. A heredoc is evaluated as a shell wrapper and cannot match a `peer-chat.py` execution rule, while a reserved basename keeps both invocations matchable without exposing message text in process arguments. Prepared files use a mode-0700 per-user spool and mode-0600 one-shot entries. Target and session resolution completes before the file is opened, so an unrelated target refusal preserves it for retry. Once identity, ownership and link checks pass, the entry is removed before mode, size and body validation. Setup includes two narrow Codex execution rules for reserving and sending these files.

Known limitations:

- Post-type verification confirms a 40-character Codex prefix or a 20-character Claude body fragment, so it can miss a dropped middle segment in a long message. Chunked typing with per-chunk verification is outside this patch.
- A multi-row Codex shortcut overlay is treated as busy. Stripping every indented row would also strip modal choices, so the send fails closed until the overlay closes.

Checks:

- `python3 cookbook/two-agent-chat/test_peer_chat.py` (44 tests)
- `python3 cookbook/codex-session-resume/test_codex_resume.py` (7 tests)
- `find cookbook -type f -name '*.py' -print0 | xargs -0 ruff check --ignore EXE001`
- Reinstating the shared Claude parser fails both live-prompt refusal tests; moving `read_message` ahead of target resolution fails `test_target_refusal_preserves_prepared_message`.
- `codex execpolicy check` accepts both documented file-backed command prefixes.

### DIFF
--- a/cookbook/two-agent-chat/README.md
+++ b/cookbook/two-agent-chat/README.md
@@ -24,7 +24,14 @@ agterm 0.24.0 or later, which added `surface cursor`. The recipe refuses to type
 2. Copy `SKILL-claude.md` to `~/.claude/skills/peer-chat/SKILL.md` and `SKILL-codex.md` to `~/.codex/skills/peer-chat/SKILL.md`. Both loaders require the installed file to be named exactly `SKILL.md`, so the suffix here only says which agent the file is for. Each one tells its own agent how to send, how to recognise an incoming message, and what it may not do to the other pane.
 3. Edit both copies so the path they name for `peer-chat.py` matches where you put it.
 4. If you start either agent through a wrapper script rather than as `claude` or `codex`, put that wrapper's name in the same file, as the `--target-command` value the agent should pass when sending to it. Without this the first send refuses, saying the target pane is not running the expected command.
-5. Start Codex so it can tell which pane it is in. Codex strips `AGTERM_SESSION_ID` from tool subprocesses, so start it with `codex -c "shell_environment_policy.set.AGTERM_SESSION_ID=\"$AGTERM_SESSION_ID\""` to copy the current pane's session id into every tool command, adding your normal Codex options to the same command. Without this injection the script can still resolve one matching repository checkout, but it refuses when several sessions share that checkout. If you launch Codex through a wrapper, put the flag in the wrapper and it applies to every pane.
+5. To let Codex reserve and send file-backed messages without separate approvals, add these two entries to `~/.codex/rules/default.rules`, creating the file if needed and using the command name or path from step 1:
+
+   ```python
+   prefix_rule(pattern=["peer-chat.py", "--prepare-message"], decision="allow")
+   prefix_rule(pattern=["peer-chat.py", "--to", "claude", "--message-file"], decision="allow")
+   ```
+
+6. Start Codex so it can tell which pane it is in. Codex strips `AGTERM_SESSION_ID` from tool subprocesses, so start it with `codex -c "shell_environment_policy.set.AGTERM_SESSION_ID=\"$AGTERM_SESSION_ID\""` to copy the current pane's session id into every tool command, adding your normal Codex options to the same command. Without this injection the script can still resolve one matching repository checkout, but it refuses when several sessions share that checkout. If you launch Codex through a wrapper, put the flag in the wrapper and it applies to every pane.
 
 If your `agtermctl` is not on `PATH` under that name, set `AGTERMCTL` to its full path; the script reads that variable and falls back to `agtermctl`.
 
@@ -42,6 +49,20 @@ MSG
 
 `--to claude` sends the other way. `--session` names a session explicitly; without it the script uses `AGTERM_SESSION_ID` when the caller has one, and otherwise looks for a single session whose target pane is running the expected agent.
 
+Codex can keep the send as plain command arguments by using a private one-shot message file. Reserve a fresh name first:
+
+```sh
+peer-chat.py --prepare-message peer-chat-codex-a91f.txt
+```
+
+The command prints the absolute `messageFile` path. Write the UTF-8 message to that exact file without replacing its mode, then send it by name:
+
+```sh
+peer-chat.py --to claude --message-file peer-chat-codex-a91f.txt
+```
+
+This form matches the two Codex execution rules in *Setup*. A heredoc uses shell redirection, so Codex evaluates the whole request as a shell wrapper and cannot match the inner command.
+
 Before every send the script confirms that the pane it is about to type into really is running the agent named by `--to`, reading the command from agterm's own view of the pane. It looks for `claude` and `codex`. If you start an agent through a wrapper script, that name is what agterm sees instead, and the send is refused until you say so:
 
 ```sh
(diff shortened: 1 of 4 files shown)


umputun:
master moved under you and this conflicts now. I merged #506 into the same two files after you opened this, without checking for an open PR first, so the conflict is mine rather than yours. Rebase when you get a chance and the two points below can go in the same push.

the `»` glyph is a real find. The 2.0 timeout is right too, someone hit the same render latency in #502, and my reason for keeping 0.6 was about probe cost and never covered how long a long body takes to draw.

**keep master's Claude verifier.** #506 replaced `content.startswith(label)` with a 20-character body fragment check that takes either a label-leading row or a `[Pasted text #N]` summary that retains a fragment, and refuses a summary that retains none. Your version accepts a bare marker, which cannot tell our message from somebody else's paste already sitting in the box. Typing the label as its own write is worth keeping on its own, but it does not make the bare-marker fallback safe.

**the pre-typing guard passes when it should refuse.** `codex_prompt_text` scans backward over 40 lines, so an older submitted `›` above a live `!` shell prompt returns the old one, the guard passes and the message is typed into shell mode, which is the case the guard exists for. This screen returns `earlier prompt`:

```
› earlier prompt
  some response output
! ls -la
  shell mode active
```

it needs the parser to stop at the newer shell prompt instead of falling back, and a test with transcript history above the live prompt. The pane read should also happen before the cursor check, leaving the caret sample as the last pre-write check instead of widening the gap after it. That narrows the window rather than closing it, the target check inside `type_text` still leaves one.

paskal:
Rebased, please recheck.

umputun:
rebase is clean and all three points are in. `composer_has_message` is byte-identical to master's, the `»` glyph and the 2.0 timeout are both right, and `test_prompt_is_checked_before_cursor` asserting `cursor_column.assert_not_called()` is a nicer pin on the read order than I would have written.

one blocking thing, then some notes that don't block.

**blocking: the pre-write refusal narrows post-write verification for Claude.**

for Codex you split the two jobs into two parsers, `codex_prompt_text` for the post-write check and `codex_live_prompt_text` for the pre-write refusal. For Claude both `prompt_text` and `live_prompt_text` call the same `claude_prompt_text`, so the new `return None` arm at the end of it tightens both. That undoes half of #506, which removed the requirement for a rule below the chevron:

```
wrapped composer, no closing rule
  master : 'Chat from Codex: a wrapped message with a continuation'
  #505   : None
```

`prompt_text` is what `wait_for_composed` reads, so `None` there is "message was typed but not verified in the target composer" with the text stranded in the box, which is the exact failure #506 fixed. `test_rule_after_prompt_is_not_required` still passes because its screen is `{RULE} e2e ─\n❯ Chat from Codex: ping`, where the chevron is the last line and takes the `index == len(lines) - 1` arm. Nothing covers the wrapped shape, which is what a real peer-chat message draws.

give Claude its own `claude_live_prompt_text` the way Codex has one, and leave `claude_prompt_text` as master has it.

**the rest, none of it blocking.**

`peer-chat.py:550` - `read_message()` consumes the spool file before `target_profile()` and `resolve_session()` run, so a refusal that has nothing to do with the message destroys it. `SKILL-codex.md:45-49` tells the agent to retry with `--target-command <name>`, and that retry re-uses the same reserved name whose entry is gone, so it dies on ENOENT instead of the refusal it was told to fix and the body is lost. Moving `message = read_message(...)` below `sid = resolve_session(...)` fixes it. Nothing the script drives reads stdin, so the move is safe.

`test_peer_chat.py:390` - `spool.mkdir(mode=0o755)` is umask-filtered, so under umask 077 the dir comes out 0700, `private_spool_fd()` never raises and the test fails with "ValueError not raised". Green at 022 and 027, red at 077. `spool.mkdir()` then `spool.chmod(0o755)` works, chmod isn't masked.

`peer-chat.py:244` - the footer strip pops one row but the comment above `CODEX_FOOTER_RE` says every footer row is indented, and Codex's `?` shortcut overlay is genuinely multi-row (`codex-rs/tui/src/bottom_pane/footer.rs`, `FooterMode::ShortcutOverlay` returns N lines plus a blank and a "customize shortcuts with /keymap" row). With that overlay up `codex_live_prompt_text` refuses and `send_with_retry` burns 5 attempts over ~40s. I'd not make the strip greedy though: an indented modal row is shape-identical to a footer row, so stripping the whole region would let a `  1. Allow` / `  2. Deny` chooser under a stale `›` pass the live check, which is the case you're guarding. Keeping the single pop and saying so in the comment plus README *Limits* seems better than weakening the modal guard.

`SKILL-codex.md:27-29` - README:51 says to write the file "without replacing its mode" and the skill doesn't repeat it, so the file the writing agent reads is missing the constraint the script enforces. Worth carrying the clause across, especially since the error text names `--prepare-message again` as the remedy and re-preparing re-creates at 0600 for the same writer to reset again.

`README.md:111` - *Limits* says the one-shot file "is removed when a send reads it", but it's also removed when the send rejects it on mode or size without reading, and when a later step aborts the send. Same edit as the `main()` one above.

`SKILL-codex.md:37` - it refers to "the global `peer-chat.py` allow rule" as something that exists, but no *Setup* step tells the reader to create one, and README:57 only says the form "can match" one.

on the file-backed input: no objection to it being here. The problem is real, the security model reads sound to me (dir_fd-relative opens with `O_NOFOLLOW`, the fstat/lstat dev+ino compare, `nlink == 1`, `O_EXCL` at 0600 inside a 0700 spool), and it's narrower than `--stdin` rather than wider. It's your recipe and it's the path you actually run.

paskal:
Addressed in `2f43fd2`. `claude_prompt_text` now matches master byte for byte, while `claude_live_prompt_text` carries the pre-write refusal; the new tests cover the wrapped post-write shape without a closing rule, both live refusal shapes and target-refusal file preservation. `read_message` now runs after target and session resolution, the public-spool test uses an explicit `chmod`, and the single-row overlay limit plus exact consumption semantics are documented. The Codex skill preserves the prepared file and its mode, and Setup supplies the two minimal execution rules. The branch remains one signed commit; all 44 peer-chat tests, all 7 resume tests and the cookbook ruff check pass. Please take another look.

umputun:
all three points from the last round check out.

`claude_prompt_text` is byte-identical to master and the pre-write refusal sits in its own `claude_live_prompt_text`. `read_message` runs after `resolve_session`, with `test_target_refusal_preserves_prepared_message` pinning that a target refusal leaves the file readable. The public-spool test is green at umask 077, 027 and 022 now the `chmod` is explicit, Setup carries the two execution rules, and the skill repeats the mode constraint.

three notes below, none of them blocking.

**`peer-chat.py:341` - typing the label separately narrows what the post-write check accepts.**

when the label write reaches Claude as its own event, the composer's first row starts with the literal label, so `composer_has_message` takes the `shown.startswith(label)` arm and the search at `:344-348` never runs. PTY reads can coalesce the two writes, so this is not every send. Claude Code 2.1.251 replaces an input event over 800 characters with `[Pasted text #N]` and inserts anything shorter literally, which leaves two shapes:

- body arrives as one event over 800: `Chat from Codex: [Pasted text #7]`. Master's bare chip hits the same post-write refusal, so the outcome is unchanged even though the row reads differently.
- body splits across a big event and a smaller one: `Chat from Codex: [Pasted text #7 +12 lines] tail fragment`. Master took the search arm and accepted this, we refuse it.

so the regression is the second shape only, and it turns on how the PTY read chunks. Making the label arm a fallback recovers it:

```python
if shown.startswith(label) and body[:MESSAGE_FRAGMENT] in shown:
    return True
```

control then reaches the search you already have. Every existing `ComposerHasMessageTests` case keeps its result and the new label-plus-chip-plus-tail case turns true, `test_label_leading_bare_paste_marker_is_refused` still refuses.

if each `session type` write arrives as one event, the split can also fix a case master gets wrong. The label is 17 characters, so an ASCII body of 784 to 800 crosses the threshold in master's combined write and becomes a bare chip master then refuses. Written on its own it stays literal and verifies.

**`peer-chat.py:288` - the live parser returns at the first rule and never looks below it.**

`claude_live_prompt_text` returns as soon as it finds a `RULE_RE` under the newest `❯`. A historical prompt can retain that closing rule, and then this screen comes back as a live composer:

```
❯ previous user prompt
────────────────────
Allow this command?
  1. Yes
  2. No
```

`test_live_stale_prompt_under_a_dialog_is_refused` passes because its screen has no rule between the prompt and the dialog. Add one and it returns `previous user prompt`. I could not establish that a real chooser leaves the caret at column 2, so this is a parser and test gap rather than a demonstrated send into a dialog. `trailing_input_block` on the codex side is the precedent: it removes one known footer row, then accepts the prompt only as the block's final row.

**README.md:31 - the second execution rule allows any tail.**

`prefix_rule` pins the prefix only, and `codex execpolicy check` against those two rules returns allow for all of these:

```sh
peer-chat.py --to claude --message-file peer-chat-codex-a91f.txt --session XYZ
peer-chat.py --to claude --message-file peer-chat-codex-a91f.txt --target-command zsh
peer-chat.py --to claude --message-file peer-chat-codex-a91f.txt --to codex
```

`parse_args` takes all three, and `--to` is last-wins, so the trailing `--to codex` beats the `--to claude` the rule matched on. `SKILL-codex.md:47` tells the agent never to pass a `--session` id it inferred, which is the right instruction, but an allow rule only suppresses the approval prompt and cannot enforce it. I have not found a rule form that anchors the tail, so the smaller ask is to say in Setup that the second rule permits any trailing arguments. I would not make the script refuse `--session` or `--target-command` alongside `--message-file`. That kills the wrapper path in Setup step 4, for exactly the callers this mode serves.

merging this round.

PR #514, opened 2026-08-31
Recover Claude tabs from bridge-only transcripts

paskal:
Claude Code can leave a transcript containing only a `bridge-session` record after `/resume`. The existing recipe treated any matching JSONL file as resumable, so a restored tab selected `--resume`, received `No conversation found`, and could not reuse the occupied UUID.

This change leaves the metadata-only transcript intact and derives the tab's next UUID deterministically. It also validates `AGTERM_SESSION_ID`, finds transcripts without reproducing Claude's private project-path encoding, and defers malformed or unfamiliar records to Claude.

The bundled regression suite runs independently under zsh and fish in CI. Its 15 cases per shell cover the original collision, generated-ID replay, malformed input, configuration isolation, and project-key variation.

Verification:

- The complete cookbook gate passes in Ubuntu 22.04: shellcheck, zsh and fish parsing, ruff, and all Python recipe tests.
- The new suite fails against the previous wrapper at seven assertions per shell, including the bridge-only collision, generated-id replay and malformed tab-id handling.
- Weakening the bridge classifier makes the truncated-record regression fail at the expected assertion.
- Claude Code 2.1.251 starts the derived UUID and reaches its prompt while the original bridge-only file remains in place.

### DIFF
--- a/.claude/rules/ci.md
+++ b/.claude/rules/ci.md
@@ -51,9 +51,9 @@ paths:
   both directions; requires kebab-case directories, a `README.md` with all six exact
   (`grep -qxF`) headings, and shebangs for `.sh`/`.zsh`/`.py`; runs `shellcheck` on `.sh`; parses `.zsh`
   with `zsh -n`; runs `ruff check` on `.py`; and executes every `test_*.py` regression script directly.
-  `shellcheck` is preinstalled. Install absent `zsh` and `ruff` in separate steps immediately before
-  their own; shellcheck cannot lint zsh, and `ruff` needs `pipx` because the runner's python is externally
-  managed.
+  `shellcheck` is preinstalled. Install absent `zsh` and `fish` together before the shell gates (`fish`
+  runs the Claude resume regression), and install `ruff` through `pipx` because the runner's python is
+  externally managed.
 - Recipes are not shell-only. A language gains a gate by adding its extension to the shebang glob plus a
   lint or parse step; until then it merges unchecked, which is why `cookbook/CONTRIBUTING.md` tells a
   contributor to flag any other language in the pull request. Keep that file, `cookbook/README.md` and
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -250,9 +250,9 @@ jobs:
         # without it a failed find would leave the step green having checked nothing
         shell: bash
         run: find cookbook -type f -name '*.sh' -print0 | xargs -0 -r shellcheck
-      - name: Install zsh
-        # not on the ubuntu-latest image, unlike shellcheck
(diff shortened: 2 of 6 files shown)


umputun (approved):
lgtm. No changes requested for this PR, a few notes:

`cookbook/claude-session-resume/claude-resume.zsh:50` - `local line found=0 other=0` runs once per transcript examined, and zsh prints a bare `local <name>` when the parameter already exists in the scope. A tab that has recovered once has two transcripts to examine, so it echoes `line=''` before Claude starts on every later launch. Possible fix: `local line='' found=0 other=0`.

`cookbook/claude-session-resume/test_claude_resume.py:307` - `status = max(status, result.returncode)` yields 0 for a child killed by a signal, so the step can report green even though one per-shell child was killed before completing. Possible fix: `if result.returncode != 0: status = 1`.

`cookbook/claude-session-resume/test_claude_resume.py:253` - `test_restored_generated_resume_is_recomputed` passes against the pre-fix wrapper too. The old `"${2:l}" == $sid` check failed, fell through to the flag scan, and echoed `--resume <generated>` verbatim, which is what the test asserts. A mutant that narrows the replay check back to that exact-`sid` comparison passes 15/15 in both shells, so the same-prefix replay branch is currently pinned by nothing. Possible fix: replay `--resume <generated(1)>` with generation 1 also metadata-only and the conversation at generation 2, expecting `--resume <generated(2)>`, which the passthrough path cannot produce.

`cookbook/claude-session-resume/test_claude_resume.py:280` - nothing here is wrong. `test_parent_config_dir_does_not_escape_fixture` does pin the harness, dropping the `env.update` override at line 126 makes both shells exit 73. What no test covers is the wrapper's own `CLAUDE_CONFIG_DIR` handling at `claude-resume.zsh:33` and `claude-resume.fish:42-44`, since the child always sees the fixture value. Possible addition: thread an optional `config_dir` through `invocation`, point it at a second fixture root holding a real conversation for `TAB_ID` with `.claude/projects` left empty, and expect `--resume TAB_ID`. A wrapper that ignored the variable emits `--session-id` instead.

`cookbook/claude-session-resume/test_claude_resume.py:122` - the fish branch runs `fish -i -c` with `HOME` redirected but `XDG_CONFIG_HOME` inherited, so a contributor who exports it runs his own `config.fish` inside the harness. An explicit `set -gx PATH ... $PATH` there prepends ahead of the fixture bin, and `command claude` finds his real binary instead of the fake. The `zsh -f` on the other branch already prevents the equivalent. Possible fix: add `XDG_CONFIG_HOME` and `XDG_DATA_HOME` under the fixture to the `env.update` dict.

PR #527, opened 2026-09-02
fix(cookbook): make peer chat delivery reliable

paskal:
A single 1,941-byte input event could reach Claude Code with 1,022 contiguous bytes missing, while sending the body and Return together could leave the complete body composed but unsent. The peer-chat recipe could not distinguish either result from delivery.

This keeps the existing `agtermctl session type --stdin` transport and separates body delivery from submission. Long text is sent as UTF-8-safe marked events capped at 197 bytes; every event and marker removal is checked against the visible source suffix, the final body must remain stable, Return is sent separately, and success requires a stable recognised empty composer. Cross-window target resolution pins the owning window and rejects blank selectors, every write rechecks the target process, and recovery removes only visible text attributable to the current send while requiring observable progress. Recovery must observe owned text before it can accept an empty composer, and delivery state changes only when the body transaction begins, so a pre-write retry interruption remains safe to classify. Failures after delivery starts are not retried. Codex receives Return by default for live steering, with `--queue` as the explicit Tab opt-in. The documented limits state that plain terminal read-back cannot distinguish a source space consumed at a visual wrap from a missing space at that boundary.

The README and both agent skills document setup, approval rules and failure semantics. The focused suite covers 128 paths across wrapping, periodic ambiguity, partial writes, stale dialogs, wrong targets, cleanup ownership, window changes, accepted and ambiguous submissions, blank selectors, and interrupt boundaries.

Verification:

- `AGTERMCTL=/definitely-missing-agtermctl python3 cookbook/two-agent-chat/test_peer_chat.py`
- `find cookbook -type f -name '*.py' -print0 | xargs -0 -r uvx ruff check --ignore EXE001`
- mutation checks for chunk receipt, stable cleanup progress, delayed body rendering, exact empty-composer acceptance, blank selector rejection and handler-entry and loop-hand-off interrupt classification
- 2,460 sequenced characters delivered byte for byte to Codex in a live split
- 2,400 sequenced characters delivered byte for byte to Claude Code in a live split

### DIFF
--- a/cookbook/two-agent-chat/README.md
+++ b/cookbook/two-agent-chat/README.md
@@ -8,7 +8,7 @@ Let Claude Code and Codex hold a conversation with each other in one agterm spli
 
 Two coding agents run side by side in a split session, one per pane, and talk to each other directly. Each sends a single line into the other's composer with `peer-chat.py`, and the reply arrives in its own pane as an ordinary prompt. You watch both halves of the exchange without relaying anything by hand.
 
-The value is disagreement. An agent working alone accepts its own reasoning; a second one with its own context attacks it first, and what comes back is a located disagreement or a checked fact rather than agreement.
+The value is disagreement. An agent working alone accepts its own reasoning; a second one with its own context attacks it first, and what comes back is a located disagreement or a checked fact instead of agreement.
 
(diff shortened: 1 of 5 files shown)


umputun:
lgtm.

the core delivery path matches the description: each typed event is capped at 197 bytes, the known-empty composer and the column-2 caret are both checked before the first write, and submit is a separate write once the body settles. The retry rule is the one that matters and it is right - retries are limited to `PromptBlocked`, which is only raised in that read-only preflight, and `delivery.phase` becomes "started" after it, so a send cannot write and then retry.

128 tests pass, ruff clean. Window pinning is a real improvement: the previous script had no window selector at all, so its `tree` reads followed the active window.

PR #529, opened 2026-09-02
feat(control): let session.paste address a pane

ssgreg:
`session.paste` was main-pane only, while its documented read-back `session.text` takes `--pane`. So
scripting a split meant writing one pane and reading another, and multi-line text could not reach a
split or scratch pane: `session.type` sends a real Return per newline, which submits it line by line.

The repo already records the gap. `cookbook/annotate-claude-replies` falls back to typing a one-line
pointer at the notes file when the chord fires in a split or scratch pane, and its README says why:
"It also has no `--pane` and always targets the session's main pane".

**The shape**

`--pane` is parsed in the dispatcher, through the same `parsePane` that `session.status` and
`session.restore` use, and the app action takes the parsed `StatusPane`. Three consequences:

- the role and position aliases the shared `validatePaneArgument` already accepts now resolve, so
  `session paste --pane split` works. Worth knowing, and the reason I did it this way: they do not
  resolve on the surface I/O commands that match raw spellings in the app target. On master today
  `session text --pane split` validates and then answers `invalid pane: split`, and `session.type`
  and `font.*` carry the same switch. That is pre-existing and this PR leaves it alone. Happy to send
  the family fix separately.
- an unparseable value answers the pinned `--pane must be left, right, or scratch`, the same string
  the CLI throws, so a raw socket client running no `validate()` is answered identically.
- an unknown pane is unrepresentable in the app target, which is where the module boundary wants the
  parsing anyway.

A pane that parses but is not laid out is refused in `session.type`'s words: `session has no split
pane`, `session has no scratch terminal`.

Omitted keeps `addressableSurface`, so existing callers behave exactly as before and a promoted split
survivor is still reached. Without the flag the CLI sends no `args`, which is why the two existing CLI
tests for `session paste` pass unchanged.

**What it leaves alone**

- `session.selectall` shares the binding arm and stays main-only. Its read-back `session.copy` has no
  pane either, and the pair has to resolve one surface.
- No `--pane-id`. Among the surface I/O commands it is on `session.text` alone, and paste follows
  `session.type` and `font`, which take the role only.
- The cookbook recipe still calls paste without `--pane`. Switching it would raise the recipe's
  minimum agterm version to an unreleased one, and that number is yours to set. Its README sentence
  is corrected to describe the recipe rather than the API. Happy to send the recipe change as a
  follow-up.
- `CHANGELOG.md` untouched, per CONTRIBUTING.

**Verification**

`make test` 2954 tests, `make test-app` 469, `swiftlint --strict` clean.

Mutation check: with the action ignoring its `pane`, `testPasteIntoTheSplitExpeditesTheSplitPane` and
`testPasteRejectsPanesTheSessionDoesNotHave` fail. The split test needs the main slot filled to mean
anything, since `addressableSurface` is `surface ?? splitSurface`, and a fixture with only the split
filled grants the same spawn permit either way.

Live run against a Debug build of this branch in an isolated instance (`AGTERM_STATE_DIR` +
`AGTERM_CONTROL_SOCKET`), ending with a raw socket client so the CLI's `validate()` is out of the way:

```
--pane split -> ok: True        # answered `invalid pane: split` before this change
  PASS alias landed in the split
  PASS right landed
  PASS main landed
  PASS main free of the split markers
raw pane='other'  -> ok: False err: --pane must be left, right, or scratch
raw pane='bottom' -> ok: True
```

The pasted text sits at the target pane's prompt unsubmitted, which is what `session.paste` promises.

### DIFF
--- a/.claude/rules/control-api.md
+++ b/.claude/rules/control-api.md
@@ -286,7 +286,11 @@ side, and reads `lastAppliedIsDark` when bare. Refuse it outside XCUITest; provi
   and an unrealized pane is `session not realized` — `readSelection` cannot tell the two apart, and copy is
   select-all's read-back, so both name that state the same way. It stays on the PANE while an overlay covers
   it, so a selection made inside one is `session.overlay.copy`'s, not this command's.
-  `session.paste` and `.selectall` run Ghostty bindings on main. They use
+  `session.paste` and `.selectall` run Ghostty bindings through one arm. `session.paste` takes `--pane`
+  so its `session.text` read-back can name the same pane; the dispatcher parses it into `StatusPane`
+  (`session.status`'s `parsePane`, so the role and position aliases resolve and a raw client gets the same
+  pinned rejection) and the arm takes the parsed value, never a spelling. `.selectall` stays on main, its
+  `session.copy` read-back having no pane either. Omitted, and for select-all always, the pane is
   `Session.addressableSurface = surface ?? splitSurface`, never focus-aware `activeSurface`, so select-all
   and copy share one pane. Read paste through text and select-all through copy.
 - Keep standard SwiftUI Edit routing. `GhosttySurfaceView` implements Copy/Paste/Select All and validation;
--- a/agterm/Control/ControlServer+SurfaceIO.swift
+++ b/agterm/Control/ControlServer+SurfaceIO.swift
@@ -4,13 +4,36 @@ import agtermCore
 /// `ControlServer` arms that reach into a live `GhosttySurfaceView` — font size, selection copy, background
 /// watermark, buffer read, in-terminal search, text injection. Split out for the swiftlint size limit.
 extension ControlServer {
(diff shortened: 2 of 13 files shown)


umputun (commented):
no blockers here. One gap worth closing, on this branch or right after.

`agtermTests/ControlServerSessionActionsTests.swift:439` - nothing proves a paste reaches the pane it addressed. `performBindingAction` throws away libghostty's return, so `ok` only says the surface was realized, and both new hosted cases end in `session not realized`, which means no test ever reaches a realized split or scratch surface. `testPasteIntoTheSplitExpeditesTheSplitPane` does pin pane-to-surface routing, it just stops before delivery.

Fix: one XCUITest next to `ControlAPIUITests.testSessionPasteInsertsClipboardText:944`, which already has the idiom. `seedPasteboard`, paste with `--pane right`, `pollPaneText(target:pane:contains:)` for the marker, then assert it is absent from left. `19b46b68` (#90) and `a17fc350` (#188) shipped that shape for `session.type --pane` and `font --pane`.

PR #534, opened 2026-09-02
Define peer chat collaboration rules

paskal:
The paired peer-chat skills currently tell each agent to answer in its own pane, which does not reach the peer, and they do not define who may write during shared work.

This sends response-bearing messages back through `peer-chat.py` in the same turn while letting closing acknowledgements end the exchange. Shared work uses a fail-closed one-writer handover: only the current owner may make a scoped offer, the sender pauses before delivery, and the receiver writes only after its acknowledgement is confirmed as submitted. Crossing offers, ambiguous sends and failed cleanup keep both agents read-only until ownership is resolved. Returning control uses the same protocol, and status or diff inspection after an interruption does not grant ownership.

Verification:

- `AGTERMCTL=/definitely-missing-agtermctl python3 cookbook/two-agent-chat/test_peer_chat.py`
- `find cookbook -type f -name '*.py' -print0 | xargs -0 -r uvx ruff check --ignore EXE001`
- style lint and diff checks for both skill files

### DIFF
--- a/cookbook/two-agent-chat/SKILL-claude.md
+++ b/cookbook/two-agent-chat/SKILL-claude.md
@@ -60,7 +60,32 @@ Answers, review results, corrections and stop signals always use the default ste
 
 Codex replies by typing into this pane, so its message arrives as an ordinary prompt opening with
 `Chat from Codex: `. Read it as the next line of a conversation, not as a task the user is asking
-for, and answer it here.
+for.
+
+A peer message that asks a question, offers a handover or reports a result that needs attention gets
+a reply through `peer-chat.py` in the same turn. Text written only in this pane's response does not
+reach the peer. Closing acknowledgements, "nothing further" messages and confirmations of work
+already completed that do not transfer ownership end the exchange without another reply. A valid
+offer that transfers or returns ownership requires the acknowledgement described below; that
+acknowledgement closes the exchange and gets no reply.
+
+## Shared work
+
+When the conversation moves into edits or other shared state, exactly one agent writes in a
+worktree at a time. Only the current owner may offer its scope. A handover offer names the scope and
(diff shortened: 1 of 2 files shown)


umputun (approved):
merging. First half is a clean fix, "answer it here" never reached the peer, and the new vocabulary matches what `peer-chat.py` reports.

notes on the Shared work section for whenever you next touch the recipe, none blocking:

- **no initial owner.** Only the current owner may offer, and nobody owns anything at the start, so by the letter ownership can never be acquired. An agent reads that as either "no one edits" or "first to edit owns", and two agents picking the second is the collision the section is there to prevent. One sentence fixes it: the agent whose prompt came from the user rather than from a `Chat from` label starts as owner.
- **worktree vs scope.** "Exactly one agent writes in a worktree at a time" is worktree-level exclusivity, the rest of the section is per-scope. Two agents holding different scopes in one worktree is both allowed and forbidden. "Scope" is never defined either, so two of them can overlap on one file.
- **a delivered offer is one-way.** After delivery the sender is read-only until control returns. The recipe's own "A reply is not promised" says it may never come, so one ignored offer leaves the sender read-only for the rest of the session.
- **"the user resolves ownership" has no channel.** It ends every failure path, but the user types into one pane and the other never hears about it.
- **the simpler shape.** An alternative to the four fixes above: one writer named for the session removes the bootstrap, the handover and the freeze. Symmetric handover only pays for itself if the two hand work to each other without the user, and the user is reading both panes anyway. Two lines are worth keeping either way: the non-writer does not edit the worktree, and after an interruption both read `git status` and the diff instead of assuming. The cost of static is the non-writer's diff arriving as prose the writer retypes, which a patch file under `/tmp` fixes. Your call, you run this workflow.

the README does not mention the protocol, and the freeze case belongs in Limits per CONTRIBUTING. Same non-blocking bucket.

PR #535, opened 2026-09-02
Simplify peer chat shared-work ownership

paskal:
Previously, shared work mixed worktree-level exclusivity with undefined scopes and depended on acknowledgements that might never arrive. After this change, the agent receiving the user's initiating request remains the sole writer for that worktree until the task ends; the peer stays read-only and may supply a verified patch through private temporary files.

Conflicting direct requests and interrupted ownership require the user to stop the other pane and assign one writer directly. The README now documents the fixed-writer rule and its limits.

Verification:

- `python3 cookbook/two-agent-chat/test_peer_chat.py` (128 tests)
- Both shipped skill files pass `quick_validate.py`
- `git diff --check origin/master...HEAD`


umputun (approved):
lgtm.

two notes for later, neither blocking:

- switching writers mid-task only reaches the new writer's pane. The stopped pane still holds the initiating request as its authority, and the resume clause ("if the writer is unclear, stay read-only") does not fire there, since from that pane the writer is not unclear.
- `mktemp` with no template puts the file in `$TMPDIR` (`/var/folders/...`) on macOS, not `/tmp`, and gives it no recognizable name, so "remove any survivors first" after an interruption means globbing a shared temp dir. An explicit template, `mktemp /tmp/peer-chat-patch.XXXXXX`, would make both the stated path and the cleanup true.

PR #538, opened 2026-09-03
Clarify peer chat writer reassignment

paskal:
Previously, stopping a writer's turn did not revoke its authority, so the old pane could resume writing after another pane was assigned. Bare `mktemp` also placed files in the macOS temporary directory instead of the documented `/tmp` path and provided no safe name for exact cleanup.

After this change, writer reassignment and conflicting direct grants require revocation in the losing pane before direct assignment in the chosen pane. Proposed patches use `/tmp/peer-chat-patch.XXXXXX`, retain exact paths, and forbid wildcard cleanup.

Verification:

- 128 peer-chat recipe tests pass.
- Both peer-chat skill files pass validation.
- The explicit temporary-file template creates a mode-0600 file under `/tmp` on macOS.
- The committed diff passes Git's whitespace check.


umputun (approved):
lgtm

PR #563, opened 2026-09-07
feat(cookbook): add the remote-claude-session recipe

andr81:
Adds `remote-claude-session`: one chord runs Claude Code in a tmux session on a remote host, so a closed laptop stops taking the agent with it. The tab reconnects on its own, survives an agterm restart, and the agent's status still lands on its sidebar row.

`open` picks a running remote session or a project and opens a tab for it. Inside, the tab's own process is a reconnect loop around `ssh -t`, pinned with `session restore` so a restarted agterm reattaches rather than replaying a bare ssh. `end` kills the remote session and closes the tab.

**Status without giving the host the control socket.** The host reaches the sidebar row through `agt-remote-relay.py` on the Mac, over an `ssh -R` forward to a unix socket in a private per-attach directory. The relay holds the tab's target as a fixed argument, accepts one line, and only if it is a `session.status` whose status is one of the four states; it then rebuilds the request with its own target. Every other line, including any target the host names, is refused. So the worst a compromised host can do to the Mac is set that one tab's status, and `AGT_REMOTE_STATUS=0` removes even that.

**Destructive behavior**, stated in *Limits*: `end` kills the remote tmux session and everything running in it, with no undo, and the picker is the only confirmation. Closing the tab any other way deliberately leaves the session running on the host, so sessions accumulate until you end them.

*Limits* also covers what `install` writes on the host, what the forwarded ssh agent is exposed to and for how long, and that the token stored on the host is your subscription.

**Requirements**: agterm 0.22.0 or later, for the custom-command `PATH` fix; `jq`, `ssh` and Python 3.9+ on the Mac; `tmux`, Python 3.9+, `uuidgen` and `git` on the host. The relay is standard library only.

The recipe has driven real sessions on my own host, which is where the later commits come from: the agent never starting because `send-keys` takes a target-pane and the `=name` prefix addresses a session; the OAuth token being accepted as an argument, where it lands in argv and shell history; and a shared `~/.ssh/agent.sock` repointed from `~/.ssh/rc`, which sshd runs for every ssh command to the host, so a short `list` left it dangling and broke `git` in every live session. Each session now owns its link to the forwarded agent.


umputun (commented):
good to merge, the recipe does what it says and the relay boundary holds. Everything below is a note, none of it blocks; fix what you want on this branch and I take the rest after.

the one that bites on an ordinary host:

`agt-remote.sh:121` - `remote list 2>&1` folds the host's stderr into the listing, and any line without a tab becomes a picker row with a null label, which `pick` cannot decode, so the chord says "picker failed (exit 1)" and opens nothing. A Debian host without the locale macOS forwards via `SendEnv LC_*` prints a setlocale warning on every ssh command, a `Banner` or an echo in bashrc does the same. Fix: `map(select(length > 1 and (.[0] == "S" or .[0] == "P")))` before the item map, or keep stderr separate.

smaller ones in the scripts:

`agt-remote-host.sh:266` - the token goes into `env.tmp` at 0644 before the `chmod 600`, and the file stays behind if the ssh channel drops before the `mv`. Fix: `umask 077` at the top of `auth()` plus `rm -f "$env.tmp"` before the grep.

`agt-remote-host.sh:155,164,176` - the three `>>` appends glue onto the reader's last line when `~/.tmux.conf`, `~/.profile` or `known_hosts` has no trailing newline, and the known_hosts entry then re-appends on every rerun. Fix: `[ -s "$f" ] && [ -n "$(tail -c1 "$f")" ] && printf '\n' >>"$f"` before each write.

`agt-remote-host.sh:185` - the `host ready` line with "hooks merged" prints right after "hooks NOT merged" on the failure path. Fix: carry the `merge_hooks` status into that line.

`agt-remote-host.sh:127` - `tmux list-sessions` is unfiltered, so every session on the host shows badged as remote, and picking a stray one ends in "remote command exited with 2". Is that deliberate? If not, emit `#{session_path}` and keep only rows directly under `$PROJECTS`.

`agt-remote.sh:192` - the pin banner only fires when the save fails; when restore mode is not Re-run the server answers ok with a text note, so the common cause of a tab coming back as a plain shell gets no warning. A `agtermctl restore mode` read in `open()` would cover it.

`agt-remote.sh:296-324` - the real attach's `-R` rides the shared ControlMaster and nothing cancels it, so each reopen leaves a stale loopback listener on the host until the master exits. From reading OpenSSH's mux.c, not reproduced. `ssh -O cancel -R ...` after each attachment returns would clean it up.

README:

`README.md:21,129,178` - "Restore running commands on restart" is gone, the setting is the "Restore sessions" picker set to "Re-run commands". Default is Fresh shells, so a reader who cannot find the toggle gets no reattach.

`README.md:138` - the runner does export `AGT_SESSION_ID` (your own `end()` reads it at :335). The token earns its place because a body without a session token can run sessionless, not because the env lacks the id.

`README.md:106` - "`~/.profile` sources the file" is false where bash has `~/.bash_profile` (RHEL, Fedora, Amazon Linux). Only a hand-opened shell is affected, worth a qualifier.

`README.md:166` - "what the host can do to the Mac is set this one tab's status, whatever else is on the host" overreaches with `ForwardAgent yes` and OSC 52 clipboard writes enabled. Limits at :182 has it right, "through this bridge" fixes the sentence.

`README.md:194` - the tmux block is global and last-wins over the reader's own `set-clipboard`, `allow-passthrough` and `history-limit`, worth a clause in Limits.

andr81:
Went through all twelve notes — every one is addressed in e6d6384. Two of the calls were not obvious, so they get a word each.

**`list` now filters.** Answering "is that deliberate?" — it was not. The format emits `#{session_path}` and only sessions whose working directory sits directly under `$PROJECTS` survive; the project name the picker row already showed comes from that same field, so there is no extra round trip. The host's own tmux sessions no longer appear badged.

**`ssh -O cancel -R` is a blind fix.** Like you, I could not reproduce it; the change rests on the same reading of mux behaviour rather than on an observation. It is cheap and safe: with no master to ask the command fails silently, and `rc` is captured before it runs, so the loop's branching is untouched.

The rest, in order:

- `remote list 2>&1` — stderr is still collected, otherwise `fail` has nothing to report, but only well-formed rows become items: `select((.[0] == "S" and length == 4) or (.[0] == "P" and length == 3))`.
- `auth()` — `umask 077` at the top and `rm -f "$env.tmp"` before the grep.
- the three `>>` appends — a shared `end_newline`, called before each write, including `known_hosts` ahead of the loop.
- `host ready` — the real `merge_hooks` status now reaches that line.
- the restore pin — the warning covers the success path too: `agt restore mode --json` is read best-effort and a `configured == none` raises a banner. On agterm older than 0.26 the command is unknown, stdout is empty and no banner appears, so the recipe's 0.22.0 floor stands.
- README — the setting is the **Restore sessions** picker set to **Re-run commands**, with **Fresh shells** named as the default, in all three places; `AGT_SESSION_ID` is described as reaching the command's environment, with the token earning its place on the line for the other reason — a body naming no session-scoped token is one agterm will fire from an empty window; `~/.profile` gained the `~/.bash_profile` qualifier; the bridge sentence narrowed to "through this bridge"; the tmux block is called out in Limits as global and last-wins over its three settings.

One extra: `PROJECTS` is normalised for a trailing slash, without which the new filter in `list` would miss on `AGT_REMOTE_PROJECTS=projects/`.

umputun (commented):
merging this. Everything below is a note, none of it blocks; fix what you want on the branch and I take the rest after.

all twelve from the last round are in. The restore-mode read is the one that came back with a gap:

`agt-remote.sh:237-242` - the check warns on `configured == none`, but Live can be configured and still refused for a given launch: `RestoreMode.launchDecision` sets `active` to none with an `unavailableReason` when zmx is unusable, most often on a Mac whose login shell is not zsh. `configured` still reads `live`, so no banner, and the tab comes back a plain shell. Fix: warn also when `configured` is `live` and `unavailableReason` is non-empty, and say in *Requirements* that Live wants zsh as the login shell.

new in this push:

`README.md:21,129` - the **Restore sessions** picker only exists from 0.26.0; 74efbb0 is where it replaced the checkbox, and on 0.22 to 0.25 the same setting is **Restore running commands on restart**, so a reader on the floor stated one line above cannot find the control the step names. Same bullet at :20 puts `session hud` among the commands older than the floor, but `--position top-center` and `--text-color` shipped in 0.22.0 (1923906), so the sentence argues against the floor it is justifying.

`README.md:178` - the pin does not fire under **Live sessions**: `LaunchSeed`'s wrapped branch hands off to `ZmxLaunch.surfaceSeed`, which never reads it, and the fallback branch sets `restoreEnabled: false`. Under Live the tab comes back because the pane's own process survived under zmx, which a force quit or a power loss defeats. Worth splitting that sentence by mode; `attach NAME PROJECT` is the recovery under Live too.

missed last round, in the scripts:

`agt-remote.sh:410-411` - `[ "$rc" -eq 0 ] || exit 0` folds a picker failure into a cancel, so `end` goes silent on exit 1, `pick already pending` or the socket unreachable, where `open` reports it. Fix: `open`'s own case block.

`agt-remote.sh:428` - `cat > '$REMOTE_BIN'` truncates the installed script before the transfer finishes and the executable mode survives, so a dropped channel leaves a partial script that still runs. Cut between functions it defines them, falls off the end and exits 0, so `remote list` comes back empty and `open` then reports "no sessions or projects yet", which names the wrong problem. Fix: copy to a temp beside it, then chmod and mv in a second ssh that runs only when the copy exited 0. `cat > tmp && mv` in one remote command does not help, since `cat` cannot tell a finished sender from a dropped channel.

`agt-remote-host.sh:69-75` - with no `uuidgen` the empty `id` is written to `NAME.claude`, the transcript glob misses, and the typed line ends `--session-id` with nothing after it, so the agent never starts. The empty marker is read back on every later recreation of that name, so installing uuid-runtime afterwards does not repair it. Fix: check the id before writing the file.

README:

`README.md:166` - a line naming another target is not answered `refused`: `clean_request` ignores the field and rebuilds the request with its own, so the answer is `ok:true`. The confinement holds, the sentence describes a rejection that does not happen, and the relay docstring says the same. Same paragraph and :182: loopback is what the client asks for, and `GatewayPorts yes` on the host overrides it to the wildcard address, so "another user on a shared host" understates who can reach it there.

`README.md:174` - the *Limits* entry for `end` lists what it destroys and leaves out the conversation id, which `kill_session` removes with the rest. Worth a clause: reopening the same name after `end` starts a fresh conversation.

`cookbook/README.md:47` - *needs* omits python3, unlike the eight rows that ship an interpreter.

PR #578, opened 2026-09-10
fix(tests): replace the session-host fixture's copied nc with a built client

### LINKED ISSUE #577: SessionHostServerTests: the fixture's copied /usr/bin/nc is SIGKILLed on Apple Silicon
## What happens

On a stock Apple Silicon Mac with SIP enabled, six tests in `SessionHostServerTests` fail, all with the same two issues each:

```
✘ hostSocketIsNotEnumeratedByStockZmx() — Expectation failed: try fixture.exchange(.stop).last == .stopped
✘ hostSocketIsNotEnumeratedByStockZmx() — Caught error: NSPOSIXErrorDomain Code=60 "Operation timed out"
```

`swift test` reports `3573 tests in 151 suites failed ... with 11 issues`. The affected tests are `hostSocketIsNotEnumeratedByStockZmx`, `malformedOversizedAndPartialConnectionsDoNotStopTheHost`, `stalledPeerHitsItsDeadlineAndOnlyThatConnectionCloses`, `forgedHelloDoesNotOverrideTheActualPeerImage`, `stopKeepsLockInodesForTheNextHost` and `killedHostReleasesLockAndListenerWhileDaemonAndShellSurvive`.

## The cause

`Fixture` copies `/usr/bin/nc` into its fake app bundle to get a socket client with the same bundle identity:

https://github.com/umputun/agterm/blob/master/agtermCore/Tests/SessionHostRuntimeTests/SessionHostServerTests.swift#L441

`/usr/bin/nc` is a universal `x86_64 arm64e` binary. Its arm64e slice is only executable because it is a platform binary in the system trust cache. A copy anywhere outside the system volume is not, and AMFI kills it on exec.

```
$ cp /usr/bin/nc /tmp/plain-copy && /tmp/plain-copy -h ; echo "exit $?"
exit 137            # 128 + 9 = SIGKILL, no output

$ cp /usr/bin/nc /tmp/resigned && codesign --force --sign - /tmp/resigned
$ /tmp/resigned -h ; echo "exit $?"
exit 137

$ cp /usr/bin/nc /tmp/x86only && lipo -extract x86_64 /tmp/x86only -output /tmp/x86only
$ codesign --force --sign - /tmp/x86only && /tmp/x86only -h ; echo "exit $?"
usage: nc [-46AacCDdEFhklMnOortUuvz] ...
exit 1              # same as the system nc
```

The ad-hoc re-signing is not what breaks it — an untouched byte copy dies the same way. Removing the arm64e slice is what makes it run.

## The failure run

1. `Fixture.start()` succeeds. The host writes its pidfile and binds its socket — observed pidfile `54332` matching `ps`, with `session-host.sock` beside it. The disclaimed spawn, the bundle identity check and the listener are all fine.
2. `fixture.list()` succeeds. That shells out to the staged `zmx`, a normally signed binary.
3. `fixture.exchange(.stop)` runs `Test.app/Contents/MacOS/test-client -U <socket>`, the copied `nc`. It is SIGKILLed before it writes anything.
4. `raw()` therefore parses an empty stdout into `[]`, so `#expect(... .last == .stopped)` fails and the host is never told to stop.
5. `waitForExit()` then throws `ETIMEDOUT` after five seconds. That is the second issue on each test.

Only the borrowed client is affected. Nothing here suggests a problem with the session host itself.

## Environment

- macOS 26.0.1 (25A362), Apple Silicon, SIP enabled
- reproduced with a single-test filter, so it is not parallel-test contention

## Why CI does not see it

CI runs this suite in the `Build app` job on `macos-26` and it passes, so something differs between a GitHub-hosted runner and a stock Mac. I could not verify what from outside; the AMFI or SIP configuration on hosted runners is the obvious candidate. Flagging it as a hypothesis, not a finding — the local reproduction above is the part I actually measured.

## Possible fixes

Listing rather than recommending, since the trade-offs are yours:

- build a small test-only client executable in the package and copy that into the bundle, which removes the platform-binary problem entirely;
- `lipo -remove arm64e` on the copy, which is one line but leaves only the x86_64 slice and so needs Rosetta;
- use a netcat that is not a platform binary.

Happy to send a PR for whichever you prefer.

p4elkin:
Fixes #577.

## The problem

`SessionHostServerTests.Fixture` copies `/usr/bin/nc` into its fake app bundle to get a socket client with the same bundle identity. `/usr/bin/nc` is a universal `x86_64 arm64e` binary, and its arm64e slice is only executable because it is a platform binary in the system trust cache. Any copy outside the system volume is not, so AMFI kills it on exec.

The ad-hoc re-signing is not the cause — an untouched byte copy is SIGKILLed the same way, and removing the arm64e slice is what makes it run:

```
$ cp /usr/bin/nc /tmp/plain-copy && /tmp/plain-copy -h ; echo "exit $?"
exit 137            # 128 + 9 = SIGKILL
```

Six tests failed as a result. Every `exchange()` parsed empty stdout into `[]`, so `.stop` never reached the host and `waitForExit()` threw `ETIMEDOUT` five seconds later. On a stock Apple Silicon Mac with SIP enabled, `swift test` reported 11 issues; CI does not see it, which #577 goes into.

## The change

Adds a small `session-host-test-client` executable target — a target, deliberately not a package product, so it never ships in the app. It speaks the same `-U SOCKET` form the fixture already passed, so only the copied path changes.

It polls the socket and stdin together rather than draining stdin first. `stalledPeerHitsItsDeadlineAndOnlyThatConnectionCloses` holds stdin open forever, and the client still has to end when the host closes its side.

`forgedHelloDoesNotOverrideTheActualPeerImage` keeps using the real `/usr/bin/nc` at its own path. That is the point of that test, and a binary run from the system volume is unaffected.

## Verification

On macOS 26.0.1, Apple Silicon, SIP enabled:

- `swift test --filter SessionHostServerTests` — 22 tests pass in 3.3s, where six previously failed
- `swift test` — 3146 tests in 128 suites pass
- `swiftlint --strict` — clean

Happy to take a different approach if you would rather not add a target; #577 lists the alternatives I considered.

### DIFF
--- a/agtermCore/Package.swift
+++ b/agtermCore/Package.swift
@@ -36,6 +36,10 @@ package.targets += [
     .target(name: "SessionHostTrampoline"),
     .target(name: "SessionHostRuntime", dependencies: ["SessionHostTrampoline", "AgtermResponsibility", "agtermCore"]),
     .executableTarget(name: "agterm-session-host", dependencies: ["SessionHostRuntime"]),
-    .testTarget(name: "SessionHostRuntimeTests", dependencies: ["SessionHostRuntime", "agterm-session-host"]),
+    // Not a product: the fixtures need a socket client they can copy into a test bundle, which
+    // /usr/bin/nc cannot be (see the target's own comment).
+    .executableTarget(name: "session-host-test-client"),
+    .testTarget(name: "SessionHostRuntimeTests",
+                dependencies: ["SessionHostRuntime", "agterm-session-host", "session-host-test-client"]),
 ]
 #endif
--- a/agtermCore/Sources/session-host-test-client/main.swift
+++ b/agtermCore/Sources/session-host-test-client/main.swift
@@ -1,0 +1,63 @@
+import Darwin
+import Foundation
+
+// A stand-in for `nc -U` inside the test bundle. The fixtures need a client whose executable sits
+// beside the host's, so the host's peer check resolves the same bundle identity; `/usr/bin/nc` cannot
+// serve because its arm64e slice is only executable as a platform binary, and any copy of it outside
+// the system volume is killed by AMFI (issue #577).
+
+let arguments = CommandLine.arguments
+guard arguments.count == 3, arguments[1] == "-U" else {
+    FileHandle.standardError.write(Data("usage: session-host-test-client -U SOCKET\n".utf8))
+    exit(2)
+}
+
+let connection = socket(AF_UNIX, SOCK_STREAM, 0)
+guard connection >= 0 else { exit(1) }
+var address = sockaddr_un()
+address.sun_family = sa_family_t(AF_UNIX)
+address.sun_len = UInt8(MemoryLayout<sockaddr_un>.size)
+let path = arguments[2].utf8CString
+guard path.count <= MemoryLayout.size(ofValue: address.sun_path) else { exit(1) }
+withUnsafeMutableBytes(of: &address.sun_path) { buffer in path.withUnsafeBytes { buffer.copyMemory(from: $0) } }
+let connected = withUnsafePointer(to: &address) {
+    $0.withMemoryRebound(to: sockaddr.self, capacity: 1) { connect(connection, $0, socklen_t(MemoryLayout<sockaddr_un>.size)) }
+}
+guard connected == 0 else { exit(1) }
+
+// Both directions stay live at once: a stalled peer keeps stdin open forever, and the run must still
+// end when the host closes its side.
+var input = Int32(STDIN_FILENO)
+var buffer = [UInt8](repeating: 0, count: 65536)
+while true {
+    var entries = [pollfd(fd: connection, events: Int16(POLLIN), revents: 0)]
+    if input >= 0 { entries.append(pollfd(fd: input, events: Int16(POLLIN), revents: 0)) }
+    guard poll(&entries, nfds_t(entries.count), -1) >= 0 || errno == EINTR else { break }
+    if entries[0].revents != 0 {
+        let count = read(connection, &buffer, buffer.count)
+        if count <= 0 { break }
+        var written = 0
+        while written < count {
+            let step = buffer.withUnsafeBytes { write(STDOUT_FILENO, $0.baseAddress!.advanced(by: written), count - written) }
+            if step <= 0 { exit(0) }
+            written += step
+        }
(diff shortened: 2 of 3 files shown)


umputun (approved):
the approach is right. A first-party client drops the dependency on nc's slice layout and signing status entirely, and keeping it a target rather than a product keeps it out of the app build. Verified that last part: `package.products` is untouched, `project.yml` builds helpers by `--product` name, and `scripts/release.sh` signs an explicit list, so the new target never reaches a Release app.

one thing to fix first. The comment at `agtermCore/Sources/session-host-test-client/main.swift:4-7` and the PR body both say any copy of nc outside the system volume is killed by AMFI. That is not true on macOS 26.6.2. A plain copy dies exactly as you describe, but the ad-hoc re-signed copy the old fixture produced runs fine:

```
$ cp /usr/bin/nc /tmp/plain && /tmp/plain -h ; echo "exit $?"
exit 137

$ cp /usr/bin/nc /tmp/resigned && codesign --force --sign - /tmp/resigned
$ /tmp/resigned -h >/dev/null 2>&1 ; echo "exit $?"
exit 1
```

so with the staged zmx in place, master passes all 22 in `SessionHostServerTests` here. I can't tell you why our results differ, different hardware and OS revision and possibly SIP state, so pls qualify the claim to the environment you measured instead of stating it universally.

that doesn't change my view of the PR. A fixture that turns on undocumented AMFI-versus-codesign behavior, which already differs between two machines, is worth removing whichever one you happen to be on.

optional, not blocking: the socket at `main.swift:15` doesn't set `SO_NOSIGPIPE`. `malformedOversizedAndPartialConnectionsDoNotStopTheHost` sends `maximumFrameBytes + 1`, so the client always makes a second write after the 65536-byte one, and the host closes on `frameTooLarge` at that point. If the close wins the race, the client dies of signal 13 instead of reaching its `exit(1)` at line 56. I ran that test 30 times without seeing it, and the two existing test-side sockets don't set it either, so this is hardening rather than a fix:

```swift
var noSigPipe: Int32 = 1
setsockopt(connection, SOL_SOCKET, SO_NOSIGPIPE, &noSigPipe, socklen_t(MemoryLayout<Int32>.size))
```

`agtermCore/Sources/agtermctlKit/SocketClient.swift:70-75` has the same three lines with a comment on the case.

PR #608, opened 2026-09-14
cookbook: agst tool

Rulexec:
Add `long-commands-status` cookbook recipe: a `agst` shell wrapper that sets the current agterm session's status to `active` while a command runs, then `completed --auto-reset` on success or `blocked --auto-reset` on failure. Covers the case the bundled agent-status hooks don't — plain shell commands (`make test`, `cargo build`, etc.) rather than an agent's own lifecycle.


umputun (changes requested):
two things to fix before merge, the rest are notes.

`cookbook/long-commands-status/agst.sh:31-44,66,73,75` - the option values are collected into one string and expanded unquoted. Shellcheck fails with SC2086 five times, which is the cookbook CI gate, and a `--socket` path with a space (the default one lives under `Application Support`) is split into several arguments, the error is swallowed by `|| :`, and no status shows up at all. Fix: keep each option in its own variable and build the status calls from quoted positional parameters (`set -- "$@" --socket "$socket"`). A `# shellcheck disable=SC2086` alone silences CI but leaves the path case broken.

`cookbook/long-commands-status/agst.sh:60` - `status()` sends only `--target`, so a status set from a split pane is recorded against the main pane. That lets `agst` in a split overwrite the blocked glyph of an agent waiting in the main pane, sends attention navigation for a failed command to the main pane instead of the split, and in the mirror case the write is refused silently. Fix: add `--pane "$AGTERM_PANE"` and `--pane-id "$AGTERM_PANE_ID"` when they are set, never with an empty value, same as the other status recipes do.

not blocking:

- `README.md:19` and the index row - `--auto-reset` predates 0.7.1, it is in the earliest tagged release, and `--shape` needs 0.17.0. Either state 0.17.0 and name `--shape` as what shipped in it, or drop `--shape` and state 0.4.0 for `--sound`.
- `agst.sh:60` - `$AGTERM_SOCKET` is not forwarded, the other recipes do. Seed the socket from it and let an explicit `--socket` win, quoted.
- `agst.sh:69-77` - when the command dies from Ctrl-C the wrapper dies with it and the end status is never posted. Worth a sentence in Limits.
- `agst.sh:46` - `agst --help` runs `--help` as the command, exits 127 and marks the row blocked. A `-h|--help) usage; exit 0 ;;` arm fixes it.
- Limits could mention that `tree` reports the pane's foreground as the `sh` wrapper argv rather than the wrapped command while it runs.

Rulexec:
> two things to fix before merge, the rest are notes.
> 
> `cookbook/long-commands-status/agst.sh:31-44,66,73,75` - the option values are collected into one string and expanded unquoted. Shellcheck fails with SC2086 five times, which is the cookbook CI gate, and a `--socket` path with a space (the default one lives under `Application Support`) is split into several arguments, the error is swallowed by `|| :`, and no status shows up at all. Fix: keep each option in its own variable and build the status calls from quoted positional parameters (`set -- "$@" --socket "$socket"`). A `# shellcheck disable=SC2086` alone silences CI but leaves the path case broken.
> 
> `cookbook/long-commands-status/agst.sh:60` - `status()` sends only `--target`, so a status set from a split pane is recorded against the main pane. That lets `agst` in a split overwrite the blocked glyph of an agent waiting in the main pane, sends attention navigation for a failed command to the main pane instead of the split, and in the mirror case the write is refused silently. Fix: add `--pane "$AGTERM_PANE"` and `--pane-id "$AGTERM_PANE_ID"` when they are set, never with an empty value, same as the other status recipes do.
> 
> not blocking:
> 
> * `README.md:19` and the index row - `--auto-reset` predates 0.7.1, it is in the earliest tagged release, and `--shape` needs 0.17.0. Either state 0.17.0 and name `--shape` as what shipped in it, or drop `--shape` and state 0.4.0 for `--sound`.
> * `agst.sh:60` - `$AGTERM_SOCKET` is not forwarded, the other recipes do. Seed the socket from it and let an explicit `--socket` win, quoted.
> * `agst.sh:69-77` - when the command dies from Ctrl-C the wrapper dies with it and the end status is never posted. Worth a sentence in Limits.
> * `agst.sh:46` - `agst --help` runs `--help` as the command, exits 127 and marks the row blocked. A `-h|--help) usage; exit 0 ;;` arm fixes it.
> * Limits could mention that `tree` reports the pane's foreground as the `sh` wrapper argv rather than the wrapped command while it runs.

Both blocking issues fixed in `agst.sh`:

- SC2086 / `--socket` path splitting: each option now lives in its own variable and every status call is built from quoted positional parameters (`set -- "$@" …`). No `# shellcheck disable`, shellcheck clean.
- `--pane` / `--pane-id`: forwarded when set, never empty — same as the stock wrapper and the other status recipes.

Non-blocking applied:

- Seeded `socket` from `$AGTERM_SOCKET`, explicit `--socket` wins, quoted.
- `-h|--help` arm added, exits 0 before the command check.
- Requirements bumped to 0.17.0 naming `--shape` (covers every flag the recipe uses); index row mirrored.
- Two Limits bullets added: Ctrl-C kills the wrapper so no end status posts, and `tree` shows the `sh` wrapper argv mid-run.

-----

I had thought about trapping Ctrl+C, but I do not think it is worth doing. It would create a large amount of code and logic, but the next command call would fix it anyway, or it can be manually fixed by calling agst echo. At least for me it is not very bothersome.

umputun (approved):
lgtm

PR #620, opened 2026-09-17
fix(cookbook): guard the stuck-shape compare against a blanked variable

x9x9x9x9x9x91:
Follow-up to #462, the two non-blocking notes from the approval.

- `status-sweep.sh:367` and `:395`: the stuck-shape comparisons never got the guard the work-shape case has. With `AGT_SHAPE_STUCK` blanked, every row with no shape override compared equal and the branch pulsed over the machinery tint. Both sites now require a non-empty stuck shape first; `:416` was already right.
- `README.md:186`: narrowed the adapter claim. It stays silent while a second agent binary sits between the hook and the pane, and fails open when the chain is unreadable or the spawner has already exited.

`bash -n` and shellcheck clean.


umputun (approved):
lgtm

PR #624, opened 2026-09-18
fix(zmx): drain the child's pipes while waiting for it

### LINKED ISSUE #623: tree and window list stall the main thread for 3 s once four Live daemons exist
With `Live sessions` restore mode and four or more zmx daemons, every `agtermctl tree` and `window list` call blocks the app's main thread for the full `ZmxClient` timeout (3 s). The UI freezes for that time on each call, and since the agent hooks call `tree` constantly, the app stutters continuously. Seen on 0.30.1 after updating; reproduced on `master` at `17e37f3`.

**Cause.** `ZmxClient.run` (`agterm/Ghostty/ZmxClient.swift`) waits for the child to exit before reading its stdout. `zmx list` writes row by row, so the pipe never grows past its initial 512 bytes (xnu only grows a pipe buffer on an empty pipe). Four rows (~590 bytes) fill it: zmx blocks in `writev`, the app blocks in `finished.wait`, the timeout fires, zmx is killed, and `sessionLeaderPIDs()` returns nil. `buildTree` runs this on the main thread since #574, so it is paid on every `tree`.

**Repro** (`master` `17e37f3`, Debug build, isolated `AGTERM_STATE_DIR`, `settings.json` = `{"restoreMode":"live"}`, one Live pane open):

```sh
ZMX=build/DerivedData/Build/Products/Debug/agterm.app/Contents/MacOS/zmx
export ZMX_DIR=$(agtermctl zmx list --json --socket "$AGTERM_STATE_DIR/agterm.sock" | jq -r .result.zmx.endpoint.socketDirectory)
for i in 1 2 3; do $ZMX run dummy-$i -d cat; done      # 4 daemons including the pane's own
time agtermctl tree --json --socket "$AGTERM_STATE_DIR/agterm.sock" > /dev/null
```

Measured:

| daemons | `zmx list` bytes | `tree --json` |
|---|---|---|
| 1–3 | 124–434 | 0.03 s |
| 4 | 589 | 3.07 s |
| 61 | 9424 | 3.06 s |

`sample` of the app during the call: main thread in `ControlServer.buildTree` → `ZmxClient.sessionLeaderPIDs` → `ZmxClient.run` → `semaphore_timedwait_trap`; the `zmx list` child in `main.list` → `Io.Writer.defaultFlush` → `writev`.

**A second, related hang.** `Pipe()` fds are inherited by anything spawned meanwhile. libghostty spawns surface commands from its io thread, so a spawn landing between `Pipe()` and `process.run()` puts a write end into a long-lived `zmx attach`, and the read after exit never sees EOF. This hung `SessionHostAcceptanceTests.testPrimaryDaemonLostAfterInventoryIsRecreatedByHost` for 19 minutes in one run here (the test host's `zmx attach` grandchild held both ends of the pipe per `lsof`).

Fix in the linked PR: drain both pipes on their own threads while waiting, mark the fds `FD_CLOEXEC`, and fail the call if EOF still does not arrive after exit. With it, 60 daemons: `tree` in 0.03 s.

p4elkin:
Fixes #623.

`ZmxClient.run` read the child's stdout only after it exited. zmx writes the listing row by row, so the pipe buffer stays at its initial 512 bytes; four Live daemons fill it, zmx blocks on write, the app blocks on exit, and every `tree` and `window list` call ends in the 3 s timeout on the main thread with the leader snapshot dropped.

- Each pipe is now drained to EOF on its own thread (`Sink`, same shape as `RemoteCommandProcessRunner.Sink`) while the call waits for exit.
- The pipe fds are marked `FD_CLOEXEC`. libghostty spawns surface commands from its io thread with plain inheritance, so a spawn landing between `Pipe()` and `process.run()` carried a write end into a long-lived `zmx attach` and held EOF back for its lifetime, hanging the caller (seen as a 19-minute hang of `SessionHostAcceptanceTests.testPrimaryDaemonLostAfterInventoryIsRecreatedByHost`).
- Should a write end still leak, the join after exit gives up after one `terminationGrace` for both pipes and the call throws `timedOut`, so a listing fails rather than coming back silently short. `captureWallClockLimit` is unchanged.

Tests: `testRunDrainsOutputLargerThanThePipeBufferBeforeTheProcessExits` (400 KB across both pipes; fails with `timedOut` before the fix, 35 ms after) and `testRunGivesUpOnAWriteEndHeldPastTheChildsExit` (`sh -c 'sleep 30 & echo row'`; throws in 0.26 s instead of hanging 30 s). `ZmxClientTests` and `SessionHostAcceptanceTests` pass on this branch; `make lint` is clean.

Measured on `master` `17e37f3` with the repro from the issue: 4 daemons `tree --json` 3.07 s before, 60 daemons 0.03 s after.

https://claude.ai/code/session_01JyKCk3HSJFejBkrdM5Y7gi


umputun (approved):
lgtm

